PR #10, opened 2016-07-07
Defining build to run on Alfred

giorgiosironi:
Works, but at the end of the build http://ci.alfred.elifesciences.org/job/test-builder/2/console I get:
```
Coverage.py warning: No data was collected.
...
09:25:55 .unittest.sh: line 15: [: -le: unary operator expected
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,10 @@
+elifePipeline {
+    stage 'Checkout'
+    checkout scm
+
+    stage 'Update'
+    sh './update.sh --exclude virtualbox vagrant'
+
+    stage 'Test'
+    sh './test.sh'
+}


lsh-0:
yeah, I seen that occasionally. what should happen is that after running the tests, it tests the coverage and fails if it's below a certain threshold. the check is in `.unit-test.sh` and is failing on this line: `coverage report | grep TOTAL | awk '{print $6}' | sed 's/%//'` - basically it should be returning a number, it returns nothing and the the `-le` expression is compared against nothing and you get that error.

will check version of coverage.py installed and such, see if I can't pin this one down

lsh-0:
hrm. I don't know what coverage.py isn't collecting data to run the report.

giorgiosironi:
You approved the card but didn't merge this :eyeglasses:

PR #14, opened 2016-07-11
Simplified stacknames to only contain two segments: project name and cluster - `./bldr deploy` takes 2 mandatory arguments - Backward compatibility for ssh access of existing new builder machines

giorgiosironi:


src/buildercore/core.py
```diff
@@ -120,35 +120,30 @@ def stack_conn(stackname, username=config.DEPLOY_USER, **kwargs):
...
-    
 def is_master_server_stack(stackname):
     return 'master-server--' in str(stackname)
 
-@testme
 def is_prod_stack(stackname):
-    pname, inst_id = instanceid_from_stackname(stackname)
-    return inst_id in ['master', 'production']
+    pname, cluster = parse_stackname(stackname)
+    return cluster in ['master', 'production', 'prod']
```
lsh-0: we should get rid of the 'production' here.

PR #15, opened 2016-07-11
Changing hostnames to cluster--project.elifesciences.org

giorgiosironi:


lsh-0:
the `builder-base-formula` will need to be changed too. Looking at the severity of that now

lsh-0:
ok - I can see what needs to be done:
https://github.com/elifesciences/builder-base-formula/blob/master/_modules/elife.py#L117

this 'derived data' section is going to have to be cut out and anything using 'derived' values should pull them directly from what is (now) available after builder the context in cfngen.py. some of that stuff is duplicated now too.

PR #16, opened 2016-07-11
Improving the coverage of utils to 73%

giorgiosironi:
The project gets to 57%. See commit messages for details, mostly
- removing dead code
- covering OS-based functions by mocking the primitives they are using
- smoke testing some things such as pwd()

### DIFF
--- a/src/tasks.py
+++ b/src/tasks.py
@@ -39,7 +39,7 @@ def ami_for_project(pname):
     
     print len(results),"results"
 
-    utils.table(results, ['id', 'root_device_type', 'virtualization_type', 'name'])
+    print utils.table(results, ['id', 'root_device_type', 'virtualization_type', 'name'])
 
     # good for figuring out filters 
     #print results[0].__dict__
--- a/src/tasks.py
+++ b/src/tasks.py
@@ -15,6 +15,7 @@ from buildercore.decorators import osissue, osissuefn
 
 @debugtask
 @requires_project
+@echo_output
 def ami_for_project(pname):
     "returns possible AMIs suitable for given project"
     conn = core.connect_aws_with_pname(pname, 'ec2')
@@ -39,7 +40,7 @@ def ami_for_project(pname):
     
     print len(results),"results"
 
-    print utils.table(results, ['id', 'root_device_type', 'virtualization_type', 'name'])
+    return utils.table(results, ['id', 'root_device_type', 'virtualization_type', 'name'])
 
     # good for figuring out filters 
     #print results[0].__dict__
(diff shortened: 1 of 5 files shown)


src/tasks.py
```diff
@@ -39,7 +39,7 @@ def ami_for_project(pname):
     
     print len(results),"results"
 
-    utils.table(results, ['id', 'root_device_type', 'virtualization_type', 'name'])
+    print utils.table(results, ['id', 'root_device_type', 'virtualization_type', 'name'])
```
lsh-0: there is a decorator called `echo_output` that prints the result of the task. I prefer using that to printing reports of things, especially if a task can be called by itself or as part of another function (like this one could)

src/tests/test_utils.py
```diff
@@ -1,13 +1,27 @@
...
     def tearDown(self):
         pass
 
+    def test_git_remote_branches(self):
+        branches = utils.git_remote_branches('https://github.com/elifesciences/builder')
+        self.assertIsInstance(branches, list)
+        self.assertGreaterEqual(len(branches), 1)
+        self.assertIn('master', branches)
+
+    @patch('sys.stderr')
```
lsh-0: if we're mocking `sys.stderr` shouldn't we be testing what is captured rather than the return value of `utils.errcho` ?

src/tests/test_utils.py
```diff
@@ -21,3 +35,33 @@ def test_rmval(self):
...
+    def test_uin(self, get_input):
+        value = utils.uin('project')
+        self.assertEqual('lax', value)
+
+    @patch('utils.get_input', return_value='')
+    def test_uin_default(self, get_input):
+        value = utils.uin('project', default='lax')
+        self.assertEqual('lax', value)
+
+    def test_mkdir(self):
```
lsh-0: if this causes a side effect that will fail the test, this should be noted

src/utils.py
```diff
@@ -156,35 +86,16 @@ def uin(param, default=0xDEADBEEF):
...
-        
-def updatein(data, path, newval):
-    path_bits = path.split('.')
-    bit, rest = path_bits[0],path_bits[1:]
-    if not rest:
-        # we've come to the end of the path
-        data[bit] = newval
-        return newval
-    return updatein(data[bit], ".".join(rest), newval)
+def get_input(message):
```
lsh-0: why are we wrapping this?

giorgiosironi: because `mock` can only patch functions on a module, not global variable; so we can mock `utils.get_input` but not `raw_input`

PR #20, opened 2016-07-22
What was needed to bring up a new master

giorgiosironi:
- Configuration for us-east-1 projects
- Deduce region from stackname whenever the latter is passed as an argument
- Improved error messages and sanity checks

### DIFF
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -125,9 +125,8 @@ def stack_pem(stackname, die_if_exists=False, die_if_doesnt_exist=False):
     # for when we really need it to exist
     if die_if_doesnt_exist and not os.path.exists(expected_key):
         raise EnvironmentError("keypair %r not found at %r" % (stackname, expected_key))
-    # for when we don't want to accidentally overwrite it
     if die_if_exists and os.path.exists(expected_key):
-        raise EnvironmentError("keypair %r found at %r" % (stackname, expected_key))
+        raise EnvironmentError("keypair %r found at %r, don't want to accidentally overwrite it" % (stackname, expected_key))
     return expected_key
 
 @contextmanager
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -126,7 +126,7 @@ def stack_pem(stackname, die_if_exists=False, die_if_doesnt_exist=False):
     if die_if_doesnt_exist and not os.path.exists(expected_key):
         raise EnvironmentError("keypair %r not found at %r" % (stackname, expected_key))
     if die_if_exists and os.path.exists(expected_key):
-        raise EnvironmentError("keypair %r found at %r, don't want to accidentally overwrite it" % (stackname, expected_key))
+        raise EnvironmentError("keypair %r found at %r, not overwriting." % (stackname, expected_key))
     return expected_key
 
 @contextmanager
--- a/src/buildercore/project/files.py
+++ b/src/buildercore/project/files.py
@@ -44,7 +44,7 @@ def all_projects(project_file): #, project_file=config.PROJECT_FILE):
     allp = utils.ordered_load(open(project_file))
     if allp is None:
         return ({}, [])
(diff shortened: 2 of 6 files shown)


src/buildercore/core.py
```diff
@@ -125,9 +125,8 @@ def stack_pem(stackname, die_if_exists=False, die_if_doesnt_exist=False):
     # for when we really need it to exist
     if die_if_doesnt_exist and not os.path.exists(expected_key):
         raise EnvironmentError("keypair %r not found at %r" % (stackname, expected_key))
-    # for when we don't want to accidentally overwrite it
     if die_if_exists and os.path.exists(expected_key):
-        raise EnvironmentError("keypair %r found at %r" % (stackname, expected_key))
+        raise EnvironmentError("keypair %r found at %r, don't want to accidentally overwrite it" % (stackname, expected_key))
```
lsh-0: `"keypair %r found at %r, not overwriting."`

src/buildercore/project/files.py
```diff
@@ -42,6 +42,8 @@ def write_project_file(new_project_data, project_file):
 @testme
 def all_projects(project_file): #, project_file=config.PROJECT_FILE):
     allp = utils.ordered_load(open(project_file))
+    assert allp is not None, ("Project file %s seems to be empty" % project_file)
+    assert "defaults" in allp, ("Project file %s does not seem to have a `default` key" % project_file)
```
lsh-0: this is good

src/buildercore/project/files.py
```diff
@@ -42,6 +42,8 @@ def write_project_file(new_project_data, project_file):
 @testme
 def all_projects(project_file): #, project_file=config.PROJECT_FILE):
     allp = utils.ordered_load(open(project_file))
+    assert allp is not None, ("Project file %s seems to be empty" % project_file)
```
lsh-0: I would suggest if there is an empty file to just return empty values: `return ({}, [])`

however, if there is a non-empty project file *without* a defaults section, then die.

src/deploy.py
```diff
@@ -28,6 +28,7 @@ def impose_ordering(branch_list):
 @echo_output
 def deploy(pname, cluster=None, branch='master'):
     pdata = project.project_data(pname)
+    assert pdata['repo'], ('The `deploy` task needs a `repo` to be specified in the project %s configuration' % pname)
```
lsh-0: this case should be handled in the `@requires_branch_deployable_project` - the logic should proceed with the assumption that a project with a valid repo has been supplied. if that decorator is letting these projects past, then it's broken.

giorgiosironi:
See new commits (and their messages)

lsh-0:
re: `elife-us.yaml` - I don't think this file should be part of builder.

if you want to do specific overrides of elife.yaml settings, put them in a file called `./projects/elife/foo.yaml`.

their contents should look like:
```
defaults:
    aws:
        region: us-east-1

elife-bot:
    ami: ami-foobarbaz
```

src/decorators.py
```diff
@@ -51,16 +51,15 @@ def wrap1(func):
...
+            project_list = project.filtered_projects(filterfn)
+            if not pname or not pname.strip() or pname not in project_list:
                 pname = utils._pick("project", sorted(project_list), default_file=deffile('.project'))
             return func(pname, *args, **kwargs)
         return wrap2
     return wrap1
 
 #pylint: disable=invalid-name
-requires_branch_deployable_project = requires_filtered_project(lambda pname, project: project.has_key('repo'))
+requires_branch_deployable_project = requires_filtered_project(lambda pname, project: project.has_key('repo') and project.get('repo'))
```
lsh-0: so - in this case, the project is specifying a `repo`, but the `repo` value evaluates to None? changing this to simply project.get_key

lsh-0:
for the record, this merge reproduces an issue with the project merging. the ./projects/elife/us.yaml may be deleted later.

PR #21, opened 2016-07-22
optional shellcheck support and script linting

lsh-0:
added optional dependency on 'shellcheck' that lints the scripts in and ./scripts. 
scripts linted.

I feel it could do a better job in some cases. I'm not sure of the benefits of supporting sh+bash+dash+ksh but the changes were minimal

### DIFF
--- a/.activate-venv.sh
+++ b/.activate-venv.sh
@@ -21,5 +21,5 @@ fi
 # 'ARCHFLAGS' fixes a problem with OSX refusing to compile a dependency
 export ARCHFLAGS="-Wno-error=unused-command-line-argument-hard-error-in-future"
 
-source venv/bin/activate
+. ./venv/bin/activate
 pip install -r requirements.txt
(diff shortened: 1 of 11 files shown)


.activate-venv.sh
```diff
@@ -21,5 +21,5 @@ fi
 # 'ARCHFLAGS' fixes a problem with OSX refusing to compile a dependency
 export ARCHFLAGS="-Wno-error=unused-command-line-argument-hard-error-in-future"
 
-source venv/bin/activate
+. ./venv/bin/activate
```
giorgiosironi: did it explain the actual difference between `source` and `.`?

lsh-0: so (iirc) `source` is a *bash* language construct! if you do a `which source` it won't be able to find it on most systems. the `.` in `sh` has the same semantics and both `bash` and `sh` support the  `.`. I also recall Graham having a problem with `source` once upon a time on windows, but windows isn't really supported (yet).

PR #22, opened 2016-07-26
Removing multiple find_region versions and use the most popular one

giorgiosironi:


lsh-0:
not going to merge this as-is. the reason there are two is because one is a hard fail and the other is a kinder failure with a prompt to pick one. the majority of the logic is duplicated, but the code in buildercore doesn't (and shouldn't, ever) rely on user input. I would suggest keeping the buildercore `find_region`, catching the exception when more > 1 found and wrapping it in a `_pick`

giorgiosironi:
Updated as specified

lsh-0:
this looks good. if you could resolve the conflicts I'm happy to merge

giorgiosironi:
Rebased on master to solve conflicts, now the build fails but I believe in the same way as master was, thus making no negative difference.

PR #25, opened 2016-07-27
Porting buildvars.fix from old builder

giorgiosironi:
It may be a side-effect of the other bugs on keys stopping the bootstrap, but the api-dummy--* machines were created without build vars files, so I had to fix them.


giorgiosironi:
We used this with @thewilkybarkid just now, works fine.

PR #27, opened 2016-07-28
Avoid additional unnecessary symlinks

giorgiosironi:
At the first run:
```
/srv/salt -> /project/salt
```
but at the second run:
```
/srv/salt/salt -> /project/salt
```
gets created. Since `/srv/salt` is `/project/salt`, this created a
leftover file in the formula repo called `salt/salt`.

The same happens for `/srv/pillar -> /project/salt/pillar`, the leftover file is `salt/pillar/pillar`.

For the solution, see:
http://blog.endpoint.com/2009/09/using-ln-sf-to-replace-symlink-to.html
one never stops learning new Unix subleties.

### DIFF
--- a/scripts/init-vagrant-minion.sh
+++ b/scripts/init-vagrant-minion.sh
@@ -16,13 +16,15 @@ if [ ! -d /vagrant/cloned-projects/builder-base-formula/.git ]; then
     git clone ssh://git@github.com/elifesciences/builder-base-formula /vagrant/cloned-projects/builder-base-formula
 fi
 
-# project's `salt` file is mounted at `/srv/salt/` within the guest
+# project's `salt` file is mounted at `/project` and symlinked at `/srv/salt`
+# and `/srv/pillar` within the guest
 # by default the project's top.sls and pillar data is disabled by file naming.
 # hook that up now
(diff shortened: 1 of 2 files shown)


scripts/init-vagrant-minion.sh
```diff
@@ -16,13 +16,15 @@ if [ ! -d /vagrant/cloned-projects/builder-base-formula/.git ]; then
...
-# project's `salt` file is mounted at `/srv/salt/` within the guest
+# project's `salt` file is mounted at `/project` and symlinked at `/srv/salt`
+# and `/srv/pillar` within the guest
 # by default the project's top.sls and pillar data is disabled by file naming.
 # hook that up now
 cd /srv/salt/ && ln -sf example.top top.sls
-if [ ! -e /srv/pillar ]; then 
-    ln -sf /srv/salt/pillar/ /srv/pillar
-fi
+# this seems redundant wrt bootstrap.sh, which creates /srv/pillar?
```
lsh-0: looks like it. please remove the commented out lines. these scripts have enough noise already

I did some shuffling until almost all of the bootstrap logic now lives in these three files (bootstrap, init-vagrant-minion, init-master) and the common stuff has been split out into update-master and highstate.

PR #28, opened 2016-08-01
remove what appears to be an errant `-`

IanMulvany:


lsh-0:
I don't think it was errant, it was one of YAMLs strange list-of-single-dict. anyway - that declaration has been removed now

PR #30, opened 2016-08-02
prep_stack should leave the stack without a Salt master key

giorgiosironi:
Deployed and delete `api-dummy--fakeenv` to test this, solves the problem of `/etc/salt/pki/minion/minion_master.pub` interrupting the setup phase.


giorgiosironi:
Merging to avoid seeing the same error again

PR #31, opened 2016-08-03
Selecting aws-alt automatically using cluster name

giorgiosironi:
In the new builder formulas `master` is the default branch for every
project, so they come up by default with the alternate configuration
(including RDS), which is undesirable.

By selecting the alternate configuration using the cluster name instead,
we can specify to use RDS only in `end2end` and `prod`, for example.

Alternate configurations that do not match a cluster name, such as
`large`, can still be selected manually using `aws_launch_instance`
instead of `deploy`.

### DIFF
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -48,13 +48,16 @@ def build_context(pname, **more_context):
         # when this was first introduced, instance_id was synonmous with stackname
         'instance_id': None, # must be provided by whatever is calling this
 
-        'rds_dbname': None, # generated from the instance_id
-        'rds_username': 'root', # could possibly live in the project data, but really no need.
-        'rds_password': utils.random_alphanumeric(length=32), # will be saved to buildvars.json
 
         'branch': project_data['default-branch'],
         'revision': None, # may be used in future to checkout a specific revision of project
     }
+
+    if 'rds' in project_data['aws']:
+        defaults['rds_dbname'] = None # generated from the instance_id
+        defaults['rds_username'] = 'root' # could possibly live in the project data, but really no need.
+        defaults['rds_password'] = utils.random_alphanumeric(length=32) # will be saved to buildvars.json
+
     context = copy.deepcopy(defaults)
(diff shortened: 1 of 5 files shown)


giorgiosironi:
As an aside, we are now building PRs on Alfred. However, only one build can run at a given time, see http://ci--alfred.elifesciences.org/job/pull-requests-projects/job/builder/job/PR-31/1/console for what happens when two were concurrent. The `Jenkinsfile` modification mutually excludes them from running.

giorgiosironi:
Added the `fetch_cert` here as I have to stay on this branch to use builder for creating new instances.

giorgiosironi:
Piling up some small commits that I need to be on top of this branch to work.

src/buildercore/cfngen.py
```diff
@@ -48,13 +48,16 @@ def build_context(pname, **more_context):
...
 
-        'rds_dbname': None, # generated from the instance_id
-        'rds_username': 'root', # could possibly live in the project data, but really no need.
-        'rds_password': utils.random_alphanumeric(length=32), # will be saved to buildvars.json
 
         'branch': project_data['default-branch'],
         'revision': None, # may be used in future to checkout a specific revision of project
     }
+
+    if 'rds' in project_data['aws']:
```
lsh-0: something to this effect already lives in the mainline

giorgiosironi: I think that the recipes like https://github.com/elifesciences/lax-formula/blob/master/salt/lax/init.sls#L85 only rely on `cfn.outputs.RDSHost` to decide whether to connect their project to RDS, but I'm not sure.
Notice that originally these values were not `None`, but there was a fake username and password being generated but not used; I'll modify this to add them as `None`.

lsh-0: > rely on cfn.outputs.RDSHost to decide whether to connect their project to RDS, but I'm not sure.

if they're doing something other than that, then it's a bug. fwiw I don't recall using anything other.

PR #32, opened 2016-08-09
An option can be passed to turn off ssh forwarding

giorgiosironi:
This is useful to test what a remote machine does without polluting its keys with the developer's local stuff. Example:
```
./bldr ssh:elife-alfred--ci,forward_agent=False
```
The name of the option is the same as the one used by Fabric in its
settings, for consistency.

### DIFF
--- a/src/cfn.py
+++ b/src/cfn.py
@@ -5,6 +5,7 @@ import aws, utils
 from decorators import requires_project, requires_aws_stack, requires_steady_stack, echo_output, setdefault, debugtask
 import os
 from os.path import join
+from distutils.util import strtobool
 from buildercore import core, cfngen, utils as core_utils, bootstrap, project, checks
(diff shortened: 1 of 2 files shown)


giorgiosironi:
Not ready, debugging it.

giorgiosironi:
Ok now fixed.

lsh-0:
I'd much rather we removed the use of forward agent entirely and tested to ensure everything still works, which I think it may do. I suspect I put it in there when I was attempting to clone builder-private without a deploy token.

This just doesn't seem like a useful option to have.

giorgiosironi:
Yes #34 makes this out-of-date. Will cherry-pick only the useful stuff.

giorgiosironi:
Just cherry-picked the removal of commented code of b93d3acc41ff451e38fe4309d901898a70fb1d9a and closing this.

PR #33, opened 2016-08-09
isolates and fixes the bug with alt-config project configurations and default overrides

lsh-0:


src/cfn.py
```diff
@@ -150,6 +150,11 @@ def download_file(stackname, path, destination):
 @requires_aws_stack
 def upload_file(stackname, local_path, remote_path, overwrite=False):
     with stack_conn(stackname):
+        print 'stack:',stackname
```
giorgiosironi: sure you want to let the debug statements in?

PR #34, opened 2016-08-09
Do not forward ssh agent during update_stack

giorgiosironi:
Both the deploy and launch command work correctly without it now.
Also, there was a method which appears to be dead code

### DIFF
--- a/src/buildercore/bootstrap.py
+++ b/src/buildercore/bootstrap.py
@@ -95,26 +95,6 @@ def create_stack(stackname):
         keypair.delete_keypair(stackname)
         raise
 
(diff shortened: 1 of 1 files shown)


lsh-0:
Debatable if it's dead code. I did deliberately orphaned it in the move to the new builder. 

Cloudfront allows you to 'update' the stack template you gave it during creation. It looks at the new template you give it and the one used to create it and does the necessary changes. This may or may not involve downtime of the various bits of infrastructure. The updates are often not quick...

I've found it easier to just create and delete stacks.

hrm.

yah. keep it deleted.

PR #35, opened 2016-08-10
Removing ssh forwarding

giorgiosironi:
The last place where it seems to be used is in the manual ssh commands. These aren't used by many people (on the new builder, they use it on the old one) nor by automated builds, so we aren't likely to break stuff

### DIFF
--- a/src/cfn.py
+++ b/src/cfn.py
@@ -127,7 +127,7 @@ def aws_stack_list():
 def ssh(stackname, username=DEPLOY_USER):
     public_ip = core.stack_data(stackname)['instance']['ip_address']
     # -A forwarding of authentication agent connection
-    local("ssh %s@%s -A" % (username, public_ip))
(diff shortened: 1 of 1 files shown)


giorgiosironi:
One thing to be aware of is that `./bldr update_master` may fail for some masters because `git pull` in `/opt/builder-private` was working before thanks to the agent. `master-server--2016-07-22` had this problem, so I regenerated its `/root/.ssh/id_rsa` and added as deploy key of `builder-private`.

giorgiosironi:
However, the same problem is true for `crm-formula` and I cannot add it as a deploy key there because Github refuses it saying it is already in use. I will have to add it to a user, which user?

giorgiosironi:
(for the moment I'll add it to myself, but when we know more let me know where should I move it)

lsh-0:
are we limited in our use of deploy keys maybe?

PR #36, opened 2016-08-10
Test settings for elife-anonymous-user

giorgiosironi:
From another AWS account, but no sensitive data.


giorgiosironi:
Tested by using the anonymous AWS account.

PR #37, opened 2016-08-10
Accepting commands such as ./bldr launch:master-server,some_instance_id

giorgiosironi:


giorgiosironi:
Needed for automating `launch` in builds, which require user interaction.

PR #38, opened 2016-08-10
Option to use owner_ssh key for download_file

giorgiosironi:
To use in extreme cases, like downloading the key of a failed master server to set it up for Github's builder-private repository

### DIFF
--- a/.lint.sh
+++ b/.lint.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 set -e
 # E1129=doesn't like Fabric's context manager
-pylint -E src/* salt/salt/_modules/* --disable=E1129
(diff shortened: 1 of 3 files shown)


giorgiosironi:
Will sort out tomorrow why Pylint is angry.

giorgiosironi:
https://docs.python.org/2/distutils/apiref.html#distutils.util.strtobool the function seems to be there for our Python version

giorgiosironi:
Pylint problem solved.

giorgiosironi:
Need this for the more precise lint, to fix the build currently being red because of a `distutils` call:
http://ci--alfred.elifesciences.org/job/test-builder/114/console

PR #40, opened 2016-08-11
Creating SNS topics, SQS queues, and connecting them

giorgiosironi:
## Project configuration

Every project's `aws` key now has (non-exhaustive list):
- `ec2`: just a `bool` for now, but it will become a dictionary with all the keys like `ami` moved inside. Didn't want to do everything in a single shot, this is already huge.
- `rds`: `None` or the requested RDS configuration
- `sns`: list of SNS topics to create.
- `sqs`: dictionary of SQS queues. Each `subscriptions` key of a queue is the list of topics it should subscribe to.

Furthermore, SNS and SQS names can contain the `{cluster}` placeholder, which will be replaced with the cluster name, for example creating the topic `articles-end2end` and `articles-prod`.

## Phases

1. Generate a `context` that can be dumped, replacing template of names (`articles-{cluster}`) and generating names where necessary (`rds_password`)
2. Generate a Cloud Formation template (`cfngen` and `trop` module) with SQS and SNS queues.
3. Create the stack on AWS (`bootstrap.create_stack`)
4. Perform setup of the resources (`bootstrap.setup_*` functions)
5. Update the stack the first time (`bootstrap.update_stack`)

`1` has to store also the generated context:
```
$ ls .cfn/contexts/ -l
total 12
-rw-rw-r-- 1 giorgio giorgio 3552 Aug 26 11:10 api-dummy--fakeenv.json
-rw-rw-r-- 1 giorgio giorgio 3649 Aug 26 09:49 journal-cms--end2end.json
-rw-rw-r-- 1 giorgio giorgio 1216 Aug 26 11:55 search--end2end.json
```
`3` consists of `prep_stack` for EC2 resources, but also of subscribing to topics for SQS resources instead. `4` only has something to do for EC2 resources.

## Projects using it

- `bus` is the project everyone will depend upon, containing just the SNS topics.
- `search` is the first project, defining a single SQS queue subscribing to a topic.


lsh-0:
> 1 has to store also the generated context
It may become obvious as I go through the changes, but what is the reason for this?

projects/elife.yaml
```diff
@@ -363,3 +371,48 @@ search:
...
+
+bus:
+    default-branch: null
+    domain: null
+    int_domain: null
+    aws:
+        ec2: false
+        sns:
+            # bus-articles-prod, bus-articles-end2end, etc.
+            - bus-articles--{cluster}
```
lsh-0: I was initially against this placeholder, but it's growing on me. The flexibility it provides will be important no doubt. Not sure if we should go with `cluster` though. `env` perhaps?

giorgiosironi:
Why storing the context: there are additional setup actions to perform after the Cloud Formation template is in the `CREATE_COMPLETE` state, in this case for SQS. The setup actions is to connect the queues with the relevent SNS topics, which I could not find a way to do in Cloud Formation without dropping and recreating the SNS topic itself (which is in another project, `bus`).

We could avoid storing the context if we passed it around up to `buildercore.bootstrap.create_stack()`, but that seemed to complicate things even more.

projects/elife.yaml
giorgiosironi: Choose a name and I will change it, no preference on how it's named as long as it is consistent. The format is that of Python's `string.format` method.

lsh-0: lets go with `instance`. I'll make the change.

giorgiosironi: uhm that smells if we want to extend the format to `journal--prod--1` for multiple machines later

lsh-0: ok, so a stackname looks like:
<project>--<instanceid>[--clusterid]

so: lax--prod, lax--end2end, lax--2015-01-01, lax--foo

and in a load balanced setup:
lax--prod--1, lax--prod--2, lax--prod--3

would you have:
bus-articles--1 or bus-articles--end2end ?

giorgiosironi: I would have `bus-articles--end2end` because this is an SNS topic, so it's a singleton; the same goes for a queue like `queue-name--end2end` which may be shared by two EC2 instances `search--end2end--1` and `search--end2end--2`.

lsh-0: I think we're on the same page here, just our vocabulary is unclear.

I'm going to go with `instance` and update any references to `cluster` that may have slipped in.

lsh-0:
I'll bring up an instance or two with this PR tomorrow to satisfy myself and then I'll merge it.

lsh-0:
testing a lax rds instance now

lsh-0:
lax rds built successfully, merging.

PR #41, opened 2016-08-23
Map Medium port on Vagrant

thewilkybarkid:


projects/elife.yaml
```diff
@@ -336,6 +336,8 @@ medium:
     formula-repo: https://github.com/elifesciences/medium-formula
     vagrant:
         ram: 1024
+        ports:
+            1243: 80
```
giorgiosironi: is there a pattern with which we choose the magic numbers such as 1243?

giorgiosironi:
:+1:

projects/elife.yaml
thewilkybarkid: Previous one + 1.

PR #42, opened 2016-08-23
Ability to specify only one DNS record

giorgiosironi:
Useful for backends, where we don't want to expose them with a public DNS but only make them accessible from api-gateway


giorgiosironi:
Works good for the `medium` project, merging before I forget to checkout the branch before creating a new instance.

PR #45, opened 2016-08-30
Keeping track of agent forwarding change

giorgiosironi:
Keeping track of https://github.com/elifesciences/builder/commit/8c1138f43759e620109e2a4a5b55a38c18c6b052

If the master servers' keys are added to the `elife-master-builder` user, this forwarding should become unnecessary.


giorgiosironi:
There are no more masters on the new builder apart from `master-server--2016-07-22` which has this setup. Therefore merging, let's keep an eye on possible master updates failing.

PR #46, opened 2016-09-02
Adding a canonical DNS to production instances

giorgiosironi:
They will have `ExtDNS` but also `ExtDNSProd` which will be used as the `DomainName`. They remain reachable as `prod--project.example.org`, but become also reachable as `project.example.org`.

### DIFF
--- a/src/buildercore/trop.py
+++ b/src/buildercore/trop.py
@@ -30,6 +30,7 @@ DBSUBNETGROUP_TITLE = 'AttachedDBSubnet'
 EXT_TITLE = "ExtraStorage"
 EXT_MP_TITLE = "MountPoint"
 R53_EXT_TITLE = "ExtDNS"
+R53_EXT_PROD_TITLE = "ExtDNSProd"
 R53_INT_TITLE = "IntDNS"
 
 KEYPAIR = "KeyName"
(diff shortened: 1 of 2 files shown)


lsh-0:
This is related to the PR45 (the big one) I merged in yesterday. If you update your branch with the changes in master it should fix this problem.

```
13:49:04   File "/var/lib/jenkins/workspace/pull-requests-projects/builder/PR-46/src/buildercore/cfngen.py", line 68, in build_context
13:49:04     assert context['stackname'] != None, "a stackname wasn't provided."
13:49:04 AssertionError: a stackname wasn't provided.```

**Comment by giorgiosironi** [comment]
`lax--prod` is now a thing, so this can be reviewed and possibly merged. Any thoughts?

**Comment by lsh-0** [comment]
Just so the intent of this PR is clear:

for `--prod` instances, the default DNS records are the *canonical* project domains (`lax.elifesciences.org`) rather than the project *instance* domain (`prod--lax.elifesciences.org`), however the 'project instance' domain records are still created.

is this correct?

**Comment by giorgiosironi** [comment]
Yes, and: `DomainName` in the outputs becomes `lax.elifesciences.org` for them, which is useful for things like http->https redirects to use the most common name.

**Comment by giorgiosironi** [comment]
After having done multiple DNS for CloudFront I feel this is the wrong solution, we should do:
```
- ext_dns:
    - {instance}--lax
    - prod
```
in `aws-alt`

lsh-0:
thank you for updating this issue. I felt a little guilty ignoring it all this time.

PR #49, opened 2016-09-12
Clusters of ec2 instances

giorgiosironi:
Tentative specification:
- we want multiple nodes serving a project, e.g. `journal` will need to scale out and balance traffic between the instances.
- All these nodes should be identical in software, code version and configuration.
- Nodes can be distinguished through a numerical id to suffix to the stackname, which can be used as the Salt minion id.

The first step is to implement a fixed number of copies of the same server. These nodes do not have an actual (internal nor external) hostname which will be adopted by a load balancer when there is one.

This code is able to create two node into a `medium--clustered` stackname (could have been called `medium--prod`).


src/buildercore/core.py
```diff
@@ -111,13 +111,23 @@ def connect_aws_with_stack(stackname, service):
     return connect_aws_with_pname(pname, service)
 
 def find_ec2_instance(stackname):
-    "returns ec2 instance data for a *specific* stackname"
+    "returns list of ec2 instances data for a *specific* stackname"
     # filters: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html
-    kwargs = {
+    filter_by_cluster = {
```
giorgiosironi: I'm not sure what is the style convention here for multi-line nested dictionaries, I am following what is already there

src/buildercore/core.py
```diff
@@ -111,13 +111,23 @@ def connect_aws_with_stack(stackname, service):
...
+            'tag:Cluster':[stackname],
+            'instance-state-name': ['running']}}
+    filter_by_name = {
         'filters': {
             'tag:Name':[stackname],
             'instance-state-name': ['running']}}
-    return connect_aws_with_stack(stackname, 'ec2').get_only_instances(**kwargs)
+    # TODO: may need a fallback
+    conn = connect_aws_with_stack(stackname, 'ec2')
+    cluster_result = conn.get_only_instances(**filter_by_cluster)
```
giorgiosironi: clunky but it's just a fallback, there may be a better Pythonic way

src/buildercore/trop.py
```diff
@@ -94,44 +96,58 @@ def rds_security(context):
...
         ec2.Tag('Owner', context['author']),
+        # hierarchically ordered
         ec2.Tag('Project', context['project_name']),
+        ec2.Tag('Cluster', context['stackname']),
     ]
-
-def ec2instance(context):
+    if node:
+        tags.append(ec2.Tag('Name', '%s--%d' % (context['stackname'], node)))
+        tags.append(ec2.Tag('Node', '%s' % node))
```
giorgiosironi: seems that even `Node` has to be a string according to Troposphere code

src/buildercore/trop.py
```diff
@@ -94,44 +96,58 @@ def rds_security(context):
...
+    return tags
+
+def ec2instance(context, node):
     lu = partial(utils.lu, context)
+    build_vars = dict(context)
+    build_vars['node'] = node
+    build_vars['nodename'] = "%s--%s" % (context['stackname'], node)
+    # the above context will reside on the server at /etc/build-vars.json.b64
+    # this gives Salt all (most) of the data that was available at template compile time.
+    build_vars_serialization = base64.b64encode(json.dumps(build_vars))
```
giorgiosironi: there was a large if for adding build vars to context only in the case of EC2 instances being there. Moving the serialization to base64 here avoids adding to `context` a `build_vars` key containing the encoding of the whole `context` again. Also, it's naturally executed only on EC2 instances given the function we are in.

src/buildercore/trop.py
```diff
@@ -230,16 +246,17 @@ def render(context):
...
-
         template.add_resource(secgroup)
-        template.add_resource(instance)
+
+        for node in range(1, context['ec2']['cluster-size'] + 1):
+            instance = ec2instance(context, node)
+            template.add_resource(instance)
+            cfn_outputs.extend(_ec2_outputs(node))
 
         template.add_parameter(Parameter(KEYPAIR, **{
```
giorgiosironi: the keypair is identical for all nodes in the cluster

src/cfn.py
```diff
@@ -189,12 +190,15 @@ def _user(use_bootstrap_user):
...
 @task
 @requires_aws_stack
-def cmd(stackname, command=None):
+def cmd(stackname, command=None, username=DEPLOY_USER):
     if command is None:
         abort("Please specify a command e.g. ./bldr cmd:%s,ls" % stackname)
-    with stack_conn(stackname):
-        with settings(abort_on_prompts=True):
-            run(command)
+    print "Connecting to: %s" % stackname
```
giorgiosironi: In places where multiple instance support is needed, usages of `with stack_conn(stackname)` become calls to `stack_all_ec2_nodes` passing a function to repeatedly execute.

src/buildercore/core.py
```diff
@@ -142,22 +152,45 @@ def stack_pem(stackname, die_if_exists=False, die_if_doesnt_exist=False):
 def stack_conn(stackname, username=config.DEPLOY_USER, **kwargs):
     if 'user' in kwargs:
         LOG.warn("found key 'user' in given kwargs - did you mean 'username' ??")
-    data = stack_data(stackname)
+
+    data = stack_data(stackname, ensure_single_instance=True)
```
giorgiosironi: fails fast if being used over a cluster with multiple EC2 instances

src/deploy.py
```diff
@@ -48,3 +49,10 @@ def deploy(pname, instance_id=None, branch='master'):
 
     bootstrap.create_update(stackname)        
     setdefault('.active-stack', stackname)
+
```
giorgiosironi: Putting this here to avoid a mutual dependency between `buildvars` and `buildercore.bootstrap`. This way the dependencies are `buildercore.bootstrap -> buildvars`, and `deploy -> buildvars`

giorgiosironi:
This code correctly works, but I am getting up the coverage to the required 65% (currently 63%), which means I am adding tests that create dummy instances for real.

giorgiosironi:
Finally to 65% coverage, ready for review.

src/buildercore/core.py
lsh-0: I don't care? It's just data. I'm sure there is some pep8 rule, but I really just don't care

src/aws.py
```diff
@@ -33,27 +33,6 @@ def stack_list(region=None):
 #
 #
 
-@task
```
lsh-0: you got rid of the `create_ami` task? granted we haven't used it in the new builder, but should still work ...

src/buildercore/bootstrap.py
```diff
@@ -19,6 +19,7 @@
 from boto.exception import BotoServerError
 from kids.cache import cache as cached
 from buildercore import cfngen
+import buildvars
```
lsh-0: is this the `buildvars.py` from `src/` ? if so, it doesn't belong here. refactoring is needed. 
nothing inside `buildercore` relies on the interface code in it's parent.

src/buildercore/core.py
```diff
@@ -111,13 +107,22 @@ def connect_aws_with_stack(stackname, service):
     return connect_aws_with_pname(pname, service)
 
 def find_ec2_instance(stackname):
-    "returns ec2 instance data for a *specific* stackname"
+    "returns list of ec2 instances data for a *specific* stackname"
```
lsh-0: this comment is misleading if the instance isn't part of a cluster - it's not going to return a list.

`"returns ec2 instance data for a specific stackname, or a list of ec2 instance data if clustered"`

src/buildercore/core.py
```diff
@@ -111,13 +107,22 @@ def connect_aws_with_stack(stackname, service):
...
+        'filters': {
+            'tag:Cluster':[stackname],
+            'instance-state-name': ['running']}}
+    filter_by_name = {
         'filters': {
             'tag:Name':[stackname],
             'instance-state-name': ['running']}}
-    return connect_aws_with_stack(stackname, 'ec2').get_only_instances(**kwargs)
+    conn = connect_aws_with_stack(stackname, 'ec2')
+    cluster_result = conn.get_only_instances(**filter_by_cluster)
```
lsh-0: does this mean the normal case is going to be two requests to get instance data? 
can this be determined before the requests are made? or figure out a better filter, or ... ?

src/buildercore/core.py
```diff
@@ -233,26 +260,35 @@ def describe_stack(stackname):
...
         # a CloudFormation's outputs go stale! because we can't trust the data it
         # gives us, we sometimes take it's instance-id and talk to the instance directly.
-        #inst_id = data['indexed_output']['InstanceId']
-        #inst = get_instance(inst_id)
-        stacks = find_ec2_instance(stackname)
-        assert len(stacks) == 1, ("while looking for %s, found these stacks: %s" % (stackname, stacks))
-        inst = stacks[0]
-        data['instance'] = inst.__dict__
+        ec2_instances = find_ec2_instance(stackname)
+        assert len(ec2_instances) >= 1, ("while looking for %s, found no ec2 instances" % stackname)
```
lsh-0: `"no ec2 instances found for %r" % stackname`

src/buildercore/core.py
```diff
@@ -233,26 +260,35 @@ def describe_stack(stackname):
...
+        if len(ec2_instances) == 1:
```
lsh-0: this section is gross. if `find_ec2_instance` always returned a list of ec2 instances rather than one or the other, we could treat this simply. I'll make this change.

src/buildercore/trop.py
lsh-0: You're creating a tag called `Node` .. does `Node` have some other special AWS meaning?

src/buildercore/trop.py
```diff
@@ -94,44 +97,58 @@ def rds_security(context):
...
+    if node:
+        tags.append(ec2.Tag('Name', '%s--%d' % (context['stackname'], node)))
+        tags.append(ec2.Tag('Node', '%s' % node))
+    else:
+        tags.append(ec2.Tag('Name', context['stackname']))
+    return tags
+
+def ec2instance(context, node):
     lu = partial(utils.lu, context)
+    build_vars = dict(context)
```
lsh-0: what is context if not a `dict` here?

src/buildercore/trop.py
```diff
@@ -23,7 +25,8 @@
 osissuefn("embarassing code. some of these constants should be pulled form project config or given better names.")
 
 SECURITY_GROUP_TITLE = "StackSecurityGroup"
-EC2_TITLE = 'EC2Instance'
+EC2_TITLE = 'EC2Instance1'
```
lsh-0: is the suffix of `1` necessary?

src/deploy.py
lsh-0: this is a hack. see my comment about moving buildvars into buildercore

src/tasks.py
```diff
@@ -45,6 +47,28 @@ def ami_for_project(pname):
     # good for figuring out filters 
     #print results[0].__dict__
 
+@task
```
lsh-0: oh - you just moved the ami stuff here. um. was there a good reason for the move? it's not part of this builder, but there was some code for using `packer` to build vagrant boxes that may come back around

src/tests/test_buildercore_bootstrap.py
```diff
@@ -1,8 +1,13 @@
 from . import base
-from buildercore import bootstrap
+from buildercore import bootstrap, cfngen
+import cfn
 from mock import patch, call, MagicMock
 
 class TestBuildercoreBootstrap(base.BaseCase):
+    def __init__(self, *args, **kwargs):
+        super(TestBuildercoreBootstrap, self).__init__(*args, **kwargs)
+        self.stacknames = []
```
lsh-0: you don't need this. use the `self.setUp()` and `self.tearDown()` functions.

src/tests/test_decorators.py
```diff
@@ -0,0 +1,39 @@
+from . import base
+from mock import patch
+import decorators
+
+class TestDecorators(base.BaseCase):
+    def test_timeit_smoke_test(self):
+        @decorators.timeit
+        def some_task():
+            pass
+        some_task()
```
lsh-0: are we just aiming for code coverage? is there something here we can actually compare to an expected working state? perhaps have `some_task` sleep for a second and have the `timeit` decorator emit some value?

lsh-0:
ok - so, this is a huge PR and I'm 100% happy with the code. I'm going to:

* investigate the `find_ec2_instance` function and improve it. this may involve many smaller changes to functions expecting a data dict rather than a list of data dicts

* move buildvars into buildercore and attempt to preserve the current buildvars interface

src/tasks.py
giorgiosironi: mutual dependencies: keeping it in `aws` means `aws` has to import `bakery`, which triggered a circular dependency

giorgiosironi:
I'll go through the comments to answer them, will be able to get back on this after the work on the Mac

src/aws.py
giorgiosironi: as you already found out, only moved it

src/buildercore/bootstrap.py
giorgiosironi: the problem to solve is to be able to read the node name/number (`journal--prod--1` or just `1` or `2`). Maybe we can get it in another way, like with the EC2 Tag? If so, the build vars do not need to be ported here and we will avoid this dependency

giorgiosironi: see the minion_id variable in the middle of this file

src/buildercore/core.py
giorgiosironi: I recall that in the not-a-cluster case this returns a list of 1 element, which is the instance data; it was already used with `[0]` after the call

src/buildercore/core.py
giorgiosironi: we could invert the conditions so that the normal case is 1 instance, and it succeed on the first call. The backward compatibility is towards existing EC2 instances, which do not have the `Cluster` tag

src/buildercore/core.py
giorgiosironi: `find_ec2_instance` should already do that, but it is this method that has the two possible cases (dictionary for one, or list of dictionaries if more than one); that's the reason for the two branches. But if *this* method always returns a list, then it can be unified

lsh-0:
don't do any more on this today, I'll tweak it and then we can revisit it together

src/buildercore/trop.py
giorgiosironi: Reason: the code that creates the EC2 part of the template is parameterized with a for loop to create `EC2Instance%d`, so in the single instance case the DNS have to refer to `EC2Instance1`.

Then there is `ext`, the additional volume. That just needs to be extended to make multiple copies and be added to `EC2Instance%d`, but I didn't want to add even more changes so I started to support only multiple EC2 whose project does not have `ext`.

src/buildercore/trop.py
giorgiosironi: Apart from Node.js, I can't find usages from their docs.

src/buildercore/trop.py
giorgiosironi: we create `context in `cfngen`, so it should always be a dict (this does a copy so that we can add stuff to the single node's context.) Time for an assertion?

src/deploy.py
giorgiosironi: as wrote somewhere up in this PR, if we can build a unique minion id without using the build vars, that's great and we can remove this hack

src/tests/test_buildercore_bootstrap.py
giorgiosironi: correct, possibly delegating to `base`

src/tests/test_decorators.py
giorgiosironi: smoke test in the sense of "does not explode with runtime errors" :)
I think here a `@patch` can be used to capture the output and actually check it

lsh-0: sure, I'm all for boosting coverage, but I think the logic should be tested at the same time, otherwise it will be forgotten about and the high coverage will give us a false sense of security.

lsh-0:
ok, so, things still feel a little iffy, but I'm happier with some of the core changes now.

two things I'd like done before we merge this:

* fix up that `timeit` decorator test so we know it's emitting _some_ value
* ensure the stack creation/any long running tests are executed separately to the fast ones.

then I'm happy merging.

src/buildercore/bootstrap.py
```diff
@@ -252,12 +252,17 @@ def _update_ec2_node():
         install_master_flag = str(is_master).lower() # ll: 'true' 
         master_ip = master(region, 'private_ip_address')
 
+        # TODO: this is a little gnarly. I think I'd prefer this logic in the script:
+        #       if [ cat /etc/build-vars.json | grep 'nodename' ]; then ... fi
```
giorgiosironi: I am wondering how this works in Vagrant, do we have a fallback? I think we should handle the case of the build vars being missing at all

giorgiosironi: This code is not executed in Vagrant, right.

giorgiosironi: And the build vars should always be there on any EC2 instance.

giorgiosironi: also, I shall tweet this commit message

lsh-0: it was just pseudo-code ;) 
```
if build-vars exists
    minion_id = 
        open build-vars
        extract and return nodename
otherwise, 
    use given stackname
```

and I was just thinking out loud. We can leave it as-is for now. I put in the `# message` as a message for my future self.

src/tests/test_decorators.py
```diff
@@ -3,11 +3,19 @@
...
             pass
-        some_task()
+        some_task(42, option='value')
+        (args, _) = info.call_args
+        self.assertIsInstance(args[0], str)
+        self.assertEqual(args[1], 'some_task')
+        self.assertEqual(args[2], (42,))
+        self.assertEqual(args[3], {'option': 'value'})
+        self.assertIsInstance(args[4], float)
+        self.assertGreater(args[4], 0.0)
```
lsh-0: this line is all I was after, good to know `mock` can do all that other magic.

.run-tests.sh
```diff
@@ -1,15 +1,15 @@
...
+    modules="$*"
 fi
 
 export PYTHONPATH="src"
-green --run-coverage tests"$module"
+green --run-coverage $modules
 
 # only report coverage if we're running a complete set of tests
-if [ -z "$module" ]; then
+if [ "tests integrations_tests" = "$modules" ]; then
```
lsh-0: not sure if this line is behaving as expected, but it's not a deal-breaking to merging.

PR #51, opened 2016-09-16
Disparate but small integration testing improvements

giorgiosironi:


giorgiosironi:
Will merge this to avoid conflicts while expanding the integration tests with start and stop of instances.

PR #52, opened 2016-09-16
Adding an ELB on top of a cluster of EC2 nodes

giorgiosironi:
Elastic Load Balancers are a set of machines managed from AWS that divide traffic on a port to a set (possibly dynamic) of EC2 instances.

## Configurations
- new default values: `redundant-subnet-id` and `redundant-subnet-cidr`, only necessary if you want an ELB. They are created into two subnets corresponding to different availability zones, for redundancy if an availability zone has issues (EC2 nodes would need to do the same, but let's not get ahead of ourselves.)
- The `aws.elb` key decides whether an ELB should be created for this project. May be used only for certain environments, like `end2end` and `prod`. Just a `bool` now, but will naturally evolve into a dictionary.
- Moved `ami` into `aws.ec2` continuing the trend of isolating there what is necessary only for EC2 instances.

## DNS
When there is only one EC2 instance, it can have both internal (`*.elife.internal`) and external DNS entries (`*.elifesciences.org`).
However, ELB must be explicitly `internet-facing` or `internal`. Therefore the project as a whole must set `domain` or `intdomain` to `False` to choose to have only an internal or only an external DNS.
Moreover, an ELB has its own unreadable DNS entry created with the template (`internal-medium-c-ElasticL-7PNQ88Y406VJ-88688279.us-east-1.elb.amazonaws.com`). To that, we add another A record, which is aliased in Route53 to point to it (intentionally not a CNAME as aliasing is better integrated for ELB).

## Tasks
- `deploy` also creates the ELB as part of the template. Same for `launch` and `update`, this change is transparent to them.
- `ssh` and `owner_ssh` now work also on clusters. You can specify an optional `node` argument to decide which node to login into.

## Configuration options
- `ConnectionDrainingPolicy` let instances that are going to be disconnected finish serving their existing requests.
- `CrossZone` means even if the ELB nodes are in different availability zones, they will route traffic across all the EC2 nodes, no matter where they are.
- `Instances` is the fixed set of EC2 nodes at creation time at the moment, no autoscaling.
- `Listeners` specifies to balance `HTTP` on port `80` to the port `80` of the EC2 nodes.
- the default `HealthCheck` is a TCP connection on port 80, even if no GET can be made
- `SecurityGroups` is the same as the EC2 nodes being created.
- `Scheme` related to internal or external DNS.
- `Subnets` related to the availability zones.


giorgiosironi:
Test run:
```
elife@ci--medium:~$ for i in {1..1000}; do curl -v http://clustered--medium.elife.internal/medium-articles; done
```
(from this machine because medium is an internal project, not user-facing)
```
root@ip-10-0-2-22:/home/elife# wc -l /var/log/nginx/medium.access.log 
505 /var/log/nginx/medium.access.log
root@ip-10-0-2-70:/home/elife# wc -l /var/log/nginx/medium.access.log                
527 /var/log/nginx/medium.access.log
```
(additional requests were previous tests)

projects/elife.yaml
```diff
@@ -387,6 +392,7 @@ medium:
         clustered:
             ec2:
                 cluster-size: 2
+            elb: True
```
lsh-0: if the default for clusters of machines is for them to be behind an elb, this `elb: True` should be in `defaults` at the top.

lsh-0: wait, ignore me. `clustered` is the name of a alt-config.

giorgiosironi: yes, didn't find a better name, but it's temporary

src/cfn.py
```diff
@@ -125,18 +125,21 @@ def aws_stack_list():
...
-    "maintainence ssh. uses the pem key and the bootstrap user to login."
-    public_ip = core.stack_data(stackname, ensure_single_instance=True)[0]['instance']['ip_address']
+def owner_ssh(stackname, node="1"):
+    "maintenance ssh. uses the pem key and the bootstrap user to login."
+    public_ip = core.stack_data(stackname)[_node_index(node)]['instance']['ip_address']
     # -i identify file
     local("ssh %s@%s -i %s" % (BOOTSTRAP_USER, public_ip, stack_pem(stackname)))
 
+def _node_index(node_argument):
+    assert node_argument.isdigit, "You must pass a node number (greater or equal than 1)"
```
lsh-0: There is a function in `buildercore.utils` called `die` with a similar structure to the `assert` statement that won't be compiled away if optimisations are ever turned on. It raises a `RuntimeError` but perhaps should raise an `AssertionError`.

giorgiosironi: :+1: for changing this to a `die()` call

lsh-0:
I'm happy for this to be merged after Alfred passes everything.

PR #53, opened 2016-09-20
HTTPS support in ELB

giorgiosironi:
ELB support a certificate reference that points to our wildcard certificate being safely stored into AWS IAM.

Moreover, we configure other interesting parameters of the ELB like sticky sessions and the connection timeout, using them for `journal--end2end` which is currently being recreated.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -158,7 +158,7 @@ journal:
                 cluster-size: 2
             elb:
                 stickiness: True
-                procotol: https
+                protocol: https
     vagrant:
         ram: 4096
         ports:
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -63,6 +63,8 @@ defaults:
             # elb defaults only used if an 'elb' section present in project
             stickiness: False
             protocol: http
+            idle_timeout: 60
+            certificate: ASCAIRXYIRFBOR5QSDP5M
         sqs: []
         sns: []
     aws-alt:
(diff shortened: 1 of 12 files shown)


projects/elife.yaml
```diff
@@ -59,6 +59,12 @@ defaults:
             # external hdd
             size: 10 # GB
             device: /dev/sdh
+        elb:
+            # elb defaults only used if an 'elb' section present in project
+            stickiness: False
+            protocol: http
+            idle_timeout: 60
+            certificate: ASCAIRXYIRFBOR5QSDP5M
```
lsh-0: is this some kind of ID?

lsh-0:
I'm happy for this to be merged so long as that mysterious value I commented on isn't sensitive

projects/elife.yaml
lsh-0: perhaps give this a comment in the defaults?

`# id for ... ll: ABCDEFGHIJ123ABC5M`

giorgiosironi:
Incidentally, I had to replace the mysterious reference with its ARN for the ELB to correctly recognize it; so now it's much more readable.

lsh-0:
I was wondering about why the ARN wasn't being used. I'm pleased it is now.

Ready to merge?

giorgiosironi:
Almost, I am still missing testing cookie stickiness and have to rebuild the stack one more time.

giorgiosironi:
Added not only stickiness, but also buildvars updates to make them work on multiple nodes. When we deploy, we must change the buildvars on all EC2 instances and then run the update as normal. The `buildvars.*` tasks now act on all nodes too.

PR #54, opened 2016-09-22
Adding Environment tag to EC2 and similar instances

giorgiosironi:
This can be useful to filter which are prod or end2end or ci or demo instances


giorgiosironi:
Doesn't hurt for future filters.

PR #56, opened 2016-09-28
S3 to SQS notification support

giorgiosironi:
Still in progress, but the largest change is to start storing `.cfn/contexts` on S3 like for `.cfn/keypairs`. Contexts are loaded in more and more places, in this case updating the S3 notifications, so it makes sense to have them available also on builders that are not the one that created an instance.


giorgiosironi:
Ready for review.

giorgiosironi:
I need the `context` module developed here to add the DNS update after stopping/starting EC2 instances, so I'll merge soon after the build is green.

PR #59, opened 2016-09-30
Auto-update builder-base-formula for Vagrant VMs

giorgiosironi:
It seems there is no tasks that does this for a local builder while spawning Vagrant VMs, although it is done for the master server.


giorgiosironi:
Solving conflicts and merging.

PR #60, opened 2016-10-03
Adds stop, start and stop_if_next_hour_is_imminent tasks

giorgiosironi:
Sample commands:
`./bldr start:elife-bot--end2end`
`./bldr stop:elife-bot--end2end`
`./bldr stop_if_next_hour_is_imminent:elife-bot--end2end`

The first two act on EC2 stop/state commands, bringing each node in the stack to the state `stopping` and then `stopped`; or to the state `pending` and then `running`.

The command `stop_if_next_hour_is_imminent` will be used by Alfred in a periodical task to stop nodes that have been running when they have outlived their usefulness. Nodes will be lazily started by builds when needed, but will not run for more than one hour (which is the minimum EC2 billing after a `stopped->running` transition.)

### DIFF
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -293,7 +293,7 @@ def stack_data(stackname, ensure_single_instance=False):
         ec2_instances = find_ec2_instance(stackname)
 
         if len(ec2_instances) < 1:
-            raise RuntimeError("found no ec2 instances for %r" % stackname)
+            raise RuntimeError("found no running ec2 instances for %r" % stackname)
         elif len(ec2_instances) > 1 and ensure_single_instance:
             raise RuntimeError("talking to multiple EC2 instances is not supported for this task yet: %r" % stackname)
 
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -122,20 +122,25 @@ def connect_aws_with_stack(stackname, service):
     pname = project_name_from_stackname(stackname)
     return connect_aws_with_pname(pname, service)
 
-# TODO: rename to find_ec2_instances
-def find_ec2_instance(stackname):
+def find_ec2_instances(stackname, state='running', node_ids=None):
     "returns list of ec2 instances data for a *specific* stackname"
     # http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html
-    filter_by_cluster = {
+    conn = connect_aws_with_stack(stackname, 'ec2')
+    return conn.get_only_instances(filters=_all_nodes_filter(stackname, state=state, node_ids=node_ids))
+
+def _all_nodes_filter(stackname, state, node_ids):
+    query = {
         'tag-key': ['Cluster', 'Name'],
         'tag-value': [stackname],
-        'instance-state-name': ['running'],
     }
-    conn = connect_aws_with_stack(stackname, 'ec2')
-    return conn.get_only_instances(filters=filter_by_cluster)
+    if state:
+        query['instance-state-name'] = [state]
+    if node_ids:
+        query['instance-id'] = node_ids
+    return query
 
 def find_ec2_volume(stackname):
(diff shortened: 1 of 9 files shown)


src/buildercore/utils.py
```diff
@@ -224,11 +224,13 @@ def ymd(dt=None, fmt="%Y-%m-%d"):
         dt = datetime.now() # TODO: replace this with a utcnow()
     return dt.strftime(fmt)
 
-def die(assertion, msg):
+# TODO: this name is the opposite of assert, assertions read like "die if 1 == 1"
+# I'd try ensure(assertion, msg)
+def die(assertion, msg, *args):
```
giorgiosironi: thoughts on this name?

lsh-0:
just to verify this for myself (we're ok):
https://aws.amazon.com/premiumsupport/knowledge-center/ec2-instance-hour-billing/

src/buildercore/utils.py
lsh-0: `ensure` is good, use that.

src/lifecycle.py
```diff
@@ -0,0 +1,31 @@
...
+@requires_aws_stack
+@echo_output
+def last_start_time(stackname):
+    return lifecycle.last_start_time(stackname)
+
+@task
+@requires_aws_stack
+@timeit
+def stop_if_next_hour_is_imminent(stackname, minimum_minutes='55'):
+    # TODO: can we write a description of the @task somewhere?
```
lsh-0: Like this?
http://docs.fabfile.org/en/1.12/api/core/tasks.html#fabric.tasks.Task

giorgiosironi: I didn't know the docstring was exposes in the task list, then I'll remove the comment as the description is there.

lsh-0: oh, I understand your comment now. yes - the docstring for the function with the `task` annotation appears alongside the task name (albeit, truncated) when one does `./bldr -l`

each of the decorators should be using the `functools.wraps` decorator that ensures the wrapper func gets the docstring and other attributes of the wrapped function.

giorgiosironi: All of them do, I can confirm.

src/buildercore/lifecycle.py
```diff
@@ -0,0 +1,104 @@
...
+    
+    e.g. {'i-6f727961': 'stopped'}"""
+    return {node.id:node.state for node in _nodes(stackname, node_ids=node_ids)}
+
+def _nodes(stackname, state=None, node_ids=None):
+    "returns list of ec2 instances data for a *specific* stackname"
+    conn = _connection(stackname)
+    return conn.get_only_instances(filters=_all_nodes_filter(stackname, state=state, node_ids=node_ids))
+
+def _all_nodes_filter(stackname, state, node_ids):
```
lsh-0: could this logic be merged into `buildercore.core:find_ec2_instance` ?

giorgiosironi: More or less. I'll try doing it by changing a few names to make them all consistent.

src/buildercore/core.py
```diff
@@ -122,20 +122,25 @@ def connect_aws_with_stack(stackname, service):
...
     # http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html
-    filter_by_cluster = {
+    conn = connect_aws_with_stack(stackname, 'ec2')
+    return conn.get_only_instances(filters=_all_nodes_filter(stackname, state=state, node_ids=node_ids))
+
+def _all_nodes_filter(stackname, state, node_ids):
+    query = {
         'tag-key': ['Cluster', 'Name'],
         'tag-value': [stackname],
-        'instance-state-name': ['running'],
```
lsh-0: the default state of 'running' doesn't appear to have been preserved.

giorgiosironi: amended the last commit

lsh-0:
cool - please merge once Alfred is happy

PR #61, opened 2016-10-03
Cleaning up output of tasks

giorgiosironi:
- removing noisy comment
- consistently using LOG and stderr in `buildercore`
- human dates and times


giorgiosironi:
Solved conflicts, merging to avoid solving them again later.

PR #62, opened 2016-10-04
Waiting conditions for usable EC2 nodes

giorgiosironi:
Just because an EC2 instance is in the `running` state it doesn't mean it's ready to use. cloud-init is the designated service that logs when the daemons have started and you can start using the node again


giorgiosironi:
Feel free to request changes afterwards, but need to get this in to ensure build stability.

PR #64, opened 2016-10-04
Updating Route53 public DNS after restart

giorgiosironi:
We use the `context` of a stack to find out its public and private hostnames, and to update them with the current public and private ip addresses in the DNS entries for the public and private hosted zones. The EC2 nodes involved need to be running for this to be possible. The private ip shouldn't really change for an EC2 node inside a VPC, but it seems more robust and comes at no cost to update both public and private DNS entries.

Context has been saved to S3 only for a few, newer stacks. There is now a fallback that tries, in this order:
- the `.cfn/contexts/{stackname}.json` file
- the S3 key `contexts/{stackname}.json`
- the `/etc/build-vars.json` file on the EC2 node


src/buildercore/s3.py
```diff
@@ -77,5 +78,10 @@ def download(key, output_path):
     assert not os.path.exists(output_path), "given output path exists, will not overwrite: %r" % output_path
     k = builder_bucket().get_key(key)
     assert k, ("Cannot find %s in bucket %s" % (key, builder_bucket()))
+    _log_for_key(key, "downloading key %s")
     k.get_contents_to_file(open(output_path, 'w'))
     return output_path
+
+def _log_for_key(key, message):
```
lsh-0: I don't think we need this function. It doesn't improve readability and we don't gain much.

giorgiosironi: Ok, inlining it

src/buildercore/contexts.py
```diff
@@ -4,6 +4,8 @@
 from os.path import join, exists
 from . import config, s3
 from .decorators import if_enabled
+# temporary fallback
```
lsh-0: I don't understand this comment

giorgiosironi: clarified with link to function that needs these imports

lsh-0:
ok, I'm happy with the state of commit and we can merge it (assuming I haven't broken anything - Alfred?).

I do feel like there must be a better solution to this problem than storing state remotely.

giorgiosironi:
Looking at context, there are a few things that cannot be regenerated of various importance like `rds_password` and `date_rendered`. Given that a solution may be to always pull them from a node, but not all projects have an EC2 node and some have multiple ones. So unclear for now, but may be revisited in the future.

Alfred says yes so let's merge and get the `end2end` environment up again.

lsh-0:
> Given that a solution may be to always pull them from a node, but not all projects 
> have an EC2 node and some have multiple ones.

central store on master? which is, I guess, the storing of the context data in `.cfn/contexts` ...

PR #65, opened 2016-10-05
It appears that only the `instance` key is accessed here

giorgiosironi:
Apparently the code calling `stack_data` only reads the `instance` key of the dictionary, which comes from another call. So we can spare one call, which is used in many places and hence can speed up the connection


giorgiosironi:
Sneakily getting this in to build upon it other simplifications.

PR #67, opened 2016-10-06
more linting!

lsh-0:
added autopep8 which modifies files in-place, removing whitespace and improving indenting etc.

@giorgiosironi - this may affect testing if the developer isn't running the .lint.sh or test.sh scripts prior to committing - the test run will modify files. 

if this is a problem I'll add a flag or something to turn it off


lsh-0:
after a conversation with Giorgio, I don't see this affecting his work. Merging.

PR #68, opened 2016-10-07
Upgrade Python to 2.7.12

giorgiosironi:
Python is such a hard dependency of Ubuntu and Salt that it's safer to
update it as part of the base packages instead of using a Salt state.

This is also a portable way of doing it no matter what is the basebox or
AMI being used.

Tangentially, also the virtualenvs have to be recreated after upgrading,
because the `pip` installed there complains about not being able to find
`urllib3` as part of its `requests` Python module dependency. This is
somewhat due to distro maintainers un-vendoring dependencies and taking
them out of the packages themselves; also, so much for virtual env being
an isolated Python.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -36,10 +36,30 @@ else
     fi
 fi
 
+upgrade_python=false
+install_git=false
+# Python is such a hard dependency of Salt that we have to upgrade it outside of it to avoid changing it while it is running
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Just for sanity check, before merging I'm testing this with my local builder on some `--continuumtest` instances.

giorgiosironi:
`lax--continuumtest` now has an upgrade Python, and the bootstrap is idempotent. Merging as agreed on Friday.

PR #69, opened 2016-10-07
SSH tasks gracefully ask to start an EC2 node if it's stopped

giorgiosironi:


giorgiosironi:
Needing this in everyday work, will also open up the opportunity for autostarting in other places like `update`.

PR #70, opened 2016-10-11
Updating the template, but only with new resources

giorgiosironi:
`Outputs` and `Resources` are compared with a newly generated template by this task. Only the newly appearing entities will be used as a change set against the current template.

The source of the current template is Cloudformation itself, but it is cached locally for the duration of the current task.

### DIFF
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -169,8 +169,8 @@ def write_template(stackname, contents):
 
 def read_template(stackname):
     output_fname = os.path.join(STACK_DIR, stackname + ".json")
-    with open(output_fname, 'r') as f:
-        return json.loads(f.read())
+    with open(output_fname, 'r') as template_file:
+        return json.loads(template_file.read())
 
 def validate_aws_template(pname, rendered_template):
     conn = core.connect_aws_with_pname(pname, 'cfn')
@@ -250,8 +250,8 @@ def template_delta(pname, **more_context):
     context = build_context(pname, **more_context)
     template = json.loads(render_template(context))
     return {
-        'Outputs': {title:o for (title, o) in template['Outputs'].iteritems() if title not in old_template['Outputs']},
-        'Resources': {title:r for (title, r) in template['Resources'].iteritems() if title not in old_template['Resources']}
+        'Outputs': {title: o for (title, o) in template['Outputs'].iteritems() if title not in old_template['Outputs']},
+        'Resources': {title: r for (title, r) in template['Resources'].iteritems() if title not in old_template['Resources']}
     }
 
 def merge_delta(stackname, delta):
@@ -259,8 +259,6 @@ def merge_delta(stackname, delta):
     template = read_template(stackname)
     for component in delta:
         assert component in ["Resources", "Outputs"]
-        for title in delta[component]:
-            template[component].update(delta[component])
+        template[component].update(delta[component])
     write_template(stackname, json.dumps(template))
     return template
-
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -137,6 +137,19 @@ def build_context(pname, **more_context): # pylint: disable=too-many-locals
     return context
 
 
+def choose_config(stackname):
+    (pname, instance_id) = core.parse_stackname(stackname)
+    pdata = project.project_data(pname)
+    more_context = {
+        'stackname': stackname,
+    }
+    if instance_id in project.project_alt_config_names(pdata):
+        LOG.info("using alternate AWS configuration %r", instance_id)
+        # TODO there must be a single place where alt-config is switched in
+        # hopefully as deep in the stack as possible to hide it away
+        more_context['alt-config'] = instance_id
+
+    return more_context
 #
 #
 #
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -19,6 +19,7 @@ A developer wants a temporary instance deployed for testing or debugging.
 import os, json, copy
 from slugify import slugify
 from . import utils, trop, core, project, context_handler
+from .utils import ensure
 from .config import STACK_DIR
 
 import logging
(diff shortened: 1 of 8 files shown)


giorgiosironi:
First failed because `lax--end2end` was not running, afterwards succeeded, adding the 2 queues:
https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stack/detail?stackId=arn:aws:cloudformation:us-east-1:512686554592:stack%2Flax--end2end%2F903d1a90-8a4c-11e6-8a0a-50d5ca63261e

lsh-0:
sorry, completely forgot about this PR, on it now

src/buildercore/cfngen.py
```diff
@@ -154,6 +167,11 @@ def write_template(stackname, contents):
     open(output_fname, 'w').write(contents)
     return output_fname
 
+def read_template(stackname):
+    output_fname = os.path.join(STACK_DIR, stackname + ".json")
+    with open(output_fname, 'r') as template_file:
```
lsh-0: there is a more straightforward: `return json.load(open(output_fname), 'r'))`

the `s` in `loads` and `dumps` is for `string`

src/buildercore/cfngen.py
```diff
@@ -223,3 +241,24 @@ def generate_stack(pname, **more_context):
...
+    return {
+        'Outputs': {title: o for (title, o) in template['Outputs'].iteritems() if title not in old_template['Outputs']},
+        'Resources': {title: r for (title, r) in template['Resources'].iteritems() if title not in old_template['Resources']}
+    }
+
+def merge_delta(stackname, delta):
+    """Merges the new resources in delta in the local copy of the Cloudformation  template"""
+    template = read_template(stackname)
+    for component in delta:
+        assert component in ["Resources", "Outputs"]
```
lsh-0: um - do you even need this assertion here?

src/tests/test_buildercore_cfngen.py
```diff
@@ -35,3 +35,9 @@ def test_validation_elife_projects(self):
             sleep(0.25)
 
         self.switch_in_test_settings()
+
+    def test_empty_template_delta(self):
```
lsh-0: can we get a more readable test, please? for example:

```
test_foo():
  render CFN template
  ensure some values in CFN template 
  update CFN template
  ensure old values haven't changed (immutable)
  ensure new values present
```

giorgiosironi: Would like to, I'll have to manually modify the template, because rendering will always produce the same result.

src/cfn.py
```diff
@@ -41,11 +44,31 @@ def ensure_destroyed(stackname):
 @task(alias='aws_update_stack')
 @requires_aws_stack
 @timeit
-def update(stackname):
+def update(stackname, part_filter=None):
```
lsh-0: `part_filter` means nothing to me. I can't even deduce it's purpose looking at the `buildercore` code. This parameter needs a more obvious name and the docstring should be updated

lsh-0:
I'm 50/50 on this PR - I'm not confident AWS isn't going to do something stupid like restart ec2 when sqs queues are added - their update logic is such a black box.

Also, are we modifying the originally saved .cfn template?

Also, a question for another PR perhaps, is there a way for another user of builder to download the saved .cfn templates? if I needed to update a template and only you had the template on your machine, I wouldn't be able to. This is solved with the ssh keypairs by storing them in s3 and providing a download/upload task. It was solved in the old elife-builder with a `@sync` decorator that would do a destructive copy down from s3 then an non-destructive upload to s3 afterwards. this had it's drawbacks but worked surprisingly well otherwise.

src/cfn.py
giorgiosironi: Proposals:
- `filter` is too generic and shadows the `filter` built-in function
- `resource_filter` is more precise but you have to understand it's referring to the template's resources. However you may have two EC2 nodes but it's the filter is still just `ec2`
- `service_filter` is how AWS would call it, `ec2` and `s3` are `Services` in their console, still `service` has many meanings in this world
- `aws_service_filter` starts to become too long
Choice?

lsh-0: can you explain what it does? what value would you pass to it?

giorgiosironi:
> I'm 50/50 on this PR - I'm not confident AWS isn't going to do something stupid like restart ec2 when sqs queues are added - their update logic is such a black box.
 
It is, we have not much control, although the alternative of deleting the stack and recreating it ensures that the EC2 will be restarted (actually recreated).

> Also, are we modifying the originally saved .cfn template?

In particular, yes (see later)

> Also, a question for another PR perhaps, is there a way for another user of builder to download the saved .cfn templates? if I needed to update a template and only you had the template on your machine, I wouldn't be able to. This is solved with the ssh keypairs by storing them in s3 and providing a download/upload task. It was solved in the old elife-builder with a @sync decorator that would do a destructive copy down from s3 then an non-destructive upload to s3 afterwards. this had it's drawbacks but worked surprisingly well otherwise.

This PR introduces Cloudformation itself as the source of the template, so the path is:
- Cloudformation as the authoritative source
- local copy in `.cfn/stacks` (updated through `cfngen` if a change is needed)
- in-memory copy everytime `cfngen.read_template` is called
so there was no need to store them also on S3, where we have instead to put the `.cfn/contexts` for example.

src/tests/test_buildercore_cfngen.py
giorgiosironi: Tried doing this, the problematic step is `update CFN template` as I can update the old one but not the new one (that is always generated). Therefore the test sounds like `remove stuff from CFN template, check that the delta adds them` which is also confusing. Thinking again...

lsh-0:
... that's right, because we can download the template we originally sent to CFN. Fair enough. I didn't glean that looking at your changes, sorry.

lsh-0:
I think I'm going to spend a bit of time getting re-acquainted with buildercore soon. Lots of new things in there I wouldn't mind polishing up.

giorgiosironi:
Check out the existing PRs before diving into `buildercore` to avoid conflicts, which are already happening

lsh-0:
we need to look at those other PRs at some point too, yes.

PR #73, opened 2016-10-18
Build vars revamp

giorgiosironi:
- Manipulating buildvars with the bootstrap users
- A small task to force/repair/change a buildvars value when needed
- `buildvars.fix` instead of writing just a couple of fields, fully regenerates them from
If you think about it, using the bootstrap users makes sense as EC2 instances have build vars right when they are created, without waiting for Salt to run. So we should be able to manipulate them at that level, without relying on the Salt update to deliver the `elife` user and access to it.
The keys of users configured in `pillar.elife` are added to both deploy and
bootstrap users, so this works for anyone who was running buildvars
commands (like Alfred) before.

### DIFF
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -207,12 +207,18 @@ def stack_all_ec2_nodes(stackname, workfn, username=config.DEPLOY_USER, **kwargs
     if isinstance(workfn, tuple):
         workfn, work_kwargs = workfn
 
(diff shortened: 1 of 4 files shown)


giorgiosironi:
Have been using this for a week with no ill-effect, and need it to repair some contexts which are missing recently introduced or changed keys like `instance_id`, merging.

PR #74, opened 2016-10-19
Trying to setup a nicer message for interrupted SSH connection

giorgiosironi:


giorgiosironi:
This one's easy, let's see it in action.

PR #75, opened 2016-10-25
Create EC2 instances for the same stack in separate AZs

giorgiosironi:


giorgiosironi:
So this works to create the two instances is `us-east-1d` (with all the others) and `us-east-1e`. However, the Salt master is in the subnet of `us-east-1d`, so it's not reachable from the node in the other subnet at the moment. Checking out how we would connect them.

giorgiosironi:
The last commit addresses this security groups problem, making the master reachable by all the EC2 nodes inside eLife's VPC.

giorgiosironi:
Have been using this for recreating journal--end2end, as seen on https://console.aws.amazon.com/ec2/v2/home?region=us-east-1#Instances:search=journal-end2end;sort=instanceId and it seems to work well.

PR #76, opened 2016-10-28
Better error message on not valid build vars

giorgiosironi:


src/buildvars.py
```diff
@@ -54,10 +54,11 @@ def _validate():
             'build vars not found (%s). use `./bldr buildvars.fix` to attempt to fix this.',
             buildvars
         )
-        core_utils.ensure(
-            core_utils.hasallkeys(buildvars, ['stackname', 'instance_id', 'branch', 'revision', 'is_prod_instance']),
-            'build vars are not valid. use `./bldr buildvars.fix` to attempt to fix this.'
-        )
+        for key in ['stackname', 'instance_id', 'branch', 'revision', 'is_prod_instance']:
```
lsh-0: if you need the missing keys for the error message, try `utils.missingkeys`

PR #77, opened 2016-10-28
Noticing Salt failures

giorgiosironi:
Just leaving this here, this log is automatically sent to Loggly so we could set up an alert on it. Needs some more testing


giorgiosironi:
After merging this, Loggly will receive a log of all Salt failures triggered by a builder command - which means all builds. If this works well we can also think about sending the same error message in the `daily-system-updates`.

PR #80, opened 2016-11-03
Patching up AMI task to make it work

giorgiosironi:


giorgiosironi:
By the way, the `basebox` only builds the `elife.java` state on top of the Ubuntu image. Maybe we should add some more from the `elife` state? Maybe excluding things like the SSH public keys...

giorgiosironi:
@lsh-0 check out this which will make faster to try out new C4 instances for bot-lax-adaptor.

PR #82, opened 2016-11-09
Moved to 4GB memory for Search

stephenwf:
@giorgiosironi @lsh-0

This should fix the memory issue on CI.


giorgiosironi:
Merging, then I will destroy and recreate the stacks to get the new instance type.

PR #83, opened 2016-11-10
Feat formula dependencies

lsh-0:


scripts/bootstrap.sh
```diff
@@ -119,15 +119,18 @@ if [ -d /vagrant ]; then
...
-    fi
-    
-    # overwrite the general purpose /etc/salt/minion file created above with 
-    # this more complex one for dev environments only
-    cp /vagrant/scripts/salt/minion /etc/salt/minion
+    # vagrant makes all formula dependencies available, including builder base formula
+
+    # overwrite the general purpose /etc/salt/minion file created above 
+    # with this more complex one for dev environments only
+    cp /vagrant/scripts/salt/minion.template /etc/salt/minion
```
giorgiosironi: this `cp` could go in the `else` that follows for simmetry?

Vagrantfile
```diff
@@ -244,6 +244,18 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
                     end
                     formula_path
                 }
+                # write the minion file
+                minion_cfg = YAML.load_file("scripts/salt/minion.template")
+                more_base_paths = formula_paths.map{|p| "/vagrant/" + p + "/"}
+                minion_cfg['file_roots']['base'].insert(1, *more_base_paths)
```
giorgiosironi: so this keeps inserting the depended-upon formulas in the place in the list where `builder-base-formula` was before, which is good

giorgiosironi:
:+1: @stephenwf should try this out when the Java 8 repositories are up again.

Vagrantfile
lsh-0: yes, it preserves the ordering in the template with the custom salt directory first for any overrides and the project-local /srv/salt at the end. the funny `.insert(1, *more_base_paths)` is something I've not seen done in Python. It does `[1 2 3].insert(1, *[a b c]) => [1 a b c 2 3]`

lsh-0:
yes, I'd like for him to give it a shot before I merge it in

Vagrantfile
lsh-0: huh! in Python you can assign to a slice operation: `x = [1,2,3]; x[1:1] = ['a', 'b']; => [1,'a','b',2,3]`

scripts/bootstrap.sh
giorgiosironi: yeah I don't care very much about this, but in general I tend to limit the states of a system to as few as possible e.g. `file missing` -> `correct file present` rather than `file missing` -> `incorrect file present` -> `correct file present` or if it gets rerun during an update `correct file present` -> `incorrect file present` -> `correct file present`
(this not for performance reasons but for simplicity, e.g. a crash can leave the system in any of the intermediate state, from that it needs to recover and there are more possible inputs to consider, etc.)

lsh-0: feel free to make the modification

giorgiosironi: looking back at this now, here we are in a long conditional that is only executed in the Vagrant case. Do we need the else at all? The original `/vagrant/scripts/salt/minion.template` shouldn't work at all as it's missing the `builder-base-formula`; therefore I propose to put an `exit 1` inside the else here rather than using it and finding out some states are missing.
This would become something like:
```
    custom_minion_file="/vagrant/scripts/salt/$minion_id.minion"
    if [ ! -e "$custom_minion_file" ]; then
        echo "couldn't find $custom_minion_file"
        exit 1
    fi
    cp "$custom_minion_file" /etc/salt/minion
```

stephenwf:
I'm testing now, I've added `dummy-api` to the top file and its building, but I have a feeling that port 80 is going to be contested. Since the API dummy will be used in this way as a dependency for other project, can it be changed to use a non-standard web port?

ping @thewilkybarkid

stephenwf:
Also @lsh-0 I'm getting:

`No matching sls found for 'api-dummy.api-dummy' in env 'base`

or 

`No matching sls found for 'api-dummy' in env 'base`

 When changing the top file locally. How do we reference these new dependencies?

lsh-0:
in the formula `salt/top.sls` file you should be referencing api-dummy states similar to:

```
elife-search--*:
    - ...
    - api-dummy
    - api-dummy.somestate
   - ...
```

Are you changes committed ? I'll bring up a local instance and see what can I see

stephenwf:
If you swap out the search `formula-repo` to my fork: https://github.com/stephenwf/search-formula you can use the master branch there.

lsh-0:
sorry for the delay, firefighting another problem. I'm bringing up an instance now to see if I can't find your problem. Bear with me.

stephenwf:
No problem, thanks @lsh-0

lsh-0:
ok, I see what the problem is. fix in progress

lsh-0:
so, two changes have been made. first one is to the Vagrant file to fix the generated paths. The second is to `builder-base-formula` to add a 'salt' directory, as it's structured slightly differently to regular formulas.

if you could pull down the changes to builder, stop and start your vagrant machine then do a 'vagrant provision', it should pull down the changes to builder-base-formula

stephenwf:
Great, building now, will let you know how it goes. I still think port 80 may cause a conflict

giorgiosironi:
Port 80 should be used by both sites served by the same nginx, *but* we probably need to add to the formulas the virtual host with `server_name`. Right now they would both try to respond to a request to localhost:80, while each (or at least the api-dummy) will have to be limited to responding to a set of hosts like `api-dummy.local`.

lsh-0:
could you have the api-dummy running on port 8080 while in dev?

giorgiosironi:
That would be incompatible with how the api-dummy runs when it's in an `api-dummy--*` stack, which uses the same formula.

stephenwf:
Can we split out the sls files for api-dummy so we can define our own NGinx config:
https://github.com/elifesciences/api-dummy-formula/blob/master/salt/api-dummy/init.sls#L71-L80

stephenwf:
@giorgiosironi @lsh-0 

I just tried a test where I split the nginx configuration into its own sls file in this repo:
https://github.com/stephenwf/api-dummy-formula

I then used this branch, edited the repos for the salt files to point to my 2 forks of the formula and tried to install. (Adding in my own nginx into the search formula)

I destroyed the vagrant box and up + provisioned.. now there is no search at all, not in the /srv folder or the /etc/nginx.

Not sure whats happening with it.

lsh-0:
so we have a `dev` or `env=='dev'` value that is used to determine if we're within Vagrant: https://github.com/elifesciences/builder-base-formula/blob/master/pillar/elife.sls#L11

what I was suggesting is that api-dummy, when operating within Vagrant, uses a different port to avoid conflicts.

lsh-0:
> Not sure whats happening with it.

yeah, I'm going to need more information if you want help debugging this.

stephenwf:
@lsh-0 Here is the console for the commands:

vagrant up
vagrant provision

and then the ssh into the box itself to find the empty nginx configuration and folders. All built with my two salt stack forks

giorgiosironi:
Your two salt forks would mean you have some different values in projects/elife.yaml? Can you post that or its diff with respect to master? First step would be to reproduce it here.

giorgiosironi:
Or which remotes and commit (sha) you have for `builder-base-formula` and `search-formula`

stephenwf:
Here is the patch (most of it is quite space)

```diff
diff --git a/projects/elife.yaml b/projects/elife.yaml
index 492b13e..8297543 100644
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -76,23 +76,23 @@ defaults:
                 ami: ami-9eaa1cf6 # Ubuntu 14.04
     vagrant:
         box: ubuntu/trusty64 # Ubuntu 14.04
-        # box-url not needed for boxes hosted on Atlas
-        box-url: null
+        # box-url not needed for boxes hosted on Atlas
+        box-url: null
         ip: 192.168.33.44
         ram: 1024
         cpus: 2
         cpucap: 100 # percent (vagrant default)
-
+
 basebox:
     aws:
         ec2:
             ami: ami-9eaa1cf6 # Ubuntu 14.04 (correct, but older)
         ports:
-            - 22
+            - 22
     vagrant: {}

 master-server:
-    # formula-repo for the 'master-server' project should contain the
+    # formula-repo for the 'master-server' project should contain the
     # confidential pillar data, master config and state top file.
     # see: https://github.com/elifesciences/builder-private-example
     formula-repo: ssh://git@github.com/elifesciences/master-server-formula
@@ -256,7 +256,7 @@ journal-cms:
         ports:
             - 22
             - 443
-            - 80:
+            - 80:
                 cidr-ip: 10.0.2.0/24 # access via VPC ip range only
         sqs:
             # journal-cms--prod, journal-cms--end2end, etc.
@@ -410,7 +410,7 @@ anonymous:
 jats4r:
     aws:
         ec2:
-            ami: ami-c60b90d1 # us-east-1 xenial 16.04 LTS amd64 hvm:ebs-ssd 20160815 hvm
+            ami: ami-c60b90d1 # us-east-1 xenial 16.04 LTS amd64 hvm:ebs-ssd 20160815 hvm
         type: t2.micro
         ports:
             - 22
@@ -436,7 +436,7 @@ medium:
         clustered:
             ec2:
                 cluster-size: 2
-            elb:
+            elb:
                 protocol: http

 bus:
@@ -462,14 +462,17 @@ search:
     domain: False
     subdomain: search
     repo: https://github.com/elifesciences/search
-    formula-repo: https://github.com/elifesciences/search-formula
+    #formula-repo: https://github.com/elifesciences/search-formula
+    formula-repo: https://github.com/stephenwf/search-formula
     formula-dependencies:
         - https://github.com/elifesciences/builder-base-formula
         - https://github.com/elifesciences/api-dummy-formula
     vagrant:
-        ram: 1024
+        ram: 4096
         ports:
             1244: 80
+            1245: 8080
+            9920: 9200
     aws:
         type: t2.medium # 4 GB of RAM: once we know more about which software
                         # will be installed, we can choose the right machine
@@ -500,7 +503,7 @@ jats4r-validator:
     vagrant:
         ports:
             1239: 80
-
+
 statusbase:
     subdomain: statusbase
     repo: https://github.com/ep320/statusbase
```

stephenwf:
I manually changed the repo origin in the cloned projects for the api dummy
but it is using the master branch from here:

https://github.com/stephenwf/api-dummy-formula

(here is the diff: https://github.com/stephenwf/api-dummy-formula/commit/676cc4c5483158719ee78d02269b275b55c82c5d)

lsh-0:
> +            9920: 9200

just a note with these sorts of ports: if they're unlikely to clash you can just pass them through like: `9200:9200`

but with ports like `80` where you may have multiple virtual machines using them, best to use some sort of port forwarding.

stephenwf:
Sorry, I have a local elasticsearch that is bound to 9200!

lsh-0:
ha! yeah, another good reason to use port mapping.

lsh-0:
ok, I'm using your two forks and I can successfully provision the vagrant machine.

lsh-0:
if you delete the repo in `./cloned-repos/` you have to stop/start Vagrant because it loses it's pointer to the mounted directory. not sure what your situation is. 

I'll destroy and rebuild it now to see how that goes

stephenwf:
Thats strange!

does the /srv/ folder have both search and dummy?

I'm half way through another destroy/rebuild

lsh-0:
> does the /srv/ folder have both search and dummy?

`/srv` would only have the project you're building - in this case just the search project. `builder` is finding the other dependant formulas using the `/vagrant/cloned-repos/` paths. You can inspect it's paths with `$ cat /etc/salt/minion`

stephenwf:
Heres a dump from the build I just ran:

```
elife@search--vagrant:~$ cat /etc/salt/minion
---
file_client: local
log_level: info
fileserver_backend:
- roots
file_roots:
  base:
  - "/srv/custom/salt/"
  - "/vagrant/cloned-projects/builder-base-formula/salt/"
  - "/vagrant/cloned-projects/api-dummy-formula/salt/"
  - "/vagrant/cloned-projects/search-formula/salt/"
  - "/srv/salt/"
pillar_roots:
  base:
  - "/srv/custom/pillar/"
  - "/vagrant/cloned-projects/builder-base-formula/salt/pillar/"
  - "/vagrant/cloned-projects/api-dummy-formula/salt/pillar/"
  - "/vagrant/cloned-projects/search-formula/salt/pillar/"
  - "/srv/pillar"
elife@search--vagrant:~$ cd /srv
elife@search--vagrant:/srv$ ll
total 12
drwxr-xr-x  3 root  root  4096 Nov 14 14:42 ./
drwxr-xr-x 24 root  root  4096 Nov 14 14:36 ../
drwxr-xr-x 12 elife elife 4096 Nov 14 14:42 api-dummy/
lrwxrwxrwx  1 root  root    23 Nov 14 14:42 custom -> /vagrant/custom-vagrant/
lrwxrwxrwx  1 root  root    20 Nov 14 14:42 pillar -> /project/salt/pillar/
lrwxrwxrwx  1 root  root    13 Nov 14 14:42 salt -> /project/salt/
elife@search--vagrant:/srv$
```

stephenwf:
Salt file looks like this:

```bash
elife@search--vagrant:/srv/salt$ cat example.top
base:
    '*':
        - elife
        - elife.php7
        - elife.nginx
        - elife.nginx-php7
        - elife.puli
        - elife.gearman
        - search.java
        - search
        - search.processes
        - api-dummy
        - search.elasticsearch
        - elife.proofreader-php
        - search.api-dummy
```

lsh-0:
just built without a problem. my config looks like yours. going to modify the port mappings like you have now.

stephenwf:
Okay @lsh-0 I'm going to clone a new builder repo and try again, thanks for your help

lsh-0:
ok - no problems there. is there a command or something you want me to run?

lsh-0:
```
$ curl localhost
{
    "title": "No route found for \"GET /\"",
    "type": "about:blank"
}
```

lsh-0:
I hear you're in the office tomorrow. Perhaps just move your current builder to one side so we can debug it's state then and figure out just what was happening

stephenwf:
`curl -I localhost:8080/blog-articles`

and

`curl -I localhost/search`

These would be useful just to get an idea if the two projects are indeed running side-by-side (broken or not!)

Yes I think thats a good idea, I will move on to something else for now

lsh-0:
```
elife@search--vagrant:~$ curl -I localhost:8080/blog-articles
curl: (7) Failed to connect to localhost port 8080: Connection refused

elife@search--vagrant:~$ curl -I localhost/search
HTTP/1.1 200 OK
Server: nginx/1.4.6 (Ubuntu)
Content-Type: application/vnd.elife.search+json; version=1
Connection: keep-alive
Cache-Control: no-cache
Date: Mon, 14 Nov 2016 15:09:53 GMT
```

stephenwf:
That looks like the same issue I am having, can you `vagrant ssh` and run `ls /srv` and see if the search was built

lsh-0:
ok, I can see the /project mount is broken

screenshot at 2016-11-14 15-36-39

lsh-0:
ok, I can replicate this

lsh-0:
so the problem was that the /project directory (an alias for `./cloned-projects/some-project-formula`) wasn't being mounted on a `vagrant reload` but was on `up` and `provision`.

still getting this however: 
```
elife@search--vagrant:~$ curl -I localhost:8080/blog-articles
curl: (7) Failed to connect to localhost port 8080: Connection refused
```

lsh-0:
oh - you're expecting something like `/srv/search/`, right? let me take a looksee

lsh-0:
so I think what is happening is that it's finding the api-dummy `top.sls` file first and using that. I'm going to try re-ordering the contents of `/etc/salt/minion` to ensure the project top file is hit first

lsh-0:
so this is what I get now with the formula dependencies re-ordered:
screenshot at 2016-11-14 16-06-13

It looks like you're duplicate states or state IDs between formulas rather than re-using them. 

I'll work on getting the dependencies right.

lsh-0:
Ok, I've made the ordering of the formula dependencies more strict but you'll need to fix up those state id's between projects to get this to work

stephenwf:
Can confirm this is now working as expected. Had to run this following using Vagrant SSH

```
cd /src/search/
sudo chmod -R 777 cache
php bin/console search:setup --env=dev
php bin/console gearman:worker --env=dev &
php bin/console gearman:import all --env=dev
exit
```

stephenwf:
@lsh-0 I think this is ready to go, thanks for you help getting the build working.

lsh-0:
regarding those commands you ran - are they captured in the formula at all? I can help if not.

stephenwf:
the command `sudo chmod -R 777 /src/search/cache` is causing an issue, according to @giorgiosironi both the `elife` user and the `www` user needs write access, not exactly sure how to do that in the formula.

Apart from that, I think I have the other commands covered, going to make another long running process for watching SQS too

lsh-0:
can you list the operations both users need to do? I've found it easier - with php applications anyway - to do everything as www and have other processes that need to call it from the cli su to the right user

stephenwf:
Currently working on the commands here: https://github.com/elifesciences/search-formula/pull/3/files

Including this command too: https://github.com/stephenwf/search-formula/blob/575986b5b4de6452a78429eaa1e85beb959247e6/salt/search/init.sls#L52-L58

and this: https://github.com/stephenwf/search-formula/blob/575986b5b4de6452a78429eaa1e85beb959247e6/salt/search/config/etc-init-search-gearman-worker.conf

The last command that I'm adding just now will be similar to the conf file above, all commands will be ran through `./bin/console`

giorgiosironi:
The `elife` user is part of the `www-data` group so we usually set up things with 775 (dir) and 664 (file) permissions. However, there is also a sticky bit involved that makes it so that the created files will have the `www-data` group even if they are created by `elife` (it's not his primary group):
https://github.com/elifesciences/journal-formula/blob/master/salt/journal/init.sls#L54
and a `umask` because the default seems to be `022` instead of `002` and that would make the new files not-writable and non-deletable by `www-data` group:
https://github.com/elifesciences/journal-formula/blob/master/salt/journal/init.sls#L77
The umask we usually set in the bootstrap of the application with http://php.net/umask but also in the formula like that for commands we have no control over the source code, like `composer install` or any PHP command that touches that folder.

stephenwf:
so if we set umask(022) inside of the console command and bootstrap (kernel) that will solve the issue?

stephenwf:
@giorgiosironi @lsh-0 

Is this ready to be merged now?

giorgiosironi:
Umask plus the updates in the formula (I have linked journal as a reference). This one is not related to the search project in particular, but to project dependencies in general so it's good to go (I have some concerns related to the order of salt folders but I will propose updates later.)

lsh-0:
> but I will propose updates later

I noted your proposal, but lets get this in and working first.

lsh-0:
just dropped the project's ram to 2048. after that I'll merge

lsh-0:
just dropped the project's ram to 2048. after that I'll merge

lsh-0:
just dropped the project's ram to 2048. after that I'll merge

PR #84, opened 2016-11-14
Destroying leftover DNS entries

giorgiosironi:
We modify these DNS entries when we start and stop an EC2 instance. CloudFormation then abandons them and does not delete them on destroy. Therefore, we destroy them manually

### DIFF
--- a/src/cfn.py
+++ b/src/cfn.py
@@ -28,12 +28,11 @@ def destroy(stackname):
         print 'got:'
         print '\n'.join(difflib.ndiff([stackname], [uin]))
         exit(1)
-    return bootstrap.delete_stack(stackname) and core_lifecycle.delete_dns(stackname)
-
+    return bootstrap.delete_stack(stackname)
 @task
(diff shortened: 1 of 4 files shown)


lsh-0:
is it this you're seeing? https://forums.aws.amazon.com/thread.jspa?threadID=148764

src/cfn.py
```diff
@@ -28,12 +28,12 @@ def destroy(stackname):
         print 'got:'
         print '\n'.join(difflib.ndiff([stackname], [uin]))
         exit(1)
-    return bootstrap.delete_stack(stackname)
+    return bootstrap.delete_stack(stackname) and core_lifecycle.delete_dns(stackname)
```
giorgiosironi: should be possible, yes it's that problem

giorgiosironi:
Have been using this for several days, works well.

lsh-0:
ah, good to know. let me take a peek and see if we can merge

PR #87, opened 2016-11-24
Added support for creating S3 buckets

giorgiosironi:


giorgiosironi:
@lsh-0 I have used this from my own machine to add some S3 buckets for the elife-bot without recreating the stack instances. Old versions of the builder should just ignore the additional S3 metadata in contexts until merged.

lsh-0:
can we add support for the deletion policy like we do with RDS? I want it made explicit that the contents of the S3 bucket will be destroyed when the stack is torn down

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html

giorgiosironi:
Added the default retention in the template, which for this bucket is `Delete` as it's a temporary folder for silent corrections. However I also made it configurable so that other buckets can use the `Retain` policy in the future.

giorgiosironi:
Used for the silent corrections bucket, needed for the new published CDN-related bucket, merging.

PR #93, opened 2016-12-09
Precisely specifying the Salt version

giorgiosironi:
Newly created minions are at 2016.3.4, while older machines are at 2016.3.1 which has a bug that doesn't show `Total run time: ... s` when running highstate.

Since the grep in `scripts/bootstrap.sh` is on this value, the version did not get upgraded. With this change, everyone will get `2016.3.4`.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1,6 +1,6 @@
 defaults:
     description: defaults for all projects in this file
-    salt: '2016.3' # the version of salt these project use
+    salt: '2016.3.4' # the version of salt these project use
     # use False with a subdomain to assign internal addresses only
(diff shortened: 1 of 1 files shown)


giorgiosironi:
@lsh-0

lsh-0:
does it work? that value is used in the salt bootstrap script and it can be ... picky.

giorgiosironi:
It ran with `./bldr update`, but I'll try also a Vagrant VM from scratch.

giorgiosironi:
Also working there:
```
[16:25:08][elife@medium--vagrant:~]$ salt-call --version
salt-call 2016.3.4 (Boron)
```

PR #97, opened 2017-01-09
Added missing bus topics

giorgiosironi:
Taken from https://github.com/elifesciences/api-raml/blob/develop/src/api.raml endpoints


giorgiosironi:
@thewilkybarkid check this is correct and whether we need to add the two commented ones, and with which plural name

PR #98, opened 2017-01-13
Added recommendations entry

stephenwf:
@giorgiosironi Also sorry about the whitespace fixes!

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -540,7 +540,7 @@ recommendations:
         ports:
             - 22
             - 80:
-                cidr-ip: 10.0.2.0/24 # internal access only
+                cidr-ip: 10.0.2.0/16 # internal access only
         sqs:
             recommendations--{instance}:
                 subscriptions:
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -522,6 +522,33 @@ search:
...
+        ram: 2048
+        ports:
+            1244: 80
+            1245: 8080
+    aws:
+        type: t2.medium # 4 GB of RAM: same as search for now.
+        ports:
+            - 22
+            - 80:
+                cidr-ip: 10.0.2.0/24 # internal access only
```
giorgiosironi: I think now this can be `10.0.0.0/16` to be the whole of the VPC. @lsh-0 ?

lsh-0: yeah, I think we have a few cases where I've put the very tight `/24` range in there

projects/elife.yaml
```diff
@@ -540,7 +540,7 @@ recommendations:
         ports:
             - 22
             - 80:
-                cidr-ip: 10.0.2.0/24 # internal access only
+                cidr-ip: 10.0.2.0/16 # internal access only
```
giorgiosironi: it's 10.0.0.0/16 (which means from 10.0.0.0 to 10.0.255.255)

PR #99, opened 2017-01-26
Added personalised covers to elife projects

stephenwf:
Inline with this: https://github.com/elifesciences/personalised-covers-formula/pull/1

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -557,18 +557,18 @@ personalised-covers:
         - https://github.com/elifesciences/builder-base-formula
         - https://github.com/elifesciences/api-dummy-formula
     vagrant:
-        ram: 2048
+        ram: 1024
         ports:
             1246: 80
             1247: 8080
     aws:
-        type: t2.medium # 4 GB of RAM: Unsure memory stats for PDF generation.
+        type: t2.micro # 1 GB of RAM: Unsure memory stats for PDF generation.
         ports:
             - 22
             - 80:
                 cidr-ip: 10.0.0.0/16 # internal access only
         s3:
-            "s3-personalised-covers--{instance}":
+            "{instance}-elife-personalised-covers":
                 deletion-policy: retain
 
 jats4r-validator:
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -548,6 +548,29 @@ recommendations:
                     - bus-subjects--{instance}
                     - bus-collections--{instance}
 
+personalised-covers:
+    domain: False
```
giorgiosironi: `domain: False` only if no link has to be sent outside (e.g. in emails), or if those link are directly to S3 (bad), is that correct?

projects/elife.yaml
```diff
@@ -548,6 +548,29 @@ recommendations:
...
+    subdomain: personalised-covers
+    repo: https://github.com/elifesciences/personalised-covers
+    formula-repo: https://github.com/elifesciences/personalised-covers-formula
+    formula-dependencies:
+        - https://github.com/elifesciences/builder-base-formula
+        - https://github.com/elifesciences/api-dummy-formula
+    vagrant:
+        ram: 2048
```
giorgiosironi: it can probably work with 1024? there is no ElasticSearch here

projects/elife.yaml
```diff
@@ -548,6 +548,29 @@ recommendations:
...
+        ports:
+            1246: 80
+            1247: 8080
+    aws:
+        type: t2.medium # 4 GB of RAM: Unsure memory stats for PDF generation.
+        ports:
+            - 22
+            - 80:
+                cidr-ip: 10.0.0.0/16 # internal access only
```
giorgiosironi: ok, internal access only so `domain: False` is consistent

projects/elife.yaml
```diff
@@ -548,6 +548,29 @@ recommendations:
...
+        s3:
+            "s3-personalised-covers--{instance}":
```
giorgiosironi: buckets usually have {instance} as a prefix rather than suffix, e.g. `end2end-elife-production-final`

projects/elife.yaml
stephenwf: As far as I am aware just now its serving S3 links for the Emails. I did hear something about a choice page though, so this may change in the future

PR #100, opened 2017-01-30
Add static website hosting to buckets

giorgiosironi:
To turn this on, we need a WebsiteConfiguration on the Bucket plus a related BucketPolicy to open the contents to the public

### DIFF
--- a/src/buildercore/trop.py
+++ b/src/buildercore/trop.py
@@ -356,10 +356,35 @@ def render_sqs(context, template):
 
 def render_s3(context, template):
     for bucket_name in context['s3']:
+        props = {
+            'DeletionPolicy': context['s3'][bucket_name]['deletion-policy'].capitalize()
+        }
+        bucket_title = _sanitize_title(bucket_name) + "Bucket"
+        if context['s3'][bucket_name]['website-configuration']:
+            index_document = context['s3'][bucket_name]['website-configuration'].get('index-document', 'index.html')
+            props['WebsiteConfiguration'] = s3.WebsiteConfiguration(
+                IndexDocument=index_document
+            )
+            template.add_resource(s3.BucketPolicy(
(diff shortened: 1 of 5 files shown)


giorgiosironi:
Just used this to create `error-pages--end2end` (and soon `error-pages--prod`).

src/buildercore/trop.py
```diff
@@ -360,6 +360,16 @@ def render_s3(context, template):
             'DeletionPolicy': context['s3'][bucket_name]['deletion-policy'].capitalize()
         }
         bucket_title = _sanitize_title(bucket_name) + "Bucket"
+        if context['s3'][bucket_name]['cors']:
+            props['CorsConfiguration'] = s3.CorsConfiguration(
+                CorsRules=[
+                    s3.CorsRules(
+                        AllowedHeaders=['*'],
+                        AllowedMethods=['GET', 'HEAD'],
+                        AllowedOrigins=['*']
```
lsh-0: this should be enough to satisfy @davidcmoulton requirements, cheers

giorgiosironi: it seems standard, I don't know who would be crazy enough to allow writes to a bucket from a page loaded in a website

lsh-0: maybe if it were within an intranet or secured or something. I like the idea of POSTing something to a bucket and then having the change to a bucket's contents trigger other changes, reactive style

giorgiosironi: that's called `elife-bot` and you rightly often elaborated on its level of internal quality :)

lsh-0: yeah, but the bot is a sprawling stateful mess. I'd have all of the state contained within the buckets and everything else be a dumb almost-pure function.

lsh-0:
jah, merge man.

PR #101, opened 2017-02-01
Correctly running bootstrap.sh

giorgiosironi:
Actually, we could not run bootstrap.sh in this way because it needs 4 arguments. The best way seems to delegate to `buildercore.bootstrap.update_stack()`, which contains the both `bootstrap.sh` and `highstate.sh`.


giorgiosironi:
Not a priority by the way.

lsh-0:
ha! of course. derp. I can see your changes and like the example usage. merge it if it works.

PR #103, opened 2017-03-10
added netmask within rds section. 

lsh-0:
this value can be passed directly to mysql for permissions/grants

### DIFF
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -15,8 +15,8 @@ Case 2: Ad-hoc instances
 A developer wants a temporary instance deployed for testing or debugging.
 
 """
-
 import os, json, copy
+import netaddr
 from slugify import slugify
 from . import utils, trop, core, project, context_handler
 from .utils import ensure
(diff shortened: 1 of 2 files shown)


src/buildercore/cfngen.py
```diff
@@ -83,9 +83,17 @@ def build_context(pname, **more_context): # pylint: disable=too-many-locals
     # post-processing
     if 'rds' in context['project']['aws']:
         default_rds_dbname = slugify(stackname, separator="")
+
+        # used to give mysql a range of valid ip addresses to connect from
+        net, bits = context['project']['aws']['subnet-cidr'].split('/', 1)
+        # todo: determine mask based on bits
+        mask = '255.255.255.0'
```
giorgiosironi: this is hardcoded, but should be dynamic depending on `bits`. I can't find a Python 2 standard library as `ipaddress` is in Python>=3.3. Therefore, I would use this nice and stable library: https://netaddr.readthedocs.io/en/latest/tutorial_01.html#representing-networks-and-subnets

src/buildercore/cfngen.py
```diff
@@ -83,9 +83,17 @@ def build_context(pname, **more_context): # pylint: disable=too-many-locals
...
+        netmask = "%s/%s" % (net, mask) # ll: 10.0.2.0/255.255.255.0
+
         # alpha-numeric only
         # TODO: investigate possibility of ambiguous RDS naming here
         context.update({
+            'netmask': netmask,
```
giorgiosironi: it should be possible to extend or add a test in https://github.com/elifesciences/builder/blob/master/src/tests/test_buildercore_trop.py, which tests `context` fields before using them for the CloudFormation templates

lsh-0:
fix is in, please merge this and the journal-cms-formula PR (https://github.com/elifesciences/journal-cms-formula/pull/20/files) if you're happy

PR #107, opened 2017-03-20
Journal ELBs supporting both HTTP and HTTPS

giorgiosironi:
Both will forward to Journal EC2 instances on port 80.

Incidentally, they need their own security group opening the ports that
the protocols need.


giorgiosironi:
I had this around, but not in `master`.

PR #112, opened 2017-03-24
Downgrading Fabric for Mac compatibility

giorgiosironi:
Newer version explodes on trying to compile something with Clang.

### DIFF
--- a/.activate-venv.sh
+++ b/.activate-venv.sh
@@ -18,8 +18,13 @@ if [ ! -d venv ]; then
     virtualenv --python=`which python2` venv
 fi
 
-# 'ARCHFLAGS' fixes a problem with OSX refusing to compile a dependency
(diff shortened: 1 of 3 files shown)


giorgiosironi:
This works on the Mac I have here, but should be checked on the actual Macs we use everyday.

lsh-0:
does anything on here help? http://www.fabfile.org/installing.html#dependencies

giorgiosironi:
Yes, there are a few because people use them to develop for iOS:
http://www.macincloud.com/
https://xcloud.me/
Would still make sense to let people use this version on their Macs until the issue with the latest Fabric is resolved (this was the Fabric version in use until last week)

lsh-0:
thanks for links. I've made a tweak, if you're happy with it, please merge.

giorgiosironi:
Tested installation on a Mac, seems to work, merging.

PR #119, opened 2017-04-03
CloudFront template section can be updated

giorgiosironi:
Also, when giving up polling, do not delete EC2 SSH key.


giorgiosironi:
There were some fixes for tests failing in the mainline in here, merging.

PR #120, opened 2017-04-03
Configuring error codes in CloudFront

giorgiosironi:
Pages are hosted into a publicly readable bucket, CloudFront allows to:
- point all requests that match `???.html` to this bucket
- specify a fixed set of error codes to be mapped to these paths.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -214,8 +214,6 @@ journal:
                     pattern: "???.html"
                     codes:
                         - 404
-                        # not supported
-                        # - 410 
                         - 503
         continuumtest:
             cloudfront:
@@ -226,8 +224,6 @@ journal:
                     pattern: "???.html"
                     codes:
                         - 404
-                        # not supported
-                        # - 410 
                         - 503
         prod:
             description: prod environment for journal. ELB on top of multiple EC2 instances
@@ -247,8 +243,6 @@ journal:
                     pattern: "???.html"
                     codes:
                         - 404
-                        # not supported
-                        # - 410 
                         - 503
     vagrant:
         ram: 4096
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -220,7 +220,7 @@ journal:
                 subdomains:
                     - "{instance}--cdn-journal"
                 errors: 
-                    domain: prod-elife-error-pages.s3-website-us-east-1.amazonaws.com
+                    domain: continuumtest-elife-error-pages.s3-website-us-east-1.amazonaws.com
                     pattern: "???.html"
                     codes:
                         - 404
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -213,8 +213,7 @@ journal:
                     domain: end2end-elife-error-pages.s3-website-us-east-1.amazonaws.com
                     pattern: "???.html"
                     codes:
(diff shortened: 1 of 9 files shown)


projects/elife.yaml
```diff
@@ -223,6 +238,18 @@ journal:
...
+                subdomains:
+                    - "{instance}--cdn-journal"
+                    - journal # in the future 'beta' or... ''?
+                errors: 
+                    domain: prod-elife-error-pages.s3-website-us-east-1.amazonaws.com
+                    pattern: "???.html"
+                    codes:
+                        - 404
+                        # not supported
+                        # - 410
```
giorgiosironi: @thewilkybarkid we won't have 410 according to the docs

projects/elife.yaml
```diff
@@ -205,15 +206,29 @@ journal:
...
+                        - 404
+                        # not supported
+                        # - 410 
+                        - 503
+        continuumtest:
+            cloudfront:
+                subdomains:
+                    - "{instance}--cdn-journal"
+                errors: 
+                    domain: prod-elife-error-pages.s3-website-us-east-1.amazonaws.com
```
giorgiosironi: doesn't have its own `error-pages--continuumtest` instance, maybe it should

src/buildercore/trop.py
```diff
@@ -546,6 +546,36 @@ def render_cloudfront(context, template, origin_hostname):
             SslSupportMethod='sni-only'
         )
     }
+    if context['cloudfront']['errors']:
+        props['Origins'].append(cloudfront.Origin(
+            DomainName=context['cloudfront']['errors']['domain'],
+            # TODO: constant
+            Id='errors',
```
giorgiosironi: yeah, I need to extract that constant

projects/elife.yaml
thewilkybarkid: Makes sense that CloudFront wouldn't produce it.

src/buildercore/trop.py
```diff
@@ -546,6 +546,36 @@ def render_cloudfront(context, template, origin_hostname):
...
+            cloudfront.CacheBehavior(
+                TargetOriginId='errors',
+                ForwardedValues=cloudfront.ForwardedValues(
+                    QueryString=False
+                ),
+                PathPattern=context['cloudfront']['errors']['pattern'],
+                ViewerProtocolPolicy='allow-all',
+            ),
+        ]
+        all_codes = [400, 403, 404, 405, 414, 416, 500, 501, 502, 503, 504]
```
thewilkybarkid: Reading the CloudFront docs again, does this mean that a `404 Not Found` from Journal will result in `/404.html` being returned instead? We _only_ want these custom error pages for CloudFront-produced errors.

giorgiosironi: Then we should only configure `502` according to the docs: http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/troubleshooting-response-errors.html
However, I've seen forum reports for 503 and 504, so maybe all three?

thewilkybarkid: Hmm, probably not `503` and `504`, Journal can produce those.

`414` probably should be too.

(Again, something that should be easier with Fastly.)

PR #122, opened 2017-04-05
Updating the ELB HealthCheck in place

giorgiosironi:
We go from opening a TCP connection to port 80 to the EC2 instances (which fails if the DNS is down) to do a HTTP request for /ping (which fails if the PHP bootstrap fails).

The result of this HealthCheck failing is for the instance to be taken out of the ELB. If all machines are taken out, the ELB starts returning 503.

In theory we can update this in the CloudFormation template with `No Interruption` so the downstream DNS for the ELB should not change. Let's see.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -68,7 +68,9 @@ defaults:
             idle_timeout: 60
             certificate: arn:aws:iam::512686554592:server-certificate/cloudfront/wildcard.elifesciences.org/wildcard.elifesciences.org
             healthcheck:
+                protocol: http
                 port: 80
+                path: /ping
                 timeout: 4 
                 interval: 5
                 unhealthy_threshold: 2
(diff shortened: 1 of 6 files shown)


projects/elife.yaml
```diff
@@ -68,7 +68,9 @@ defaults:
             idle_timeout: 60
             certificate: arn:aws:iam::512686554592:server-certificate/cloudfront/wildcard.elifesciences.org/wildcard.elifesciences.org
             healthcheck:
+                protocol: http
                 port: 80
+                path: /ping
```
thewilkybarkid: `/ping` or `/status`? The former's only guaranteed to check that the service is running (though due to architecture checks a certain amount of components), but the latter should check (as much as possible) that every component is working (eg Journal checks that it can talk to the API).

giorgiosironi: I would go for being conservative with `/ping`, since otherwise any backend having problems would take down journal, while we have pages that tried to degrade the experience.

thewilkybarkid: 👍

PR #124, opened 2017-04-05
Allowing update of template of existing EC2 instances

giorgiosironi:
But not messing with adding new EC2 instances. This will give us the possibility, in case we need, to change the type of an EC2 instance. It's used in the specialized task `update_template` and it requires confirmation like all CloudFormation template updates. Requires downtime as the reboot is performed.


giorgiosironi:
If not used carefully true, but at least the changes have to be captured in the elife.yaml configuration rather than being arbitrarily done through a console

PR #126, opened 2017-04-13
CORS for elife-published bucket

giorgiosironi:
This was originally manually set, change is brought into the builder
configuration and deployed through `bldr update_template`.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -350,6 +350,7 @@ elife-bot:
             "{instance}-elife-published":
                 deletion-policy: retain
                 public: true
+                cors: True
     aws-alt:
         #prod:
         #    eip: 54.164.145.166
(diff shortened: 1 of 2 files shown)


projects/elife.yaml
```diff
@@ -350,6 +350,7 @@ elife-bot:
             "{instance}-elife-published":
                 deletion-policy: retain
                 public: true
+                cors: True
```
gnott: Should this be lower case?

cors: true

giorgiosironi: ah yeah, it's yaml; concidentally everything different from `false` works as truthy, but I'll amend

giorgiosironi: appears everything is supported http://yaml.org/type/bool.html but better to be consistent

PR #132, opened 2017-05-02
feature: support for custom rds parameter groups

lsh-0:


src/buildercore/trop.py
```diff
@@ -200,15 +219,17 @@ def rdsinstance(context, template):
         'DBSubnetGroupName': Ref(rsn),
         'DBInstanceClass': lu('project.aws.rds.type'),
         'Engine': lu('project.aws.rds.engine'),
+        # something is converting this value to an int from a float :(
```
giorgiosironi: is this still true? I can only reproduce getting out a string:
```
>>> from buildercore import context_handler
>>> c = context_handler.load_context("lax--prod")                                    
>>> from buildercore import utils
>>> utils.lu(c, 'project.aws.rds.version')
u'9.4'
```

src/tests/fixtures/projects/dummy-project.yaml
```diff
@@ -45,6 +45,7 @@ defaults:
             backup-retention: 28 # days
             # two subnets are required in two different availability zones
             # http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbsubnet-group.html
+            params: []
```
giorgiosironi: you'll need to add this default to `projects/elife.yaml` of course

lsh-0: of course - hence the leading `donotmerge` ;)

src/buildercore/trop.py
lsh-0: yeah, I just updated the comment from my memory of the problem. it's why the engine version is quoted `'9.4'` in the elife.yaml projects file. I'll remove if it's not an issue

PR #133, opened 2017-05-02
Generic CDN

giorgiosironi:
It is a CloudFront CDN with multiple, custom origins. 

- Supporting a default origin (the first one specified)
- Supporting parameterization of hostnames
- Supporting path pattern for custom origins (typically from the second cache behavior onwards)

The goal is to model with builder the cdn.elifesciences.org snowflake, at least for testing environments.

### DIFF
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -17,6 +17,7 @@ A developer wants a temporary instance deployed for testing or debugging.
 """
 import os, json, copy
 import re
+from collections import OrderedDict
 import netaddr
 from slugify import slugify
 from . import utils, trop, core, project, bootstrap, context_handler
@@ -185,7 +186,10 @@ def build_context_cloudfront(context, parameterize):
             'headers': context['project']['aws']['cloudfront']['headers'],
             'default-ttl': context['project']['aws']['cloudfront']['default-ttl'],
             'errors': errors,
-            'origins': { o_id: { 'hostname': parameterize(o['hostname']) } for o_id, o in context['project']['aws']['cloudfront']['origins'].iteritems()},
+            'origins': OrderedDict([
+                (o_id, { 'hostname': parameterize(o['hostname']), 'pattern': o.get('pattern') }) \
+                for o_id, o in context['project']['aws']['cloudfront']['origins'].iteritems()
+            ]),
         }
     else:
         context['cloudfront'] = False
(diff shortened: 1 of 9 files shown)


giorgiosironi:
Just used to create `generic-cdn--continuumtest`:
https://continuumtest-cdn.elifesciences.org/articles/00666/elife-00666-fig1-v1.jpg
https://continuumtest-cdn.elifesciences.org/articles/00666/elife-00666-v1.pdf
`generic-cdn--end2end` is on the way.

src/buildercore/cfngen.py
```diff
@@ -185,6 +186,10 @@ def build_context_cloudfront(context, parameterize):
             'headers': context['project']['aws']['cloudfront']['headers'],
             'default-ttl': context['project']['aws']['cloudfront']['default-ttl'],
             'errors': errors,
+            'origins': OrderedDict([
+                (o_id, {'hostname': parameterize(o['hostname']), 'pattern': o.get('pattern')})
+                for o_id, o in context['project']['aws']['cloudfront']['origins'].iteritems()
+            ]),
```
lsh-0: this is a little gnarly on first blush, but I get it now after a few reads. I (personally) would have done a slightly more verbose version like:

```
def tf(oitem):
    o_id, o = oitem
    return (o_id, {
        'hostname': parameterize(o['hostname']), 
        'pattern': o.get('pattern')
    }
OrderedDict(map(tf, context['project']['aws']['cloudfront']['origins'].items()))
```

but it's fine. pretty sure my inline functions are just as incomprehensible sometimes, too. Stick with `.items()` though. No need for an iterator there. In Python 3 there is no more `iteritems` as `items` becomes lazy.

I thought you might have had a problem assuming the order of the origins are preserved, but I'm that yaml is read in using `buildercore.utils.ordered_load`, so there shouldn't be a problem there either.

src/buildercore/trop.py
```diff
@@ -574,24 +602,25 @@ def render_cloudfront(context, template, origin_hostname):
...
                 ResponsePagePath=page
             ) for code, page in context['cloudfront']['errors']['codes'].iteritems()
         ]
+
+    if context['cloudfront']['origins']:
+        props['CacheBehaviors'] = [
+            _cache_behavior(o_id, o['pattern'])
+            for o_id, o in context['cloudfront']['origins'].iteritems()
+            if o['pattern']
+        ]
```
lsh-0: ha - the long-form list comprehension again. who knows, maybe I'll get used to it.

PR #141, opened 2017-05-26
Extending stickiness

giorgiosironi:
Journal should use stickiness related to its own `journal` cookie.


giorgiosironi:
It would be a bit imposing on the project to have to use a cookie named like itself, it is not a cookie generated by the AWS infrastructure but by the application code itself; it could be `JSESSIONID` in Java projects for example.

lsh-0:
fair enough

PR #143, opened 2017-06-02
Support for custom CNAMEs

giorgiosironi:
I once tried to do this automatically for production, but it's too restricting. Easier to specify the subdomains in an explicit way.

In this case, we add ui-patterns.elifesciences.org as a CNAME to prod--ui-patterns.elifesciences.org

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -98,6 +98,7 @@ defaults:
             headers: []
             errors: null
             default-ttl: 0 # seconds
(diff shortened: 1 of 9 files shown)


giorgiosironi:
Seems an easy thing to do, but the implementation can be complex to cater for all cases like CNAME to ELB vs CNAME to EC2, external or internal DNS (only externals are supported). If the review can find holes, all the better.

PR #146, opened 2017-06-13
Feedback from setting up builder on a brand new EC2 instance for thedailybugle.org

giorgiosironi:


giorgiosironi:
The mainline failed without these changes: https://ci--alfred.elifesciences.org/job/test-builder/367/console
But also this build failed.

PR #148, opened 2017-06-13
Updating documentation for eLife 2.0 infrastructure

giorgiosironi:
Apart from the text, some feedback from `thedailybugle.org` which runs now an instance of journal.

### DIFF
--- a/.prerequisites.py
+++ b/.prerequisites.py
@@ -73,6 +73,7 @@ mac_checks = [
 ]
 
 def run_checks(check_list, exclusions=[]):
+    failed_checks = 0
     for cmd in check_list:
         installed_checker = dumb_install_check
         version_checker = dumb_version_check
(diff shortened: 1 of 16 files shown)


.prerequisites.py
```diff
@@ -97,8 +98,11 @@ def run_checks(check_list, exclusions=[]):
             sys.stdout.write('NOT found. Try:\n')
             for opsys, suggestion in install_suggestions.items():
                 print('   %s: %s' % (opsys, suggestion))
+            failed_checks = failed_checks + 1
         sys.stdout.flush()
 
+    exit(failed_checks)
```
lsh-0: ha - I like this. the exit code corresponds to the number of problems

Jenkinsfile
```diff
@@ -1,12 +1,15 @@
...
-    checkout scm
+    stage 'Checkout', {
+        checkout scm
+    }
 
     lock('builder') {
-        stage 'Update'
-        sh './update.sh --exclude virtualbox vagrant'
+        stage 'Update', {
+            sh './update.sh --exclude virtualbox vagrant ssh-credentials ssh-agent'
```
lsh-0: perhaps an `--exclude-all` or `--no-checks` is in the future

PR #150, opened 2017-06-20
Redirects project

giorgiosironi:
- Support removal of resources in CloudFormation templates
- CNAMEs title should be starting from 1 like other titles
- Configuration for redirects project
- Added domains to future redirects--prod instance
- `git ci` was not recognized as a command

Already used to remove the domains from `journal--prod`, instances of `redirects` have not been created yet because they need the configuration to reach the salt master first.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -964,15 +964,18 @@ redirects:
             - 443
     aws-alt:
         prod:
-            subdomains:
-                - elifesciences.net
-                - e-lifesciences.org
-                - e-lifesciences.net
-                - elifejournal.net
-                - e-lifejournal.org
-                - e-lifejournal.com
-                - e-lifejournal.net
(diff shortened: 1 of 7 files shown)


projects/elife.yaml
```diff
@@ -959,3 +950,34 @@ iiif:
...
+    aws:
+        type: t2.nano
+        ports:
+            - 22
+            - 80
+            - 443
+    aws-alt:
+        prod:
+            cloudfront:
+                certificate_id: ASCAIEHY4AGS453E4IUJK
```
lsh-0: nothing sensitive here?

giorgiosironi: same as https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L96 I understood this is just an ID, and without the right AWS credentials/account is not accessible

PR #152, opened 2017-06-22
Updating generic-cdn to keep it identical to prod

giorgiosironi:
generic-cdn--prod is not managed by builder and has a DefaultTTL of 86400 (1 day). Applying it to other environments


src/buildercore/cfngen.py
```diff
@@ -406,7 +406,10 @@ def merge_delta(stackname, delta_plus, delta_minus):
 def apply_delta(template, delta_plus, delta_minus):
     for component in delta_plus:
         ensure(component in ["Resources", "Outputs"], "Template component %s not recognized", component)
-        template[component].update(delta_plus[component])
```
lsh-0: really minor thing, but this:
```
data = template.get(component, {})
data.update(delta_plus[component])
template[component] = data
```
avoids any forking in logic and coverage overhead. I don't think there is any complexity checking in builder yet.

PR #154, opened 2017-07-03
Allow missing context file

giorgiosironi:
We have old servers like master-server--2016-07-22 that have no context file stored on S3. Strange, we could repair them, but in general we may have the problem if the settings impose to not upload the context in the S3 bucket; at the moment 'not uploading anything to S3' is the only builder setting supported for organizations different from eLife as the bucket name is hardcoded

This is an attempt to be more conservative and still allow `bldr ssh`
and `bldr update` to work. However we may want a different direction
like making context files mandatory (hence if you have no S3 upload you
have to use the builder from a single place) and repair the missing
ones we do not have.

### DIFF
--- a/src/buildercore/context_handler.py
+++ b/src/buildercore/context_handler.py
@@ -20,7 +20,7 @@ def load_context(stackname):
     path = local_context_file(stackname)
     if not os.path.exists(path):
         if not download_from_s3(stackname):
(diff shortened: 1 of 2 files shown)


lsh-0:
this may be relevant: https://elifesciences.atlassian.net/browse/ELPP-2954

giorgiosironi:
Old builder instance shouldn't be related to this, the master server involved is under the new builder.

PR #156, opened 2017-07-04
ElastiCache: private Redis node with customizable version and type

giorgiosironi:
Will expand upon this after experimenting with real eLife projects stack instances.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -109,6 +109,8 @@ defaults:
                 - subnet-20275c68 # elife-cache-subnet-1
                 - subnet-c4033af8 # elife-cache-subnet-2
             version: "2.8.24" # closest to builder-base-formula redis.sls
+            configuration:
+                maxmemory-policy: volatile-ttl
 
     aws-alt:
         fresh:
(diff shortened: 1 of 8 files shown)


giorgiosironi:
Test instance up as `journal--cachetest`:
```
elife@continuumtest--journal:~$ redis-cli -h jou-el-1h7ejszvo06ps.yq8mzu.0001.use1.cache.amazonaws.com
jou-el-1h7ejszvo06ps.yq8mzu.0001.use1.cache.amazonaws.com:6379> KEYS ALL
(empty list or set)
```
(connecting from there as there is no redis-cli on `journal--cachetest`)

giorgiosironi:
Missing context: every journal EC2 instance has its own local Redis. It uses it for caches, but this could be more efficient if we had a centralized cache that all journal--prod instances can use and invalidate. Moreover, having a centralized Redis allows to use the cache to store session information, while now we have to resort to client affinity and send the user to the same instance if he is in a session (used sparingly in forms but will be used more in the future, for logging in).

lsh-0:
ah - righto. thanks

thewilkybarkid:
Also allows the cache to persist across deploys, so the HTTP cache isn't invalided on every highstate.

PR #162, opened 2017-07-19
Feature: prod uses multiaz

lsh-0:
* production instances using RDS now use multiple availability zones going forwards
* `cfn.update_template` now refreshes the buildvars on any EC2 instances after CloudFormation template has finished updating but before Salt's highstate is called
* `buildvars.revision` is preserved

### DIFF
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -329,7 +329,7 @@ def regenerate_stack(pname, **more_context):
     return context, delta_plus, delta_minus
 
 
-UPDATABLE_TITLE_PATTERNS = ['^CloudFront.*', '^ElasticLoadBalancer.*', '^EC2Instance.*', '.*Bucket$', '.*BucketPolicy', '^StackSecurityGroup$', '^ELBSecurityGroup$', '^CnameDNS.+$']
+UPDATABLE_TITLE_PATTERNS = ['^CloudFront.*', '^ElasticLoadBalancer.*', '^EC2Instance.*', '.*Bucket$', '.*BucketPolicy', '^StackSecurityGroup$', '^ELBSecurityGroup$', '^CnameDNS.+$', '^AttachedDB$']
 REMOVABLE_TITLE_PATTERNS = ['^CnameDNS\\d+$', '^ExtDNS$']
 EC2_NOT_UPDATABLE_PROPERTIES = ['ImageId', 'Tags', 'UserData']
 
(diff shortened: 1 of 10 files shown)


src/buildvars.py
```diff
@@ -119,6 +120,37 @@ def _update_remote_bvars(stackname, buildvars):
...
-    return FULL
+@requires_aws_stack
+def refresh(stackname, context):
+    "(safely) replaces the buildvars file on the ec2 instance(s)"
+
+    def _refresh_buildvars():
+        old_buildvars = _retrieve_build_vars()
+
+        node = old_buildvars.get('node')
+        if not node or not str(node).isdigit():
```
lsh-0: I did something stupid here and `node` was set to `None` at some point so a lot of the code below is just to handle bad node values that probably won't ever exist again.

src/buildercore/cfngen.py
```diff
@@ -329,7 +329,7 @@ def regenerate_stack(pname, **more_context):
     return context, delta_plus, delta_minus
 
 
-UPDATABLE_TITLE_PATTERNS = ['^CloudFront.*', '^ElasticLoadBalancer.*', '^EC2Instance.*', '.*Bucket$', '.*BucketPolicy', '^StackSecurityGroup$', '^ELBSecurityGroup$', '^CnameDNS.+$']
+UPDATABLE_TITLE_PATTERNS = ['^CloudFront.*', '^ElasticLoadBalancer.*', '^EC2Instance.*', '.*Bucket$', '.*BucketPolicy', '^StackSecurityGroup$', '^ELBSecurityGroup$', '^CnameDNS.+$', '^AttachedDB$']
```
giorgiosironi: we should add `^AttachedDBSubnet$` too here if you agree

src/buildvars.py
```diff
@@ -97,17 +104,11 @@ def _force_single_ec2_node():
...
-    with hide('everything'):
-        buildvars = _validate()
-    print 'found build vars'
-    print
-    return buildvars
-
 def _update_remote_bvars(stackname, buildvars):
     LOG.info('updating %r with new vars %r', stackname, buildvars)
-    assert core_utils.hasallkeys(buildvars, ['branch'])  # , 'revision']) # we don't use 'revision'
+    # TODO: do we use 'branch' anymore? if not, remove
```
giorgiosironi: I think not all projects have a `revision` (e.g. `api-gateway`). We may not be using `branch` anymore.

lsh-0: that's good to know

src/buildercore/cfngen.py
lsh-0: sounds good, adding it in

PR #165, opened 2017-08-02
Trying to replace Python 2.7 PPA

giorgiosironi:
The current one seems not available anymore. This solution causes other problems during bootstrap.sh execution


lsh-0:
what a mess. and only a few years since release. this entropy isn't great.

PR #167, opened 2017-08-08
Feature: masterless instances

lsh-0:


src/cfn.py
```diff
@@ -125,9 +126,9 @@ def helpfn(altkey):
 # these aliases are deprecated
 @task(alias='aws_launch_instance')
 @requires_project
-def launch(pname, instance_id=None):
+def launch(pname, instance_id=None, alt_config=None):
```
lsh-0: allows: `./bldr launch:lax,demo,masterless`

projects/elife.yaml
```diff
@@ -833,6 +835,12 @@ observer:
                     - bus-articles--{instance}
                     - bus-metrics--{instance}
     aws-alt:
+        masterless:
```
giorgiosironi: would rename the `masterless` key here (which is just the environment name) to avoid confusion with the `masterless: true` configuration, e.g. `masterless-sample`

scripts/bootstrap.sh
```diff
@@ -111,6 +108,10 @@ fi
 if $installing; then echo "$(date -I) -- installed $version" >> /root/events.log; fi
 if $upgrading; then echo "$(date -I) -- upgraded to $version" >> /root/events.log; fi
 
+if [ "$master_ipaddr" = "masterless" ]; then
+    # ignore given IP parameter and use the one we can detect
+    master_ipaddr=$(ifconfig eth0 | awk '/inet / { print $2 }' | sed 's/addr://')
```
giorgiosironi: why not just `127.0.0.1`?

scripts/bootstrap.sh
```diff
@@ -119,33 +120,35 @@ master: $master_ipaddr
...
     # by default the project's top.sls is disabled by file naming. hook that up here
     cd /srv/salt/ && ln -sf example.top top.sls
-    
+
     # vagrant makes all formula dependencies available, including builder base formula
 
     # overwrite the general purpose /etc/salt/minion file created above 
-    # with this more complex one for dev environments only
+    # with a custom one for masterless environments only
+    # TODO: will break on remote masterless instances
```
giorgiosironi: we are in the Vagrant case here so why talking about `masterless`?

src/buildercore/bootstrap.py
```diff
@@ -430,23 +455,25 @@ def _update_ec2_node():
...
-        # /TODO
+
+        if is_masterless:
+            # order is important.
+            formula_list = '"%s"' % ' '.join(pdata.get('formula-dependencies', []) + [pdata['formula-repo']])
+            prepo = pdata['private-repo']
+            # to init the builder-private formula, the masterless instance needs
+            # the master-builder key
+            upload_master_builder_key(master_builder_key)
+            run_script('init-formulas.sh', formula_list, prepo)
```
giorgiosironi: neat

src/buildercore/trop.py
```diff
@@ -148,7 +148,14 @@ def mkoutput(title, desc, val):
 
 def build_vars(context, node):
     buildvars = dict(context)
-    del buildvars['project']
+
+    # preseve some of the project data. all of it is too much
```
giorgiosironi: why? sensitive?

giorgiosironi: more like.. information hiding (as a good thing)

src/masterless.py
```diff
@@ -0,0 +1,60 @@
...
+from buildercore.utils import ensure
+import logging
+from functools import wraps
+LOG = logging.getLogger(__name__)
+
+def requires_master_server_access(fn):
+    @wraps(fn)
+    def wrapper(*args, **kwargs):
+        import master
+        ensure(master.server_access(), "this command requires access to the master server. you don't have it.")
```
giorgiosironi: can this be cached inside the current process? If multiple tasks each requiring access are involved, no reason to check it more than once, for better performance

src/masterless.py
```diff
@@ -0,0 +1,60 @@
...
+        return fn(*args, **kwargs)
+    return wrapper
+
+@task
+@requires_project
+@requires_master_server_access
+def launch(pname, instance_id=None):
+    import cfn
+    cfn.launch(pname, instance_id, 'masterless')
+    # opportunity to do post-launch things here
```
giorgiosironi: good Facade

src/masterless.py
```diff
@@ -0,0 +1,60 @@
...
+@task
+@requires_aws_stack
+def ssh(stackname, node=None):
+    "maintenance ssh. uses the pem key and the bootstrap user to login."
+    import cfn
+    instances = core.find_ec2_instances(stackname)
+    public_ip = cfn._pick_node(instances, node).ip_address
+    # -i identify file
+    local("ssh %s@%s -i %s" % (config.BOOTSTRAP_USER, public_ip, core.stack_pem(stackname)))
+
```
giorgiosironi: will need a way to set the formula to a branch, but that can be done later

scripts/bootstrap.sh
lsh-0: vagrant is masterless

src/buildercore/trop.py
lsh-0: yeah, I think we've talked about the buildvars before and keeping them minimal. I ended up going with passing the formulas to the init-formulas.sh script, but originally I was going to pull them from the buildvars. I'm still  undecided whether I should store the pinned revisions for the formulas in the buildvars  or even store them at all.

src/masterless.py
lsh-0: sounds do-able

src/masterless.py
lsh-0: I'm working on that today actually. Meant to do it Weds but caught up debugging and polishing this and the ubuntu 16.04 work.

src/masterless.py
lsh-0: you asked in the comments above about changing the name of the alt-aws config from 'masterless' to 'masterless-sample' - which is fine we can do that - but if we have a 'masterless' alt config like we have an 'end2end' or 'prod' config, then we can target those masterless instances in here.

although saying that, the only thing making a instance masterless right now is the `aws.ec2.masterless: True` key. I'd be wary of introducing anything else in `launch` that couldn't be done with regular `cfn.launch`. I like the fact we could have a `masterless-test` altconfig later on.

scripts/bootstrap.sh
lsh-0: erm. I don't know. lost to history? I'll see if `127.0.0.1` works and if so, will pass that around instead of `masterless`.

projects/elife.yaml
lsh-0: it's really not that confusing. I've elaborated on why `masterless` a little further down. It's not a hard and fast decision and it buys us some convenience.

src/masterless.py
lsh-0: done

scripts/bootstrap.sh
lsh-0: shifted that bit of vagrant wrangling to it's own formula and removed comment

scripts/bootstrap.sh
lsh-0: turns out vagrant and masterless don't even use that value, it gets replaced almost immediately with a different minion file. I've made the fourth parameter (master ipaddr) optional and removed this section and some wrangling in bootstrap.py

src/masterless.py
giorgiosironi: exactly, the `masterless` `aws-alt` config and the `masterless` `ec2` options are two different things, so I would like to name them with two different names

scripts/init-vagrant-formulas.sh
```diff
@@ -0,0 +1,36 @@
...
+# overwrite the general purpose /etc/salt/minion file created above 
+# with a custom one for masterless environments only
+cp /vagrant/scripts/salt/minion.template /etc/salt/minion
+custom_minion_file="/vagrant/scripts/salt/$minion_id.minion"
+if [ -e "$custom_minion_file" ]; then
+    # this project requires a custom minion file. use that instead
+    cp "$custom_minion_file" /etc/salt/minion
+else
+    echo "couldn't find $custom_minion_file"
+fi
```
giorgiosironi: I like the isolation of this script from non-Vagrant environments; a lot goes on in bootstrap.sh

scripts/update-master-formula.sh
```diff
@@ -1,19 +1,33 @@
...
     git pull --rebase
 else
-    cd /opt/formulas/
+    cd "$formula_root"
     git clone "$formula" "$pname"
 fi
+
+cd "$formula_path"
+
+if [ "$revision" != "" ]; then
```
giorgiosironi: may need a `git fetch` before, it this performed previously?
Should be needed in all cases where the commit to be checked out is not in the current branch

src/masterless.py
```diff
@@ -46,14 +56,50 @@ def ssh(stackname, node=None):
...
+        if repo not in known_formula_map:
+            print 'skipping %r, unknown formula. known formulas: %s' % (repo, ', '.join(known_formula_map.keys()))
+            continue
+
+        arglist.append((repo, known_formula_map[repo], rev))
+
+    if not arglist:
+        return 'nothing to do'
+
+    with core.stack_conn(stackname):
```
giorgiosironi: I think `stack_conn` doesn't cater for multiple nodes and `stack_all_ec2_nodes` should be used instead

PR #169, opened 2017-08-16
Feature - initial ubuntu 16.04 support

lsh-0:


projects/elife.yaml
```diff
@@ -121,6 +121,7 @@ defaults:
             ec2:
                 ami: ami-9eaa1cf6 # Ubuntu 14.04
     vagrant:
```
giorgiosironi: would include a reason why this is better than the Canonical one (long-standing bug you posted a while ago)

scripts/bootstrap.sh
```diff
@@ -44,20 +44,29 @@ fi
...
-    # let's remove it to avoid apt-get update errors
-    rm -f /etc/apt/sources.list.d/fkrull-deadsnakes-python2_7-trusty.list
-    # provides python2.7[.13] package and some dependencies
-    add-apt-repository ppa:jonathonf/python-2.7
-    # provides a recent python-urllib3 (1.13.1-2) because:
-    # libpython2.7-stdlib : Breaks: python-urllib3 (< 1.9.1-3) but 1.7.1-1ubuntu4 is to be installed
-    # due to the previous PPA
-    add-apt-repository ppa:ross-kallisti/python-urllib3
+if ! command python2.7; then
+    # python not found
```
giorgiosironi: I guess because only Python 3 is there... maybe rename the variable to `upgrade_python2.7`

projects/elife.yaml
lsh-0: done

PR #170, opened 2017-08-16
Update ext volumes

giorgiosironi:
This recreates the volumes as empty! It's the only way to update their type. The use case is iiif, which uses them as a cache, so no problem if the Salt highstate is run soon after.
e.g.
```
2017-08-16 13:44:38,662 - INFO - MainProcess - cfn - Create/update: {'Outputs': {},
 'Resources': {u'ExtraStorage1temp': {u'Properties': {u'AvailabilityZone': {u'Fn::GetAtt': [u'EC2Instance1',
                                                                                            u'AvailabilityZone']},
                                                      u'Size': u'30',
                                                      u'VolumeType': u'standard'},
                                      u'Type': u'AWS::EC2::Volume'},
               u'MountPoint1temp': {u'Properties': {u'Device': u'/dev/sdh',
                                                    u'InstanceId': {u'Ref': u'EC2Instance1'},
                                                    u'VolumeId': {u'Ref': u'ExtraStorage1temp'}},
                                    u'Type': u'AWS::EC2::VolumeAttachment'}}}
2017-08-16 13:44:38,663 - INFO - MainProcess - cfn - Delete: {'Outputs': {},
 'Resources': {u'ExtraStorage1': {u'Properties': {u'AvailabilityZone': {u'Fn::GetAtt': [u'EC2Instance1',
                                                                                        u'AvailabilityZone']},
                                                  u'Size': u'30',
                                                  u'VolumeType': u'standard'},
                                  u'Type': u'AWS::EC2::Volume'},
               u'MountPoint1': {u'Properties': {u'Device': u'/dev/sdh',
                                                u'InstanceId': {u'Ref': u'EC2Instance1'},
                                                u'VolumeId': {u'Ref': u'ExtraStorage1'}},
                                u'Type': u'AWS::EC2::VolumeAttachment'}}}
```


giorgiosironi:
Seems to produce the right delta on update and remove, remains to be seen whether it will work when passed to CloudFormation.

projects/elife.yaml
```diff
@@ -937,6 +937,8 @@ iiif:
                 cluster-size: 2
             ext:
                 size: 30 # GB
+                suffix: 
+                    1: temp
```
lsh-0: can we get a comment explaining what `1:` is in this map?

lsh-0:
> remains to be seen whether it will work when passed to CloudFormation.
can we test that before merging?

giorgiosironi:
Thanks for the review, will work to test this on iiif--end2end.

giorgiosironi:
(won't be able to do this today)

giorgiosironi:
Tried updating iiif--end2end, but it's a mess. The reason is the device `/dev/sdh` is busy, and you need to unmount it first, and at least make a 2-step update where the new volume being created has a different `device`; it would not be picked up by the formula then.
I'll rework this to:
- remove the volume on 1 (leaving the node possibly broken, but it will have a stopped nginx)
- add the volume again on 1
- remove the volume on 2
- add the volume again on 2

giorgiosironi:
You wish... adding and removing volumes while they are mounted bricked one of the iiif--end2end servers. Now I am implementing Phoenix Servers, so that I can remove it (done, easier) and then add it again (more tricky) to the template without changing the other servers.

giorgiosironi:
Worked to terminate and recreate `iiif--end2end--1`. I'll have to make the `ext`-related resource non-updataable, so that we can do:
- suppressed `1`, apply (total nodes: 1, node 2 is not touched)
- suppressed [], apply (total nodes: 2, node 1 has the new disk)
- suppressed `2`, apply (total nodes: 1, node 1 is not touched)
- suppressed [], apply (total nodes: 2, node 1 and 2 have the new disk)

giorgiosironi:
Bug: when shrinking the cluster to 1 instance, the DNS was changed to point directly to that, and never recovered. We should use the `elb` configuration to decide to point the DNS to the `elb`, not the number in `ec2.cluster-size`.

giorgiosironi:
Experimented with `ci` and the process of `suppressing` one instance and recreating it with an `override` works *for the CloudFormation part*.

Now trying to solve the Salt highstate that follows the recreation: not only we have to remove `/etc/salt/pki/master/minions/$stackname` from the `master-server`; `prep-stack.sh` is also not run, hence the EC2 instance has some leftover files from the AMI. Trying to get it to run only on new stacks, possible strategies:
- with the `UserData` (don't know yet if possible)
- explicitly running it, selecting in some way the right instances

giorgiosironi:
However, this partial solution is good enough to be merged, I'll work on seamless recreation in an additional branch.

PR #171, opened 2017-08-21
bugfix: bootstrap context data missing for older instances

lsh-0:
works around the fact that some instances may be missing context data. it instead inspects the project data for that stack. this is not great because the stack being updated my have been an alternate configuration to the default/static one generated by `project_data_for_stack`.

this isn't a big deal and is what `update_ec2_stack` had been doing up until very recently

### DIFF
--- a/src/buildercore/bootstrap.py
+++ b/src/buildercore/bootstrap.py
@@ -434,11 +434,16 @@ def update_ec2_stack(stackname, concurrency):
     script that can be downloaded from the web and then very conveniently
     installs it's own dependencies. Once Salt is installed we give it an ID
     (the given `stackname`), the address of the master server """
-    #pdata = project_data_for_stackname(stackname)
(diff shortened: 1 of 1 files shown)


lsh-0:
> elife-dashboard--end2end.json

this was the troublemaker. I assumed it was missing context data, but instead it could simply be really old data and lacking the `ec2` section that was introduced a long time ago

giorgiosironi:
Indeed, it's missing it. Perhaps we can default the `ec2` key to `True` rather than the whole context? If it's missing, it's from a time where only stacks with EC2 instances were supported.

lsh-0:
I could do that. I might write a small something that identifies all these legacy instances.

giorgiosironi:
Encountered on master and fixed with https://github.com/elifesciences/builder/commit/49336e76aa06ab3ef2f417fcc05176b0340015f5

PR #175, opened 2017-08-29
ELPP-3199 added peerscout config

de-code:


projects/elife.yaml
```diff
@@ -918,6 +918,32 @@ reviewer-suggestions:
             1241: 80
         ram: 4096
 
+peerscout:
+    description: PeerScout
```
giorgiosironi: not much of a description like for many other projects :smile_cat:

giorgiosironi:
Would probably add a `DEPRECATED` comment to `reviewer-suggestions`

projects/elife.yaml
de-code: Want to keep it consistent ;-)

giorgiosironi:
Merging on green, but not creating new instances until `peerscout-formula` is devoid of `reviewer-suggestions` references.

PR #176, opened 2017-08-29
ELPP-3120 Lax load balancing: ports and listeners

giorgiosironi:


giorgiosironi:
Will rework it to keep the feature, but change the lax configuration to be an `internal` ELB.

giorgiosironi:
Haven't found any way to expose an `internal` load balancer, so Lax's ELB will have to be `internet-facing`. Authentication of the api-gateway requests however stops working, so that's a problem to be solved.

PR #178, opened 2017-08-31
Attempt to remove remote_master

giorgiosironi:
...and consequently the builder deploy on master-server (TBD).

We try to download the template from builder-private, render it with the formula folders (the only thing that varies) and upload it again. However, the Vagrantfile would have to do the same.

An alternative would be to write this file from within update-master.sh, but parsing and modifying a YAML file from bash didn't seem very robust. You can execute a bunch of  commands but they'll soon break horribly if a space is misplaced.


lsh-0:
> but parsing and modifying a YAML file from bash didn't seem very robust

yeah ... lets not do that :| 

always in favour of removing code :+1:

giorgiosironi:
Added the same behavior in Vagrant, and removed /opt/builder (also on existing servers once they get updated). It's a larger change so new review required...

.master.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python
```
lsh-0: not happy about this snippet of logic living in it's own hidden file.

Vagrantfile
```diff
@@ -307,6 +307,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
             all_formulas = YAML.load(IO.popen("/bin/bash -c \"source venv/bin/activate && ./.project.py --formula\"").read)
             project.vm.provision("shell", path: "scripts/init-master.sh", \
                 keep_color: true, privileged: true, args: [INSTANCE_NAME, pillar_repo, all_formulas.join(' ')])
+            master_configuration = IO.popen("/bin/bash -c \"source venv/bin/activate && PYTHONPATH=src ./.master.py etc-salt-master.template | tee etc-salt-master\"").read
```
lsh-0: and these awful little lines are accruing.

src/buildercore/bootstrap.py
```diff
@@ -429,6 +429,32 @@ def download_master_builder_key(stackname):
         operations.get(remote_path=private_key, local_path=fh, use_sudo=True)
     return fh
 
+def download_master_configuration(master_stack):
+    fh = StringIO()
+    with stack_conn(master_stack):
+        operations.get(remote_path='/etc/salt/master.template', local_path=fh, use_sudo=True)
+    return fh
```
lsh-0: this is returning a file-like object and not it's string value. is this intended? 
if not, you might want: `return fh.getvalue()` instead

https://docs.python.org/2/library/stringio.html#StringIO.StringIO.getvalue

src/buildercore/bootstrap.py
```diff
@@ -429,6 +429,32 @@ def download_master_builder_key(stackname):
...
+    formula_path = '/opt/formulas/%s/'
+
+    cfg['file_roots']['base'] = ["/opt/builder-private/salt/"] + [formula_path % basename(f) for f in formulas] + ["/opt/formulas/builder-base/"]
+    cfg['pillar_roots']['base'] = ["/opt/builder-private/pillar"]
+    # dealt with at the infrastructural level
+    cfg['interface'] = '0.0.0.0'
+
+    master_configuration = StringIO()
+    core_utils.ordered_dump(cfg, master_configuration)
+    return master_configuration
```
lsh-0: another instance of a file-like object being returned instead of a string. ignore if intentional.

lsh-0:
alright - I've put my money where my mouth is and fixed up those few things that were sticking in my craw: https://github.com/elifesciences/builder/pull/180

besides those things in my PR, I think this PR is fine :+1:

PR #179, opened 2017-09-01
feat: master.remaster_minion improvement

lsh-0:
master.remaster_minion now looks for the most recent master server stack and uses it's internal ip if one isn't explicitly provided


src/buildercore/core.py
```diff
@@ -532,25 +532,27 @@ def find_region(stackname=None):
 
 @testme
 def _find_master(stacks):
```
giorgiosironi: since this is now returning a list, it could be renamed to `_find_masters()`

PR #180, opened 2017-09-04
Removing remote master tweaks

lsh-0:
* shifted `.master.py` logic into `.project.py` 
* wrapped calls to `.project.py` in Vagrantfile to reduce duplication
* shifted the rendering of master config to YAML out of `bootstrap.py`
* renamed `render_master_configuration` to `expand_master_configuration`
* gave `expand_master_configuration` a sensible default for formula list

this PR feels a bit heavy handed so I want to explain the rationale behind the changes.
1. the snippets of logic in little hidden files accumulates. we would have three now if I hadn't refactored your other PR. they don't appear to be linted, or scrubbed or tested and they live outside of the `src` or `scripts` directory. this time around I've adjusted `.project.py` so it's more obvious how to add more ad-hoc code ('tasks') as we need them.

2. I've never liked the `ordered_load` and `ordered_dump` functions in utils.py. you can still see my disabled `yaml_loads` code waiting for it's chance to be useful one day and `ordered_dump` behaves differently if you pass a stream in as well. I've now made two wrapper functions with explicit return types. 

3. I prefer the rendering/serialization of data done at the last moment if possible. Data is malleable and testable and YAML/JSON serializations of data less so.

4. still a bit blergh about those temporary files in the project root (`etc-salt-master` and `etc-salt-master.template`), I feel builder needs a temporary working directory with a blanket ignore in `.gitignore`. could download files to this directory by default as well... another problem for another time though.

if you're ok with this PR, feel free to merge it and it's parent

### DIFF
--- a/.master.py
+++ b/.master.py
@@ -1,9 +1,0 @@
-#!/usr/bin/env python
-# renders a master-server's configuration as YAML
-# only argument is the template etc-salt-master file
-import sys
-from buildercore import bootstrap, project
-
-all_formulas = project.known_formulas()
-master_configuration_template=open(sys.argv[1], 'r')
-print bootstrap.render_master_configuration(master_configuration_template, all_formulas).getvalue()
--- a/.project.py
+++ b/.project.py
@@ -2,6 +2,7 @@
 # renders a project's final configuration as YAML
 # if run without any arguments, returns a list of known projects
 # requires the builder venv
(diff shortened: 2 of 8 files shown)


lsh-0:
> they don't appear to be linted, or scrubbed or tested 

not a criticism! they just don't appear to have a mention in the lint/scrub scripts

PR #182, opened 2017-09-06
adds a default mysql setting to the minion config file.

lsh-0:
mysql disallows remote connections by default in 16.04 for tighter security and this will allow us to do away with salt's connection_username: ... type settings for mysql.

tested on vagrant on 16.04 and 14.04 and on 14.04 with and without the `connection_*: ...` fields in the `mysql_user:` states. this should be backwards compatible.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -127,6 +127,7 @@ else
 fi
 
 echo "$minion_id" > /etc/salt/minion_id
+echo "mysql.unix_socket: '/var/run/mysqld/mysqld.sock'" > /etc/salt/minion.d/mysql-defaults.conf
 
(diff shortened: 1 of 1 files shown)


lsh-0:
related PR: https://github.com/elifesciences/elife-libraries-formula/pull/3

lsh-0:
that's a good point. I'll test it on an rds backed instance.

lsh-0:
elife-libraries doesn't use an rds instance but journal-cms does, and it is doing what we do with postgresql and have a great bit if-else block adding in connection params. I'm still uncertain if that unix_socket param conflicts with explicitly specifying host and port params, so I'll launch an rds backed instance of journal-cms and see

lsh-0:
journal-cms builds on rds-backed 1604 100%

PR #184, opened 2017-09-07
VPC ip range needlessly strict

giorgiosironi:
The real range is 10.0.0.0/16, this is only one subnet (in us-east-1d). Created a problem for balancing lax as the ELBs are in two different subnets (respectively in two different availability zones).
Not urgent nor quick to perform, but I want to keep track of it so PR.


giorgiosironi:
Applying to all 12 remaining stacks...

PR #185, opened 2017-09-07
Reduce blue-green deployment timeouts

giorgiosironi:
Waiting on instances to get notices on the ELB after they have been added or removed. 10 minutes is ridiculously long and in case one of those health checks fail we want to know sooner.


giorgiosironi:
Aand the tests hit the timeout, that's why they were so high. How come they are so slow, 60 seconds to detach/attach an instance from an ELB? Will investigate.

PR #187, opened 2017-09-08
donotmerge - optimised iiif project config

lsh-0:
it seems like the normal case for iiif is to be clustered and behind an elb with an external volume.

* iiif, shifts the common attributes in alt configs into default config
* adds a simplified 'adhoc' alt config with no cdn, elb, ext vol, etc
* removed gutted alt-configs.

this PR is just for feedback, I haven't tested the changes. I have a sneaking suspicion the project merging will detect the presence of the empty (but not absent) keys and merge in the project defaults.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -947,82 +947,41 @@ iiif:
         type: t2.medium
         ports:
             - 22
-            - 80
-            - 443
+            - 80:
+                cidr-ip: 10.0.0.0/16 # ELB only
+        ec2:
+            cluster-size: 2
+        elb:
+            protocol: 
+                - https
+            healthcheck:
+                path: /
+        cloudfront:
+            subdomains: 
+                - "{instance}--cdn-iiif"
+            headers:
+                - Host
+        ext:
+            size: 30 # GB
     aws-alt:
-        # for testing new versions
+        adhoc:
+            description: a single ec2 iiif server available over HTTPS. no ELB, no CDN.
+            ports:
+                - 22
+                - 80:
+                    cidr-ip: 10.0.0.0/16 # ELB only
+                - 443
+            ec2:
+                cluster-size: 1
+            elb: []
+            cloudfront: []
+            ext: []
         ci:
-            ports:
-                - 22
-                - 80:
-                    cidr-ip: 10.0.0.0/16 # ELB only
-            ec2:
-                cluster-size: 2
-            ext:
-                size: 30 # GB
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -947,82 +947,41 @@ iiif:
...
+                - https
+            healthcheck:
+                path: /
+        cloudfront:
+            subdomains: 
+                - "{instance}--cdn-iiif"
+            headers:
+                - Host
+        ext:
+            size: 30 # GB
```
giorgiosironi: can be tested with `./bldr update_template:iiif--end2end`. It will show the changes to the CloudFormation template (there shouldn't be any), asking for confirmation (press Ctrl-C)

projects/elife.yaml
```diff
@@ -947,82 +947,41 @@ iiif:
         type: t2.medium
         ports:
             - 22
-            - 80
-            - 443
+            - 80:
```
giorgiosironi: indeed we only need 22 public; 80, internal traffic

projects/elife.yaml
```diff
@@ -947,82 +947,41 @@ iiif:
...
+                cidr-ip: 10.0.0.0/16 # ELB only
+        ec2:
+            cluster-size: 2
+        elb:
+            protocol: 
+                - https
+            healthcheck:
+                path: /
+        cloudfront:
```
giorgiosironi: not sure `cloudfront` should be needed as a default: `ci` for example doesn't need it

projects/elife.yaml
```diff
@@ -947,82 +947,41 @@ iiif:
...
-                    - iiif
-                headers:
-                    - Host
+                cluster-size: 1
+            elb: []
+            cloudfront: []
+            ext: []
+        ci:
+            description: for testing new versions. no CDN.
+            cloudfront: []
```
giorgiosironi: check that this zeroes out cloudfront on this environment

lsh-0: indeed. if it doesn't zero it out, I wonder what the convention should be for 'ignore this key' - in the bot-lax-adaptor scraper I have a special value bound to `EXCLUDE_ME` that works well. I think an empty list/map may not be obvious enough

projects/elife.yaml
lsh-0: the intent is to show the normal case for this project, not the minimal. the CI configuration is the odd one out looking at all the duplication.

PR #189, opened 2017-09-13
Balance lax--end2end

giorgiosironi:
- lax--balanced is valid again
- IntDNS is removable so projects can drop their elife.internal hostname

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -168,6 +168,7 @@ master-server:
 
 lax:
     subdomain: lax # lax.elifesciences.org
+    intdomain: False
     repo: https://github.com/elifesciences/lax.git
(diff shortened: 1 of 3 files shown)


giorgiosironi:
Manually changing `end2end--lax.elifesciences.org` DNS to point to the new ELB and running https://ci--alfred.elifesciences.org/job/test-lax/ again.

giorgiosironi:
Alfred even automatically downloaded the new context file, but the ELB isn't registering the instance.

giorgiosironi:
Patching at https://github.com/elifesciences/lax-formula/commit/512a9002a491f4ee053e592954ba0f054c851997 but could use being more strict in the allowed host names.

PR #193, opened 2017-09-15
fix - removed condition around replacing minion conf file

lsh-0:
* `/etc/salt/minion` is now updated unconditionally, overwriting any default version that may be installed
* replaced references to 'service' in scripts/* to a 'systemctl' + 'service' version.
* removed 'kick-master.sh' script. unused, probably for a long time.
* concerned about a .fragment file duplicating prep-stack.sh


src/buildercore/trop.py
```diff
@@ -180,7 +180,7 @@ def ec2instance(context, node):
     else:
         subnet_id = lu('project.aws.redundant-subnet-id')
 
-    clean_server = _read_script('.clean-server.sh.fragment')
+    clean_server = _read_script('.clean-server.sh.fragment') # this file duplicates scripts/prep-stack.sh
```
giorgiosironi: I have not found a good way of removing this duplication, short of "compiling" bash scripts from a common template. The fragment is inserted into the `UserData` of new EC2 instances so it doesn't have the `#!/bin/bash...set -e...` stuff

lsh-0: yeah, me neither, otherwise I would have attempted a fix :|

it's more of a reminder-to-self that there is a design smell there.

giorgiosironi: The full `prep-stack.sh` is only used in AMI creation now, so there may be a way to rename/simplify what it does so that they are not duplicates anymore

giorgiosironi:
Gets `elife-xpub--ci` unstuck. Don't know why I have not seen the problem before, perhaps I created it erroneously with 14.04.

lsh-0:
> originally created it erroneously with 14.04.

quite possibly. I've done this at least once (cough) many times.

PR #194, opened 2017-09-15
Adding Postgresql RDS to profiles

giorgiosironi:
For the production and production-like environments.

Execute with `./bldr update_template:profiles--end2end`.

### DIFF
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -40,7 +40,10 @@ def build_context(pname, **more_context): # pylint: disable=too-many-locals
         del more_context['existing_context']
 
     def from_existing_context(field, default_value):
-        return existing_context.get(field, default_value)
+        password = existing_context.get(field)
+        if password:
+            return password
+        return default_value
 
     supported_projects = project.project_list()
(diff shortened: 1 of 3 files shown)


src/buildercore/cfngen.py
```diff
@@ -40,7 +40,10 @@ def build_context(pname, **more_context): # pylint: disable=too-many-locals
         del more_context['existing_context']
 
     def from_existing_context(field, default_value):
-        return existing_context.get(field, default_value)
+        password = existing_context.get(field)
+        if password:
+            return password
+        return default_value
```
thewilkybarkid: Is this because of empty strings/`None` values?

thewilkybarkid: Just seen commit.

giorgiosironi: After https://github.com/elifesciences/builder/pull/191 we stopped regenerating the RDS super user password at every update, reusing the previously existing one if it's there. However, the "previously existing one" is `None` if there was no RDS in that stack; in that case, we need to actually generate it.

PR #195, opened 2017-09-19
Feature: vagrant 16.04

lsh-0:


projects/elife.yaml
```diff
@@ -131,10 +131,24 @@ defaults:
...
+            description: isolated from the master-server and uses Ubuntu 16.04 (Xenial)
+            type: t2.small
+            ec2:
+                ami: ami-1d4e7a66 # xenial, build 20170811, hvm:ebs-ssd
+                masterless: true
     vagrant:
         # why 'bento'? https://bugs.launchpad.net/cloud-images/+bug/1569237
-        # box: bento/ubuntu-16.04 # Ubuntu 16.04 "Xenial"
-        box: ubuntu/trusty64 # Ubuntu 14.04
+        box: bento/ubuntu-16.04 # Ubuntu 16.04 "Xenial"
```
giorgiosironi: 16.04 as a default for all Vagrant VMs? Given rollout is going to be project-by-project, it should go at the project level until everything is ported
(or single projects could override this back with `ubuntu/trusty64` until they are porteD)

projects/elife.yaml
```diff
@@ -143,13 +157,32 @@ defaults:
...
+    formula-repo: https://github.com/elifesciences/basebox-formula
+    aws:
+        ec2:
+            ami: ami-1d4e7a66 # xenial, us-east-1, build date 20170811, hvm:ebs-ssd
+            masterless: true
+        rds:
+            storage: 5
+            engine: MySQL
+            version: 5.7
+
```
giorgiosironi: if you want to go the extra mile, this could have a small `ext` so that we test `elife.external-volume` in builder-base-formula thoroughly

scripts/bootstrap.sh
```diff
@@ -123,6 +123,7 @@ if $upgrading; then echo "$(date -I) -- upgraded to $version" >> /root/events.lo
 # put minion id in dedicated file else salt keeps recreating file
 printf "master: %s\nlog_level: info\n" "$master_ipaddr" > /etc/salt/minion
 echo "$minion_id" > /etc/salt/minion_id
+echo "mysql.unix_socket: '/var/run/mysqld/mysqld.sock'" > /etc/salt/minion.d/mysql-defaults.conf
```
giorgiosironi: Was in https://github.com/elifesciences/builder/pull/182 so that could be merged separately or closed in favor of this

src/buildercore/bootstrap.py
```diff
@@ -517,7 +517,7 @@ def _update_ec2_node():
 
         if is_masterless:
             # order is important.
-            formula_list = '"%s"' % ' '.join(pdata.get('formula-dependencies', []) + [pdata['formula-repo']])
+            formula_list = '%s' % ' '.join(pdata.get('formula-dependencies', []) + [pdata['formula-repo']])
```
giorgiosironi: ah, ran into this, fix is correct then

giorgiosironi:
Have some small changes in https://github.com/elifesciences/builder/pull/197/files that are identical as here and should not cause conflicts.

projects/elife.yaml
lsh-0: I like the second option best. I introduced it because I was accidentally bringing up 14.04 instances because I missed a step updating the project file. as this introduces 16.04 support for all projects, I think I'll go through and add explicit 14.04 AMIs and boxes that we'll remove as we port each one

lsh-0: done.

scripts/bootstrap.sh
lsh-0: I merged it in.

src/buildercore/bootstrap.py
lsh-0: yeah, I'm sorry about that. not sure how it snuck in.

PR #203, opened 2017-10-03
Building custom example.top files

giorgiosironi:
At times it's useful to pass in a different example.top, as in the case of a building a VM more close to end2end or prod than the default one.

For example, in `journal-formula` we could have a separate `example-critical-css.top` that enables additional state files.

### DIFF
--- a/scripts/init-formulas.sh
+++ b/scripts/init-formulas.sh
@@ -105,7 +105,7 @@ clone_update() {
     # tell the minion where it can find the formula's state and pillar data
     echo "    - $repo_path/salt/" >> /etc/salt/minion.d/file_roots.conf
 
-    topfile="$repo_path/salt/example.top"
+    topfile="$repo_path/salt/${BUILDER_TOPFILE-example.top}"
 
     # successively overwrites the top file until the last one (project formula) wins
     if [ -e "$topfile" ]; then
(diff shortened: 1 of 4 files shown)


giorgiosironi:
Tested the Vagrant one, but would like some review before changing journal-formula to allow this.

lsh-0:
it's a good idea. any reason why it's being passed as an envvar rather than a script parameter? will take a closer look today

scripts/init-formulas.sh
```diff
@@ -105,7 +105,7 @@ clone_update() {
     # tell the minion where it can find the formula's state and pillar data
     echo "    - $repo_path/salt/" >> /etc/salt/minion.d/file_roots.conf
 
-    topfile="$repo_path/salt/example.top"
+    topfile="$repo_path/salt/${BUILDER_TOPFILE-example.top}"
```
lsh-0: I hadn't seen this syntax before.

giorgiosironi: it's the best way I found to specify defaults for a variable

giorgiosironi:
I started with an env variable because it's easy to pass to `vagrant up` (`PROJECT` for example is passed in this way). Then it becomes easy (perhaps too easy) to propagate to the different layers, because it's essentially global. If it could be made into an explicit parameter it would be more clear, I agree and I'll give it a try.

lsh-0:
>  I'll give it a try.
it's fine for now, really.

PR #204, opened 2017-10-05
Add schematron-validator to the elife project config

garyttierney:


projects/elife.yaml
```diff
@@ -985,6 +986,25 @@ personalised-covers:
         ports:
             1246: 80
             1247: 8080
+
+schematron-validator:
+    domain: false
```
giorgiosironi: it has no domain because it should only be accessible internally? `intdomain` is still there so this will acquire a `ci--schematron-validator.elife.internal` hostname only

projects/elife.yaml
```diff
@@ -985,6 +986,25 @@ personalised-covers:
...
+    subdomain: schematron-validator
+    default-branch: master
+    repo: https://github.com/elifesciences/schematron-validator
+    formula-repo: https://github.com/elifesciences/schematron-validator-formula
+    formula-dependencies:
```
giorgiosironi: I think this is inherited from the defaults, so doesn't need to be specified

projects/elife.yaml
```diff
@@ -985,6 +986,25 @@ personalised-covers:
...
+        - https://github.com/elifesciences/builder-base-formula
+    aws:
+        type: t2.micro
```
giorgiosironi: this is only 1GB of RAM, is that enough to run the whole software stack?

projects/elife.yaml
```diff
@@ -985,6 +986,25 @@ personalised-covers:
...
+        ports:
+            - 22
+            - 80
```
giorgiosironi: makes sense to have only 80 if it's internal

giorgiosironi:
@lsh-0 due to the good default, this new project has been created as 16.04 rather than 14.04 :champagne:

lsh-0:
hooray!

PR #207, opened 2017-10-26
Introducing RDS start and stop

giorgiosironi:
Rough implementation of RDS start and stop primitives, including the polling to ensure they go into the specified states. RDS is slower to react, but it's somewhat simpler as there is no SSH polling that we can do to check if the instances are up; and we cannot read from the API the last starting time of the node.

Since RDS transitions are much slower than EC2, there is some more customization on involved services to support use cases:
- start everything (why would you start only a piece? A project would never work with just the database or just the EC2 up)
- stop only EC2 if they have been running for half an hour (frequently)
- stop only RDS or stop everything (at the end of the week)
Some obsolete scenario like `stop_if_next_hour_is_imminent` has been removed as AWS now bills in such a way that it makes no sense to stop instances in that pattern.

Unit testing of the `lifecycle` module is extended to cover basic start/stop scenarios without resorting to creating real EC2 or RDS resources (`integration_tests`), which would make the test suite very slow and brittle. Also the S3 testing needed to support concurrent runs to avoid builds clashing with a developer running the tests.

Will require some work to clean and productionize, but it's functionally complete.

### DIFF
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -125,17 +125,17 @@ def boto_s3_conn(region):
     return boto3.client('s3', region)
 
 @cached
-def connect_aws_with_pname(pname, service, boto3=False):
+def connect_aws_with_pname(pname, service, with_boto3=False):
     "convenience"
     pdata = project.project_data(pname)
     region = pdata['aws']['region']
     LOG.debug('connecting to a %s instance in region %s', pname, region)
-    if boto3:
+    if with_boto3:
         return boto3.client('rds', region)
     else:
         return connect_aws(service, region)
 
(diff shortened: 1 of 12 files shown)


src/buildercore/core.py
```diff
@@ -125,17 +125,20 @@ def boto_s3_conn(region):
     return boto3.client('s3', region)
 
 @cached
-def connect_aws_with_pname(pname, service):
+def connect_aws_with_pname(pname, service, with_boto3=False):
```
lsh-0: https://elifesciences.atlassian.net/browse/ELPP-3239

PR #208, opened 2017-10-27
ELPP-2839 Lowering backup retention on testing environments

giorgiosironi:
It's not going to make a lot of difference (in total we spend about $50/month and snapshots are incremental so having many may not occupy too much space wrt to a few). But it's really a free lunch, just configuration.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -492,7 +492,10 @@ elife-metrics:
         end2end:
             description: end2end environment for metrics. RDS backed
             rds:
-                storage: 5
+                # currently:
+                storage: 10
+                # it should normally be
+                # storage: 5
                 backup-retention: 2 # days
         prod:
             description: production environment for metrics. RDS backed
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -182,10 +182,6 @@ basebox:
             ami: ami-9eaa1cf6 # Ubuntu 14.04 (correct, but older)
         ports:
             - 22
-    aws-alt:
-        rdstest:
-            rds:
-                storage: 5
     vagrant:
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Perhaps there is a way to do this in a `defaults.aws-alt.rds` and inherit it into all `end2end` environments automatically, perhaps not.

lsh-0:
> Perhaps there is a way to do this in a defaults.aws-alt.rds and inherit it into all end2end environments automatically, perhaps not.

um - it would be nice to have:

```
defaults:
    ...
    aws-alt:
        end2end:
            rds:
                ...
                backup-retention: 2 # days
```

right? we don't have a global `end2end` (or `ci` or `continuumtest`) alt configs, so those that never had one would gain one. I'll take a quick look today

thewilkybarkid:
Do we actually need backups of these environments? Is it just to speed up recreating from scratch, or is there something there that needs persisting?

giorgiosironi:
Usually nothing that needs persistence, we plan to do a restore from prod to have a better data set but that would be done with UBR, while this is purely for disaster recovery.
In the end, the real purpose is that when there are problems with the CloudFormation templates (in this case the snapshot-configuring fields) we find out when updating `end2end` rather than bricking `prod`.

giorgiosironi:
@lsh-0 I see no harm in getting a default `end2end` environment in a project, if it's not used (which costs money and time instead).

giorgiosironi:
All applied but `elife-metrics`, see ticket.

giorgiosironi:
Applied also to `elife-metrics--end2end`, but by changing the storage size to reflect the real one.

PR #209, opened 2017-10-30
[WIP] Looking for a way to pin dependencies

giorgiosironi:
This is the result of running `pip freeze > requirements.txt`. It pins dependencies, but has the problem that direct dependencies are lost.
How can we at the same time pin all dependencies but keep track of the direct ones? Perhaps a separate `requirements-direct.txt` file? This is sort of what PHP and Composer are doing, with `composer.json` and `composer.lock`; NPM too with `package.json` and `npm-shrinkwrap.json`; Ruby with `Gemfile` and `Gemfile.lock`.


giorgiosironi:
Need to keep track of https://github.com/pypa/pipfile @seanwiseman @lsh-0

seanwiseman:
I have actually been using [pipfile](https://github.com/pypa/pipfile) along with [pipenv](https://docs.pipenv.org/) on small person projects and have to say its a far nicer work flow with far superior functionality. Though yes, the stability is not there yet.

lsh-0:
I'm really unimpressed and worried about PIpfile. I don't like how close to code it looks, which means more logic spread and a custom dsl. 

@seanwiseman , can you elaborate on your workflow?

seanwiseman:
I think `pip` + `requirements.txt` falls flat in many areas, and `pip` knows that which is why new dependency management directions are emerging. I agree some of them are poor, but the workflow I was referring to was using `Pipfile` + `pipenv`.

I could elaborate but it is not going to be as descriptive and informative as the documentation (and video) [here](https://docs.pipenv.org/). In a nutshell, it takes elements that work well from other languages dependancy management solutions and brings them the Python. 

`pipenv` + `Pipfile` clears up splitting dev and prod dependancies, security and update checks, locking dependencies nicely, `virtualenv` management, generating dependency graphs, dependency hash management, and other bits. But on top of all that it's just far nicer to use, just try it out to really get a feel for it if you have a few minutes.

`TOML` and `Pipfile` need to stabilize before you would consider it in production. Also I will be interesting to see how `pip` decides to fully support `Pipfile` in the future.

lsh-0:
thank you for the link.

giorgiosironi:
Closing in favor of `pipenv` experimentation.

PR #210, opened 2017-10-30
ELPP-3233 Consistent tagging

giorgiosironi:
Useful to have the same tags all over, so that they can be used without confusion from the AWS billing console.

- Tag `Cluster` for RDS and ElastiCache
- Tags for EBS volumes
- Tags for S3 buckets
- Tags for ELB already in place
- Dropping `Owner` tag (like for RDS before) because it leads to updates being performed just when two different people work on the same stack

### DIFF
--- a/src/buildercore/trop.py
+++ b/src/buildercore/trop.py
@@ -117,14 +117,15 @@ def rds_security(context):
 #
 
 
-def _generic_tags(context):
-    return {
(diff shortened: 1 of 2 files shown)


lsh-0:
I was wondering just who the eff BlueReZZ was and what he was doing approving PRs on builder until I clicked the profile ;)

PR #211, opened 2017-10-31
Add CloudFront-Viewer-Country header to Journal requests

thewilkybarkid:


thewilkybarkid:
Confirmed working on continuumtest, will apply to end2end and prod now.

Was a little confusing as Journal on continuumtest doesn't know to trust CloudFront, so the WebP feature flag didn't actually work.

giorgiosironi:
What do we need to do to add the missing configuration to `journal--continuumtest`? It we keep a CDN around, it should be able to rely on it.

thewilkybarkid:
Something like https://github.com/elifesciences/journal/blob/develop/web/app_prod.php#L12-L18.

PR #212, opened 2017-11-01
bad ec2 buildvar breaks ssh access to instance

lsh-0:
affects elife-metric--ci. build vars here: [elife-metrics--ci.buildvars.json.gz](https://github.com/elifesciences/builder/files/1432684/elife-metrics--ci.buildvars.json.gz)

not sure how it's buildvars got into that state but it's preventing Alfred from doing it's job


lsh-0:
I think you're right, I vaguely recall that as well.

PR #213, opened 2017-11-01
list of ec2 stacks from core.aws_stacks was truncated to 100

lsh-0:
suspecting pagination, I took the opportunity to upgrade this small bit to boto3 and use it's paginators and saw that there are now 102 ec2.

### DIFF
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -99,7 +99,7 @@ def connect_aws(service, region):
     return conn.connect_to_region(region)
 
 @cached
-def boto_cfn_conn(region):
+def _boto_cfn_conn(region):
     return boto3.client('cloudformation', region_name=region)
 
 @cached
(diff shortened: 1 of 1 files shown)


lsh-0:
I don't really care for leading underscores or the idea that some things should be private/hidden and others public, but I care even less about changing it now. Python enforces double underscores inside objects but nothing else as far as I'm aware.

Although I was raised a Good Java Developer, member visibility, interfaces, objects - they all seem to be for crafting pretty code and then, in the name of crafting pretty code, it becomes verbose. Then, in the name of reducing verbose code, the code is spread across many modules to make it more 'readable'. Then the same is applied to all code in the codebase to make it 'consistent', then all code across all projects in an organisation. These programmer affectations really irk me and seem to originate from one's inability to just pause and read and understand a program.

Prefixing a connection with an underscore isn't going to stop me using it from outside `buildercore.core`. If one day Python complains it's private, I'll make it public. 

This rant isn't directed at you, @giorgiosironi  :(

giorgiosironi:
It's from PEP 8: https://www.python.org/dev/peps/pep-0008/
```
internal interfaces (packages, modules, classes, functions, attributes or other names) should still be prefixed with a single leading underscore.
```
I think that the Python ecosystem agrees with you in spirit, as this is only a convention rather than a language-enforced visibility check. It allows people to express that some stuff could be renamed/removed in the future, on top of a public default.

lsh-0:
I know the pep, I'm effectively snapping at you for following 'the rules'. storm in a teacup type stuff. nobody ever got fired for buying IBM.

PR #214, opened 2017-11-01
Default standalone is 16.04

giorgiosironi:
Projects have to override the default to use 14.04, so they have to do it for their standalone instance too.


giorgiosironi:
Have to go with this to fix https://github.com/elifesciences/profiles-formula/pull/18#pullrequestreview-73352418, but check I got it right.

PR #217, opened 2017-11-06
CloudFront logs

giorgiosironi:
CloudFront can fill a bucket with lots of zip files that you have to aggregate to see the traffic coming to the CDN. This can be useful for statistics, but in general to identify real user agents and real ip addresses in case you want to know the source of some traffic.

Spawned from a search incident this afternoon in which we know there is someone crawling our search pages, but we can't know where the traffic comes from.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -118,6 +118,7 @@ defaults:
             headers: []
             errors: null
             default-ttl: 0 # seconds
+            logging: false
         subdomains: []
         elasticache:
             # elasticache defaults only used if an `rds` section present in project
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -407,6 +407,8 @@ journal:
                         # ELB with no active instances
                         503: "/5xx.html"
                     protocol: http
+                logging:
+                    bucket: elife-cloudfront-logs
         continuumtest:
             elasticache:
                 engine: redis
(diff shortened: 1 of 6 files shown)


giorgiosironi:
I have done this through the console on `journal--prod`, so I am doing it in `end2end` too with this code, as a test. If everything goes well tomorrow I can disable it from the console and properly enable it from here.

lsh-0:
nothing wrong with the PR. 

I'm not sure what happened regarding Paul Kelly's data privacy push and meeting new EU regulations meeting, but even after we've identified this crawler, the best outcome we can hope for is we politely ask them to stop and they do.

A better solution and one that doesn't involve collecting more user information would be temporarily throttling traffic in KONG.

I read some of the suggestions on Slack, like returning chunks of results rather than piecemeal and they seem fine as well.

giorgiosironi:
In general we should be able to support this kind of load; but it's the second time we turn on the logs, the first time was identifying Yandex causing 404s on the IIIF CDN. So we should do this turn on/turn off through builder, rather than fiddling with the console and creating snowflakes.

giorgiosironi:
I agree that these logs are a liability, so we need to delete them, either manually after an investigation like this or periodically. I'm still unclear on the GDPR requirements so I'll investigate that separately.

giorgiosironi:
About the Kong throttling: these are all requests coming from Journal generating search result pages , so Kong sees them as internal, coming from one of the 3 `journal--prod` servers.

lsh-0:
yeah, I was worried about that. I thought there might have been a 'forwarded-for' type header in the request from journal to kong that could be used to target traffic streams.

PR #220, opened 2017-11-20
ELPP-3271 Add prod-elife-bot-sessions bucket and similar

giorgiosironi:


giorgiosironi:
The configuration creates the buckets through an underlying CloudFormation template; CloudFormation does not have the capability to pick up existing resources, so all the buckets that were existing before builder can't be put here or it would try to create them and fail (because they already exist).

We could, in theory, port the existing buckets here by creating them with new names and copying the content. It is a massive migration with low benefit however.

PR #221, opened 2017-11-27
ELPP-3301 Check that all instances are in service before disconnecting some from ELB

giorgiosironi:
In case a deploy is attempted while one of the instance is not healthy:
```
    ...
    raise SomeOutOfServiceInstances(service_status_by_id)
buildercore.bluegreen.SomeOutOfServiceInstances: {'i-0a902e0941b0f6c00': 'InService', 'i-09ec22c3eaf95a51a': 'OutOfService'}
```

### DIFF
--- a/src/buildercore/bluegreen.py
+++ b/src/buildercore/bluegreen.py
@@ -16,6 +16,7 @@ class BlueGreenConcurrency(object):
 
     def __call__(self, single_node_work, nodes_params):
         elb_name = self.find_load_balancer(nodes_params['stackname'])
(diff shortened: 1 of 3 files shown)


giorgiosironi:
I wonder how this test had been working until now, it appears without Salt running on the created nodes they would always be `OutOfService` for the duration of the test.

giorgiosironi:
`journal--prod--2` remained unhealthy after a failed daily highstate (Penthouse CSS timeout), so on the next deploy 1 and 3 were briefly detached too, leaving the ELB with no instances attached.

PR #224, opened 2017-12-05
Feature: masterless instances with formulas configured at init

lsh-0:
launch command looks like:

`./bldr masterless.launch:project,instanceid,alt-config,formula1-name@<rev-or-branch>,formula2-name@<rev-or-branch>`

for example, this works:

`./bldr masterless.launch:lax,bronin3,standalone,lax-formula@9e21f1b4e3e03fdd75984a665a3c4e2c29240d00,builder-base-formula@85ed66ae0be268aed24631691a0ad56d23fded40`

### DIFF
--- a/src/deploy.py
+++ b/src/deploy.py
@@ -1,47 +1,14 @@
 """module concerns itself with tasks involving branch deployments of projects."""
 
 from fabric.api import task
-from decorators import requires_branch_deployable_project, echo_output, setdefault, requires_aws_stack, timeit
-import utils
-from buildercore import core, bootstrap, cfngen, project
+from decorators import requires_aws_stack
+from buildercore import bootstrap
 from buildercore.concurrency import concurrency_for
-import cfn
 import buildvars
 
 import logging
 
 LOG = logging.getLogger(__name__)
-
-def impose_ordering(branch_list):
-    branch_list, removed = utils.rmval(branch_list, 'master', 'develop')
-    branch_list.sort()
-    branch_list = removed + branch_list
-    return branch_list
-
-#
-#
-#
-
-@task
-@requires_branch_deployable_project
-@echo_output
-@timeit
-def deploy(pname, instance_id=None, alt_config=None):
-    pdata = project.project_data(pname)
-    stackname = cfn.generate_stack_from_input(pname, instance_id, alt_config=alt_config)
-
-    region = pdata['aws']['region']
-    active_stacks = core.active_stack_names(region)
-    if stackname in active_stacks:
-        LOG.info("stack %r exists, skipping creation", stackname)
-    else:
-        LOG.info("stack %r doesn't exist, creating", stackname)
-        more_context = cfngen.choose_config(stackname)
-        cfngen.generate_stack(pname, **more_context)
-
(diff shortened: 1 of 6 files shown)


src/deploy.py
```diff
@@ -1,48 +1,15 @@
...
-
-#
-#
-#
-
-@task
-@requires_branch_deployable_project
-@echo_output
-@timeit
-def deploy(pname, instance_id=None, alt_config=None):
```
giorgiosironi: it seems this has been completely absorbed by `launch` now

src/masterless.py
```diff
@@ -80,11 +60,62 @@ def set_versions(stackname, *repolist):
...
+    import cfn
+    cfn.launch(pname, instance_id, alt_config, formula_revisions=repolist)
+
+
+@task
+@requires_aws_stack
+@requires_master_server_access
+@requires_masterless
+def update(stackname):
+    import cfn
```
giorgiosironi: can it be imported at the top level in this file? Some clashing names?

giorgiosironi:
Trying it with:
```
./bldr masterless.launch:lax,bronin4,standalone,lax-formula@f1784de85d1f30f8995e3911b8a31e309ccd
263d,builder-base-formula@d65af97bc41c893f5a589e5c1f440510796aaa97
```

src/masterless.py
```diff
@@ -80,11 +60,62 @@ def set_versions(stackname, *repolist):
...
+    return cfn.update(stackname)
```
giorgiosironi: do we need a separate `masterless.update` task, given it's just delegating? Or the annotations are doing very smart things?

giorgiosironi:
Running it seems to work well.

src/masterless.py
lsh-0: I kind of trusted my previous self to have made the right decision, but I don't recall a circular dependency ever occurring in masterless.py. I generally don't like importing from other task modules because there is no clear hierarchy, but that itself is just avoiding a smell

src/masterless.py
lsh-0: no, I don't think we need it. It was an early effort to provide a similar interface to cfn.py, but now it's not doing much besides being noisy.

src/masterless.py
lsh-0: ah, it wasn't a circular dependency, it's that if you define a task in one module and then import that module into another, then that other module now has the tasks `secondmodule.firstmodule` task in it:

```
    masterless.cfn.aws_delete_stack      tell aws to...
    masterless.cfn.aws_launch_instance
    masterless.cfn.aws_stack_list        returns a l...
    masterless.cfn.aws_update_stack      Updates the...
    masterless.cfn.cmd
```

src/deploy.py
lsh-0: my thoughts too, and they were both using `bootstrap.create_update` and `create_update` was only ever a convenience function, so I absorbed that into `cfn.launch`.

`deploy.switch_revision_update_instance` should probably be moved to `cfn.py`, but that can happen later. I feel there is a refactor/clean up in builder's immediate future with the move to python 3

PR #226, opened 2017-12-11
crm project description now defaults to being rds-less

lsh-0:
this formula can now restore a tiny version of the database to use


giorgiosironi:
Perhaps `ci` doesn't need it. But it's the only place where we can test stuff so makes sense to be as close to `prod` as possible.

PR #227, opened 2017-12-13
[wip] adds rds deletion policy.

lsh-0:
I made the behaviour of 'buildercore.utils.ensure' sane again. The optional third parameter is now consistent with other projects also using ensure.
Tidied up 'buildercore.cfngen.build_context' a bit. It's still fat but we're getting there.

depends on:
* https://github.com/elifesciences/builder-base-formula/pull/99

### DIFF
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -35,21 +35,11 @@ LOG = logging.getLogger(__name__)
 def build_context(pname, **more_context): # pylint: disable=too-many-locals
     """wrangles parameters into a dictionary (context) that can be given to
     whatever renders the final template"""
-    existing_context = more_context.get('existing_context', {})
-    if 'existing_context' in more_context:
-        del more_context['existing_context']
-
-    def from_existing_context(field, default_value):
-        password = existing_context.get(field)
-        if password:
-            return password
-        return default_value
 
     supported_projects = project.project_list()
-    assert pname in supported_projects, "Unknown project %r" % pname
+    ensure(pname in supported_projects, "Unknown project %r" % pname)
 
     project_data = project.project_data(pname)
-
     if 'alt-config' in more_context:
         project_data = project.set_project_alt(project_data, 'aws', more_context['alt-config'])
 
@@ -58,7 +48,7 @@ def build_context(pname, **more_context): # pylint: disable=too-many-locals
         'project': project_data,
 
         'author': os.environ.get("LOGNAME") or os.getlogin(),
-        'date_rendered': utils.ymd(),
+        'date_rendered': utils.ymd(), # TODO: if this value is used at all, more precision might be nice
 
         # a stackname looks like: <pname>--<instance_id>[--<cluster-id>]
         'stackname': None, # must be provided by whatever is calling this
@@ -67,10 +57,13 @@ def build_context(pname, **more_context): # pylint: disable=too-many-locals
 
         'branch': project_data.get('default-branch'),
         'revision': None, # may be used in future to checkout a specific revision of project
+
+        # TODO: shift these rds_ values under an 'rds' key
         'rds_dbname': None, # generated from the instance_id when present
         'rds_username': None, # could possibly live in the project data, but really no need.
         'rds_password': None,
         'rds_instance_id': None,
+
         'ec2': False,
         's3': {},
         'elb': False,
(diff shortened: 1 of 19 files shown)


src/buildercore/cfngen.py
```diff
@@ -81,53 +74,37 @@ def from_existing_context(field, default_value):
...
-    bit_keys = ['project_name', 'instance_id', 'cluster_id']
-    bits = dict(zip(bit_keys, core.parse_stackname(stackname, all_bits=True)))
-    assert bits['project_name'] == pname, \
-        "the project name derived from the `stackname` doesn't match the given project name"
+    bits = core.parse_stackname(stackname, all_bits=True, idx=True)
+    ensure(bits['project_name'] == pname,
+           "the project name %r derived from the given `stackname` %r doesn't match" % (bits['project_name'], pname))
     context.update(bits)
 
+    # is this a production instance? if yes, then we'll do things like tweak the dns records ...
```
giorgiosironi: do we? The only usage I could find is https://github.com/elifesciences/builder-base-formula/blob/29588e1f92b461be39eab77fbf7e6b7241d8a856/elife/postfix.sls#L8 which we could drop in favor of `pillar.elife.env == 'prod'`. Reducing the size of the context can make it easier to observe and debug

src/buildercore/cfngen.py
```diff
@@ -171,6 +145,43 @@ def _parameterize(string):
     return context
 
 
+def build_context_rds(context, existing_context):
```
giorgiosironi: seems a good way to express only RDS has a dependency on `existing_context`

src/buildercore/cfngen.py
```diff
@@ -171,6 +145,43 @@ def _parameterize(string):
...
+    if 'rds' not in context['project']['aws']:
+        return {}
+    stackname = context['stackname']
+
+    # deletion policy
+    # enforce here rather than allow to be specified in project config
+    deletion_policy = 'Delete'
+    if context['is_prod_instance'] or context['instance_id'] in ['continuumtest', 'end2end']:
```
giorgiosironi: tricky to have these keys specified in keys rather than configuration :thinking:

giorgiosironi: I would prefer to go for safety and here have a default of `Snapshot`; in https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L165-L170 add a `rds` key alongside `ec2` with the override to `Delete`. It lets us treat adhoc environments (I just created `journal--annotationstest` that will be used for a month or so) as akin to production, while formula builds using `standalone` really are throwaway stacks.

giorgiosironi:
I think `test_buildercore_trop` is the place to be, it should already have some RDS sample stacks.

src/buildercore/cfngen.py
lsh-0: that's what I thought

src/buildercore/cfngen.py
lsh-0: tweaking the logic so default is Snapshot is easy.

I originally had the deletion policy specified in the projects.yaml file and found myself repeating the line `deletion-policy: Delete` again and again and again. Repeating something that often is bound to cause a mistake. I prefer enforcing this in code, but there is room for an explicit override that need not be repeated. 

But yes - those keys are just magic strings floating there. I can reinforce them conf.py or core.py, but I don't seem to recall builder knowing too much about named environments. i'll investigate

src/buildercore/cfngen.py
lsh-0: oh wow - that's some primordial Salt code :) I'll fix that formula up and remove this key.

lsh-0:
did I add you as a reviewer or did you add yourself, @giorgiosironi ? I've been conscious of the fact that I don't want to impinge on another's time until a PR is ready, and that between opening a PR and getting a review more changes may happen. I could have sworn I deliberately left off adding reviewers for this

src/buildercore/cfngen.py
lsh-0: pr: https://github.com/elifesciences/builder-base-formula/pull/99

src/buildercore/cfngen.py
lsh-0: I'm grepping through builder's source and there is a distinct lack of code doing environment checking, which I like. I wish our formula's were this free of conditional behaviour. It's going to make the project files even more verbose, but I'll go with your recommendation: default to `Snapshot` and have the alt-configs explicitly set to `Delete`

giorgiosironi:
I probably added myself because I am working in a similar area for `elasticache` and don't want to diverge

lsh-0:
no worries. I'll merge this and then check your PR for any conflicts.

PR #229, opened 2017-12-13
ELPP-3335 Elasticache multiple nodes

giorgiosironi:
Apart from the multiple clusters which is configured through a number, `overrides` is useful to customized the size of one of those (which affects how much size it has for example); `suppress` is useful for migrations as there is currently a `ElastiCache` resource in `journal` templates we'll need to remove and substitute with `ElastiCache1` and `ElastiCache2`.


giorgiosironi:
Some test suite errors due to CloudFormation throttling...

PR #230, opened 2017-12-20
ELPP-3466 Use allkeys-lru for Journal's session Redis

thewilkybarkid:
Current default is `volatile-ttl`, which would evict users who logged in a while ago if it fills up. `allkeys-lru` (or `volatile-lru`) would evict users who haven't been active for a while (which also means logging in a while ago). So someone who comes back to the site won't be logged out early.

I'm tempted to say that this should actually be our default (`volatile-lru` is ElasticCache's default). This might actually produce better results for Journal's HTTP cache too (helps with ELPP-3266), as requests like the homepage are more likely to not be evicted, whereas some random search result/obscure blog article are fine to be a bit slower.

(Note this is untested.)


projects/elife.yaml
```diff
@@ -437,6 +439,10 @@ journal:
                 engine: redis
                 type: cache.t2.micro
                 clusters: 2
+                overrides:
+                    2:
+                        configuration:
```
giorgiosironi: We only create a single `AWS::ElastiCache::ParameterGroup` for all clusters in the template at the moment, which means overrides do not work for `configuration`: https://github.com/elifesciences/builder/blob/master/src/buildercore/trop.py#L779-L785

giorgiosironi: The build detected it:
`AssertionError: `configuration` override is not allowed for single elasticache clusters`

thewilkybarkid:
`volatile-X` might be better in case we ever want to not have a TTL, but these are caches...

Happy to switch the default to `volatile-lru`.

giorgiosironi:
Added, but will need to be applied with `bldr update_template`.

thewilkybarkid:
@giorgiosironi We should apply/merge this.

giorgiosironi:
Which instances are affected? `journal` ones only I guess?
Will merge `master` after the AWSELB change and update.

giorgiosironi:
```
Update to resource type AWS::ElastiCache::ParameterGroup is not supported.
```
Interesting

giorgiosironi:
Looks like `CacheParameterGroup` can't be updated in place, but we could create a new one and change `CacheParameterGroupName` on the `CacheCluster`. That is allowed but requires `Some interruptions`.

thewilkybarkid:
If we could override `configuration` we could create a new one with the new parameter, change config then remove old one? Avoids interruptions?

giorgiosironi:
We could use the suffix `1`, `2`, etc. for the various overridden `CacheParameterGroup`; every switch would logout users I guess, but it's better than starting to return 500 or similar.

thewilkybarkid:
If we change the session cache, which we _probably_ should but is probably no-way near full?

giorgiosironi:
Forgot that we have separate servers, so this should only impact the cache. Moving to Jira to keep track of it since there are some pending changed to the builder for SQS unsubscription to avoid conflicting with.

giorgiosironi:
This is the dry-run delta on `journal--end2end`:
```
2018-03-05 09:58:37,357 - INFO - MainProcess - cfn - Create: {'Outputs': {},
 'Resources': {u'ElastiCacheParameterGroup1': {u'Properties': {u'CacheParameterGroupFamily': u'redis2.8',
                                                               u'Description': u'ElastiCache parameter group for journal--end2end cluster 1',
                                                               u'Properties': {u'maxmemory-policy': u'volatile-lru'}},
                                               u'Type': u'AWS::ElastiCache::ParameterGroup'}}}
2018-03-05 09:58:37,361 - INFO - MainProcess - cfn - Update: {'Outputs': {},
 'Resources': {u'ElastiCache1': {u'Properties': {u'CacheNodeType': u'cache.t2.small',
                                                 u'CacheParameterGroupName': {u'Ref': u'ElastiCacheParameterGroup1'},
                                                 u'CacheSubnetGroupName': {u'Ref': u'ElastiCacheSubnetGroup'},
                                                 u'Engine': u'redis',
                                                 u'EngineVersion': u'2.8.24',
                                                 u'NumCacheNodes': 1,
                                                 u'PreferredAvailabilityZone': u'us-east-1d',
                                                 u'Tags': [{u'Key': u'Cluster',
                                                            u'Value': u'journal--end2end'},
                                                           {u'Key': u'Environment',
                                                            u'Value': u'end2end'},
                                                           {u'Key': u'Name',
                                                            u'Value': u'journal--end2end'},
                                                           {u'Key': u'Project',
                                                            u'Value': u'journal'}],
                                                 u'VpcSecurityGroupIds': [{u'Ref': u'ElastiCacheSecurityGroup'}]},
                                 u'Type': u'AWS::ElastiCache::CacheCluster'}}}
2018-03-05 09:58:37,362 - INFO - MainProcess - cfn - Delete: {'Outputs': {}, 'Resources': {}}
```

giorgiosironi:
Green, ready for review @thewilkybarkid

projects/elife.yaml
```diff
@@ -404,6 +404,9 @@ journal:
                 engine: redis
                 clusters: 2
                 overrides:
+                    1:
+                        configuration:
+                            maxmemory-policy: volatile-lru
```
thewilkybarkid: We should make this the default and instead override those that have the old default.

thewilkybarkid: Where is everywhere for now?

giorgiosironi: It's easy to override everywhere, as more groups will be created; then the problem is changing the default, which is still immutable. I think I have to first add a removal of the default parameter group in case it is overridden everywhere.

giorgiosironi:
This is the diff being produced now:
```
2018-03-05 13:09:40,207 - INFO - MainProcess - cfn - Create: {'Outputs': {},
 'Resources': {u'ElastiCacheParameterGroup1': {u'Properties': {u'CacheParameterGroupFamily': u'redis2.8',
                                                               u'Description': u'ElastiCache parameter group for journal--end2end cluster 1',
                                                               u'Properties': {u'maxmemory-policy': u'volatile-ttl'}},
                                               u'Type': u'AWS::ElastiCache::ParameterGroup'},
               u'ElastiCacheParameterGroup2': {u'Properties': {u'CacheParameterGroupFamily': u'redis2.8',
                                                               u'Description': u'ElastiCache parameter group for journal--end2end cluster 2',
                                                               u'Properties': {u'maxmemory-policy': u'volatile-ttl'}},
                                               u'Type': u'AWS::ElastiCache::ParameterGroup'}}}
2018-03-05 13:09:40,209 - INFO - MainProcess - cfn - Update: {'Outputs': {},
 'Resources': {u'ElastiCache1': {u'Properties': {u'CacheNodeType': u'cache.t2.small',
                                                 u'CacheParameterGroupName': {u'Ref': u'ElastiCacheParameterGroup1'},
                                                 u'CacheSubnetGroupName': {u'Ref': u'ElastiCacheSubnetGroup'},
                                                 u'Engine': u'redis',
                                                 u'EngineVersion': u'2.8.24',
                                                 u'NumCacheNodes': 1,
                                                 u'PreferredAvailabilityZone': u'us-east-1d',
                                                 u'Tags': [{u'Key': u'Cluster',
                                                            u'Value': u'journal--end2end'},
                                                           {u'Key': u'Environment',
                                                            u'Value': u'end2end'},
                                                           {u'Key': u'Name',
                                                            u'Value': u'journal--end2end'},
                                                           {u'Key': u'Project',
                                                            u'Value': u'journal'}],
                                                 u'VpcSecurityGroupIds': [{u'Ref': u'ElastiCacheSecurityGroup'}]},
                                 u'Type': u'AWS::ElastiCache::CacheCluster'},
               u'ElastiCache2': {u'Properties': {u'CacheNodeType': u'cache.t2.micro',
                                                 u'CacheParameterGroupName': {u'Ref': u'ElastiCacheParameterGroup2'},
                                                 u'CacheSubnetGroupName': {u'Ref': u'ElastiCacheSubnetGroup'},
                                                 u'Engine': u'redis',
                                                 u'EngineVersion': u'2.8.24',
                                                 u'NumCacheNodes': 1,
                                                 u'PreferredAvailabilityZone': u'us-east-1d',
                                                 u'Tags': [{u'Key': u'Cluster',
                                                            u'Value': u'journal--end2end'},
                                                           {u'Key': u'Environment',
                                                            u'Value': u'end2end'},
                                                           {u'Key': u'Name',
                                                            u'Value': u'journal--end2end'},
                                                           {u'Key': u'Project',
                                                            u'Value': u'journal'}],
                                                 u'VpcSecurityGroupIds': [{u'Ref': u'ElastiCacheSecurityGroup'}]},
                                 u'Type': u'AWS::ElastiCache::CacheCluster'}}}
2018-03-05 13:09:40,210 - INFO - MainProcess - cfn - Delete: {'Outputs': {},
 'Resources': {u'ElastiCacheParameterGroup': {u'Properties': {u'CacheParameterGroupFamily': u'redis2.8',
                                                              u'Description': u'ElastiCache parameter group for journal--end2end',
                                                              u'Properties': {u'maxmemory-policy': u'volatile-ttl'}},
                                              u'Type': u'AWS::ElastiCache::ParameterGroup'}}}
```
This should change nothing, but switch every ElastiCache cluster to its own named configuration. Have to check what does that interrupt (clears data or just temporary interruption).
Then, we can remove overrides to selectively switch to the new policy.

giorgiosironi:
Let's try `journal--end2end`.

giorgiosironi:
The `Bytes Used For Cache` graphs seem unaffected, so cache should not be cleared by this configuration change.

giorgiosironi:
Everything has been applied to `end2end`, `continuumtest`, `prod`.

PR #231, opened 2018-01-10
mysql standalone config missing important config details

lsh-0:
this is the cause of the formula failures

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1588,6 +1588,13 @@ crm:
                 ami: ami-92002785 # Ubuntu 14.04, deprecated
                 masterless: true
             rds:
+                backup-retention: 2 # days
+                engine: MySQL
+                version: '5.5'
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -1588,6 +1588,13 @@ crm:
                 ami: ami-92002785 # Ubuntu 14.04, deprecated
                 masterless: true
             rds:
+                backup-retention: 2 # days
+                engine: MySQL
+                version: '5.5'
+                storage: 30 # GB
+                type: db.t2.medium
+                params:
+                    log_bin_trust_function_creators: 1
```
giorgiosironi: I thought these were going to be inherited from `defaults`, but good that now it works

lsh-0: we could probably leave `storage` and `type` to inherit from defaults, but those params are civi-specific

PR #232, opened 2018-01-12
[wip] some tweaks to commands that have been hanging around for a bit

lsh-0:
pretty rough PR. it tweaks some of the commands in the askmaster module and introduces `continuum.py` commands, a module I thought I added ages and ages ago but apparently didn't. it has a handy command in there for identifying adhoc instances.

before/if this is merged I want to fix the problem with not being able to download a file from a multi-node stack. I keep 'fixing' it and then reverting it because the changes go a little deep.

### DIFF
--- a/src/askmaster.py
+++ b/src/askmaster.py
@@ -55,8 +55,8 @@ def installed_linux():
 
 @task
 def patched():
-    #output = str(installed_linux()).splitlines()
-    output = open('src/foo.txt', 'r').readlines()
+    output = str(installed_linux()).splitlines()
+    #output = open('src/foo.txt', 'r').readlines()
 
     def rowfn(row):
         # parses output of dpkg -l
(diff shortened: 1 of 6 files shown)


src/askmaster.py
```diff
@@ -47,3 +47,78 @@ def fail2ban_running():
 @task
 def installed_linux():
     return salt_master_cmd("'dpkg -l | grep -i linux-image && uname -r'")
+
+
```
giorgiosironi: a
bit
of 
whitespace? :)

src/askmaster.py
```diff
@@ -45,4 +46,79 @@ def fail2ban_running():
...
+#
+#
+
+
+
+
+@task
+def patched():
+    #output = str(installed_linux()).splitlines()
+    output = open('src/foo.txt', 'r').readlines()
```
giorgiosironi: probably the filename should be an input?

src/continuum.py
```diff
@@ -0,0 +1,37 @@
...
+def observer_backfill(stackname):
+    with core.stack_conn(stackname), cd('/srv/observer/'):
+        run('./manage.sh load_from_api')
+
+@task
+@echo_output
+def adhoc_instances():
+    def unrecognised(stackname):
+        iid = stackname.split('--')[-1]
+        return iid not in ['ci', 'end2end', 'prod', 'continuumtest']
```
giorgiosironi: `demo`, `continuumtestpreview`, `preview`

PR #233, opened 2018-01-15
ELPP-3360 bot hostname and ports for access in end2end tests

giorgiosironi:
- Add a public domain name to bot instances
- Add ports to expose the fake FTP server content

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -626,6 +626,10 @@ elife-bot:
                 "{instance}-elife-bot-sessions":
                     deletion-policy: delete
         end2end:
+            ports:
+                - 22
+                - 8020 # exposing FTP contents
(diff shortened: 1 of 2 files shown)


projects/elife.yaml
```diff
@@ -627,6 +626,10 @@ elife-bot:
                 "{instance}-elife-bot-sessions":
                     deletion-policy: delete
         end2end:
+            ports:
+                - 22
+                - 8020 # exposing FTP contents
```
lsh-0: is this also http? (just regular http though?)

giorgiosironi: yes, if we are accessing it through `elife.internal` it needs to be HTTP rather than HTTPS

giorgiosironi:
Applying to `ci`, `continuumtest` and `prod`; then merging.

PR #234, opened 2018-01-18
Smarter diff in EC2 updates

giorgiosironi:
EC2 have some properties that we never want to update such as UserData. If we did so, we may cause the re-creation of the instance losing its state.

The previous comparison stripped these properties from the old and the new template, so that the resource wasn't detected as modified at all in most cases. This new comparison actually allows to modify some of the properties, as the non-updatable ones will be copied from the old template


giorgiosironi:
Tried using this change on `elife-bot--prod` but it was so old it couldn't even be applied. Immutable servers or containers are easier to deal with, so closing.

PR #236, opened 2018-01-23
ELPP-3338 Allow headers forwarding to be configured per-origin

giorgiosironi:
Turns out we already had some support for multiple origins:

- one additional origin is required for error pages
- some CDNs merge contents from multiple buckets 

What was missing was being able to customize the `CacheBehavior` of each origin. In this we case, we add the capability to customize which headers are forwarded so that some paths (like `/log-in) can benefit from more info about the user; they sacrifice cacheability as they are not cacheable anyway.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -441,6 +441,9 @@ journal:
                             - Host
                             - CloudFront-Viewer-Country
                             - Referer
+                        cookies:
+                            - journal
+                            - AWSELB
                 #logging:
                 #    bucket: elife-cloudfront-logs
         continuumtest:
@@ -474,6 +477,8 @@ journal:
                             - Host
                             - CloudFront-Viewer-Country
                             - Referer
+                        cookies:
+                            - journal
         prod:
             description: prod environment for journal. ELB on top of multiple EC2 instances
             ec2:
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -518,6 +518,20 @@ journal:
                         # ELB with no active instances
                         503: "/5xx.html"
                     protocol: http
+                origins:
+                    CloudFrontCDNOrigin:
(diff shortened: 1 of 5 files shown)


giorgiosironi:
There is nothing sensible in the templates (a certificate id that still can't be accessed), so here is the before&after for comparison:
[before.txt](https://github.com/elifesciences/builder/files/1656661/before.txt)
[after.txt](https://github.com/elifesciences/builder/files/1656660/after.txt)

thewilkybarkid:
Could `errors` be part of this too?

projects/elife.yaml
```diff
@@ -430,6 +430,17 @@ journal:
...
+                origins:
+                    CloudFrontCDNOrigin:
+                        hostname: end2end--journal.elifesciences.org
+                    LogInOrigin:
+                        hostname: end2end--journal.elifesciences.org
+                        pattern: /log-in
+                        headers:
+                            - X-Requested-With
+                            - Host
+                            - CloudFront-Viewer-Country
```
thewilkybarkid: Maybe these should inherit?

giorgiosironi: Advantage of inheriting: we avoid specifying these twice. Disadvantage: we cannot remove headers for a custom origin (it could go to a bucket or other static source that doesn't need those, like in the case of `error-pages`). Can probably wait to see if necessary.

giorgiosironi:
> Could errors be part of this too?

Sort of, the `CustomErrorResponses` which is specific to it needs a separate configuration; its `CacheBehavior` could be modelled like all the others instead. One is which errors to deal with, the other is where to get the error pages from.

projects/elife.yaml
thewilkybarkid: Shouldn't need to remove, just add the headers at the appropriate level. But yeah, not that important (yet).

PR #239, opened 2018-02-06
Remove AWSELB cookie whitelisting from journal configurations

giorgiosironi:
Not relying upon this anymore; incidentally it may improve the cache rate as pages do not need to be cached separately depending on which server generated them.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -421,7 +421,6 @@ journal:
                     - Host
                 cookies:
                     - journal
-                    - AWSELB
                 errors: 
                     # TODO: this is not being deployed to, it's empty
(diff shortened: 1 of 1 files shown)


thewilkybarkid:
(Probably worth just double-checking this on `continuumtest`.)

giorgiosironi:
> (Probably worth just double-checking this on continuumtest.)

Not enabled there at all, since the stack contains a single EC2 instance and no load balancer.

thewilkybarkid:
Ah, `end2end`.

giorgiosironi:
I'll apply to `end2end` and re-run tests before going to `prod`.

giorgiosironi:
https://alfred.elifesciences.org/job/test-elife-spectrum/195/console is green, doing `prod` now.

PR #241, opened 2018-02-12
observer, removed elife-metrics events. listens to press-packages events

lsh-0:
elife-metrics is now updated via cron job

PR depends on https://github.com/elifesciences/observer/pull/37
PR related to: https://github.com/elifesciences/observer/pull/38

### DIFF
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -69,7 +69,7 @@ STEADY_CFN_STATUS = [
 
 #
 # sns
-# 
+#
 
 def _set_raw_subscription_attribute(sns_connection, subscription_arn):
     """
@@ -91,8 +91,8 @@ sns.connection.SNSConnection.set_raw_subscription_attribute = _set_raw_subscript
 def all_sns_subscriptions(region, stackname=None):
     """returns all subscriptions to all sns topics.
     optionally filtered by subscription endpoints matching given stack"""
-    sns = boto_sns_conn(region)
-    response = sns.get_all_subscriptions() # TODO: needs pagination!
+    conn = boto_sns_conn(region)
+    response = conn.get_all_subscriptions() # TODO: needs pagination!
     subs_list = response['ListSubscriptionsResponse']['ListSubscriptionsResult']['Subscriptions']
     if stackname:
         ''' looks like:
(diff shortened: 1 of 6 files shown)


lsh-0:
FYI @seanwiseman , this is where an application describes the queues it wants to subscribe to for updates

giorgiosironi:
I'm not sure there is code that actually removes the missing subscriptions. CloudFormation didn't support adding a (SQS) Queue that listened to existing (SNS) Topics, so this is implemented through API calls.

giorgiosironi:
Confirmed, https://github.com/elifesciences/builder/blob/master/src/buildercore/bootstrap.py#L190 is incomplete and should be expanded to get a list of subscriptions of the queue first, and then remove all the ones that aren't explicitly listed.

It seems it's a bit clunky to get a list however: https://boto.readthedocs.io/en/glacier/ref/sns.html#boto.sns.SNSConnection.get_all_subscriptions needs to be filtered in some way.

lsh-0:
take a look at my changes and tell me what you think. it's not debugged or tested, I've run out of time today, but I think it's heading in the right direction

src/buildercore/core.py
```diff
@@ -85,6 +88,24 @@ def _set_raw_subscription_attribute(sns_connection, subscription_arn):
 
 sns.connection.SNSConnection.set_raw_subscription_attribute = _set_raw_subscription_attribute
 
+def all_sns_subscriptions(region, stackname=None):
+    """returns all subscriptions to all sns topics.
+    optionally filtered by subscription endpoints matching given stack"""
+    conn = boto_sns_conn(region)
+    response = conn.get_all_subscriptions() # TODO: needs pagination!
```
giorgiosironi: shouldn't boto (or boto3) do pagination for you?

giorgiosironi: (perhaps with other Python functions provided, not sure)

src/buildercore/core.py
```diff
@@ -85,6 +88,24 @@ def _set_raw_subscription_attribute(sns_connection, subscription_arn):
...
+    subs_list = response['ListSubscriptionsResponse']['ListSubscriptionsResult']['Subscriptions']
+    if stackname:
+        ''' looks like:
+        {u'Endpoint': u'arn:aws:sqs:us-east-1:512686554592:observer--substest1',
+         u'Owner': u'512686554592',
+         u'Protocol': u'sqs',
+         u'SubscriptionArn': u'arn:aws:sns:us-east-1:512686554592:bus-articles--substest1:f44c42db-81c0-4504-b3de-51b0fb1099ff',
+         u'TopicArn': u'arn:aws:sns:us-east-1:512686554592:bus-articles--substest1'}'''
+        subs_list = filter(lambda row: row['Endpoint'].endswith(stackname), subs_list)
```
giorgiosironi: `Endpoint` seems a strange term for this, but from this response it really looks like the only place that can be used for filtering

src/buildercore/core.py
lsh-0: boto3 does pagination for you at a certain level. I've never seen boto2 do pagination for me

src/buildercore/core.py
lsh-0: tell me about it, I spent far too long trying to figure out just what it meant and how I could get access to what I need through which api.

lsh-0:
@giorgiosironi , I've split up `bootstrap.setup_sqs` and added a pair of tests. Could you take a look and let me if you approve?

lsh-0:
PR is still waiting on Observer to stop breaking in end2end

lsh-0:
observer is all fixed up now (thanks @giorgiosironi) merging this in

PR #242, opened 2018-02-16
Feat salt upgrade 2018.02

lsh-0:


lsh-0:
I'm happy to merge and update master and minions on Monday, presuming nothing bad jumps out at me in the release nots

giorgiosironi:
Ready to go? Trying to keep open pull requests to a minimum.

lsh-0:
Fair enough. I want to deploy a 16.04 master instance this Monday actually, which I'm sure will break all the github triggers like last time (unless we fixed that?) and then re-master all the instances. Doing that on a Friday just seemed mean spirited.

giorgiosironi:
It won't break the hooks of the existing master, but the new one needs to be added here:
https://github.com/organizations/elifesciences/settings/hooks

scripts/bootstrap.sh
```diff
@@ -119,6 +119,8 @@ if $installing; then echo "$(date -I) -- installed $version" >> /root/events.log
 if $upgrading; then echo "$(date -I) -- upgraded to $version" >> /root/events.log; fi
 
 
+# BUG: during a minion's re-mastering the `master: ...` value may get reset if the instance is 
+# updated by another process before the old master is turned off.
```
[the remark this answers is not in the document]
lsh-0: doing that programmatically would be a valuable addition to builder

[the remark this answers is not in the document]
giorgiosironi: terraform has a locking system which may help, but I don't know if it can be surfaced to the user with acquire/release primitives https://www.terraform.io/docs/state/locking.html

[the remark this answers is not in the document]
lsh-0: this is actually a severe bug and I misrepresented it. It should read "during the re-mastering transition", meaning, while there is more than one master server active any update will wipe out the master ip info and replace it with the IP of the oldest master-server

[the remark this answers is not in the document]
lsh-0: if we stored the master server IP in the buildvars as well, it might solve this elegantly.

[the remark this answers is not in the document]
giorgiosironi: It looks like all builds running today will re-master the minion to the old server then. Should block further migrations because they would be easily rolled back?

giorgiosironi:
Merging `master` as it's important to keep this branch up-to-date with various fixes.

src/master.py
```diff
@@ -81,22 +89,135 @@ def aws_update_projects(pname):
...
+    if context['ec2'].get('master_ip') == master_ip:
+        LOG.info("already remastered: %s", stackname)
+        try:
+            utils.get_input('any key to skip, ctrl-c to carry on')
+            return
+        except KeyboardInterrupt:
+            LOG.info("not skipping")
+
+    LOG.info("setting new master address")
+    context = cfngen.set_master_address(context, master_ip) # mutator
```
giorgiosironi: if this is a mutator it shouldn't really return anything, just set stuff in `context`?

lsh-0:
@giorgiosironi , could I get your feedback on the updated docs (and anything else you want).

docs/master-server.md
```diff
@@ -24,6 +26,34 @@ to complete the update. All Salt states shown should be green.
 
 ## updating the master server
 
-Whenever a change is made to the formulas or to your builder-private repository, the master server needs to be updated. It will do so every hour through a cron job, but you can always do this immediately after you want a change to be applied with:
+Whenever a change is made to the formulas, or to your `builder-private` repository, or a minion is destroyed, the master server needs to be updated. It will do so every hour through a cron job but can also be run immediately with:
+
+	./bldr master.update:master-server--prod
```
giorgiosironi: There is potential confusion with `./bldr update:master-server--prod`, state the difference (if there is one)

docs/master-server.md
```diff
@@ -24,6 +26,34 @@ to complete the update. All Salt states shown should be green.
...
+
+    ./bldr master.remaster:project--ci,master-server--newinstanceid
+    
+To re-master *all* minions:
+
+    ./bldr master.remaster_all:master-server--newinstanceid
+    
+The process of re-mastering a minion roughly follows:
+
+* replace the value of `aws.ec2.master_ip` in the `context` data
```
giorgiosironi: still has the problem that if someone runs `update_infrastructure`, the `context` is regenerated and the switch is done. So a preliminary step here should be to do what we did: pin the new master ip address in `projects/elife.yaml`. Then remastering is a quicker switch _towards what is already pinned there_, hence it won't be rolled back.

py2-requirements.txt
```diff
@@ -1,23 +1,24 @@
...
+python-dateutil==2.6.1
+python-slugify==1.2.4
 git+https://github.com/beelit94/python-terraform.git@develop#egg=python-terraform
-pytz==2016.10
+pytz==2018.3
 PyYAML==3.12
-requests==2.13.0
-troposphere==1.9.2
+requests==2.18.4
+troposphere==2.2.1
```
giorgiosironi: new major version :thinking: 
Where did all this come from? Should not be tied to the update of Salt

giorgiosironi: It seems it is bringing Python 2 dependencies up to the same versions as Python 3 one, but it should be a separate PR; perhaps there is some bug that needed fixing via upgrading?

src/master.py
```diff
@@ -68,34 +78,171 @@ def server_access():
...
-
-    if not master_ip:
-        newest_master = last(core.find_master_servers(core.active_aws_stacks(core.find_region())))
-        if not newest_master:
-            raise core.NoMasterException("no master servers found")
-        master_ip = core.stack_data(newest_master)[0]['private_ip_address']
-
-    print('re-mastering %s to %s' % (stackname, master_ip))
-
+    # TODO: turn this into a decorator
```
giorgiosironi: Yes, should be turned into something else to avoid leaving an `import` in the wrong scope

PR #243, opened 2018-02-19
Bugfix sqs subscriptions

lsh-0:
had an opportunity today to test the unsub/sub update on an actual instance and found some problems: I wasn't unsubscribing anything and you were subscribing to only what was present in the template. We both were ignoring the project data.

I had mistaken the context_sqs for project data so it was never returning any changes. After some mucking about I have it working, but I can't seem to get rid of a permission:

screenshot at 2018-02-19 13-35-53

(metrics shouldn't be there)

I saw your `parameterise` functions around but decided to integrate it better.

these functions could really do with a cleanup, but lets get the proper behaviour happening first


src/buildercore/bootstrap.py
```diff
@@ -172,16 +172,25 @@ def update_sqs_stack(stackname, **kwargs):
...
     setup_sqs(stackname, context.get('sqs', {}), pdata['aws']['region'])
 
-def unsub_sqs(stackname, context_sqs, region, dry_run=False):
+def unsub_sqs(stackname, cfn_sqs, region, dry_run=False):
     sublist = core.all_sns_subscriptions(region, stackname)
-    struct = {}
-    for queue_name in context_sqs:
-        subscriptions = context_sqs[queue_name] # ll: [bus-articles--prod, ...]
-        struct[queue_name] = [sub for sub in sublist if sub['Topic'] not in subscriptions]
+    prj_sqs = project_data_for_stackname(stackname, interpolate=True)['aws']['sqs']
```
giorgiosironi: Why don't we have everything into `context` to do unsubscriptions? This seems like a layer violation as it shouldn't be needed to recompute project data at this level.

giorgiosironi: (and afaik it also may not substitute `{instance}` with `prod`, and similar

src/buildercore/bootstrap.py
```diff
@@ -172,16 +172,25 @@ def update_sqs_stack(stackname, **kwargs):
...
+        unsub_map[queue_name] = [sub for sub in sublist if sub['Topic'] not in subscriptions]
+
     if not dry_run:
         sns = core.boto_sns_conn(region)
-        [sns.unsubscribe(sub['TopicArn']) for sub in utils.shallow_flatten(struct.values())]
-    return struct
+        for sub in utils.shallow_flatten(unsub_map.values()):
+            LOG.info("Unsubscribing %s from %s", sub['Topic'], stackname)
+            subarn = sub['SubscriptionArn']
+            sns.set_raw_subscription_attribute(subarn, val=False) # necessary? SQS:SendMessage permission still present
```
giorgiosironi: IIUC this affect only this subscription and not the other subscriptions to the topic (which may come from other projects).
If this is true, the subscription is going to be deleted on the next line, so we could avoid changing its attributed.

Note this is not a permission: it is a setting that says not to do any wrapping on the JSON object being delivered to the queue (we always use `True`).

src/buildercore/project/__init__.py
```diff
@@ -103,11 +103,20 @@ def project_list(project_locations_list=None):
     "returns a single list of projects, ignoring organization and project data"
     return project_map(project_locations_list).keys()
 
-def project_data(pname, project_locations_list=None):
-    "returns the data for a single project"
+def project_data(pname, project_locations_list=None, format_values=None):
+    "returns the data for a single project. interpolates {value} strings if `format_values` is non-nil"
```
giorgiosironi: seems too magic to arbitrarily replace the `{value}` strings, with respect to the careful support in `buildercore.cfngen`

src/buildercore/bootstrap.py
```diff
@@ -172,16 +172,25 @@ def update_sqs_stack(stackname, **kwargs):
     context = context_handler.load_context(stackname)
     setup_sqs(stackname, context.get('sqs', {}), pdata['aws']['region'])
 
-def unsub_sqs(stackname, context_sqs, region, dry_run=False):
+def unsub_sqs(stackname, cfn_sqs, region, dry_run=False):
```
giorgiosironi: The second parameter is actually unused in the method body so this may not be apparent, but a `context` sub-dictionary is passed in, not anything that should be tied to CloudFormation. It comes from
```
 setup_sqs(stackname, context.get('sqs', {}), pdata['aws']['region'])
```

src/buildercore/bootstrap.py
lsh-0: ah ok, I was trying all sorts of things. I'll revert this

lsh-0:
`context_sqs` simply doesn't have the changes to subscriptions. I'll look again, and use your preferred method if I'm wrong, but I didn't see it.

lsh-0:
`context_sqs` definitely doesn't have any project changes in there, so I'll re-generate the build vars again via cfngen and include them somehow.

lsh-0:
big change. could I get you to respond to this chunk specifically, @giorgiosironi ? https://github.com/elifesciences/builder/pull/243/files#diff-11aa22c619b41fcc4ac5151f83decddcR517

if there is a reason that download and save to disk happens, it might be better off in it's own function with an explanation in `regenerate...` as to what it's doing.

src/buildercore/cfngen.py
```diff
@@ -484,10 +491,34 @@ def apply_delta(template, delta):
...
+# def regenerate_stack(pname, **more_context):
+#    current_template = bootstrap.current_template(more_context['stackname'])
+#    current_context = context_handler.load_context(more_context['stackname'])
+#    write_template(more_context['stackname'], json.dumps(current_template))
+#    context = build_context(pname, existing_context=current_context, **more_context)
+#    delta = template_delta(pname, context)
+#    return context, delta, current_context
+
+def regenerate_stack(stackname, **more_context):
+    # what was the point of these two lines? it downloads the template body and saves it to disk and never uses it ...
```
giorgiosironi: The point of downloading the template is that `template_delta` has to generate a diff (delta) to show it to the user (as updating the template is a dangerous operation). To generate the delta, it needs to get the template from the authoritative source (the CloudFormation API) and then compare it with the new one. The filesystem is acting as a light caching layer for the template as provided by the API, so it may just be a simplification that is worth the performance trade-off. Stacks are unfrequently updated.

src/buildercore/cfngen.py
```diff
@@ -366,7 +373,7 @@ class Delta(namedtuple('Delta', ['plus', 'edit', 'minus'])):
     def non_empty(self):
         return self.plus['Resources'] or self.plus['Outputs'] or self.edit['Resources'] or self.edit['Outputs'] or self.minus['Resources'] or self.minus['Outputs']
 
-def template_delta(pname, context):
+def template_delta(context):
```
giorgiosironi: nice catch, so much for linting which should have seen this unused argument?

src/buildercore/cfngen.py
giorgiosironi: `template_delta` is using `read_template` which only looks at the filesystem at the moment, for this to work it would need to be modified to download it.

giorgiosironi:
Added a sample project and a bugfix that came out from launching it. This is what we have for the sample instance:
```
$ cat .cfn/contexts/sample-sqs--end2end.json | jq .sqs
{
  "sample-sqs--end2end": [
    "bus-articles--end2end",
    "bus-press-packages--end2end"
  ]
}
$ cat .cfn/stacks/sample-sqs--end2end.json
{
    "Outputs": {
        "SampleSqsEnd2endQueueArn": {
            "Value": {
                "Fn::GetAtt": [
                    "SampleSqsEnd2endQueue",
                    "Arn"
                ]
            }
        }
    },
    "Resources": {
        "SampleSqsEnd2endQueue": {
            "Properties": {
                "QueueName": "sample-sqs--end2end"
            },
            "Type": "AWS::SQS::Queue"
        }
    }
}
```
Now I'm going to try removal.

giorgiosironi:
Neither `update_template` nor `update` removed the subscription (though they updated the context).

I think I understand where the problem is coming from, we have two update methods but none of them covers this case:
- `update` (run by every deploy): runs highstate; also included the creation of SQS subscriptions so far, perhaps erroneously. Doesn't touch the `context` nor the `template`.
- `update_template` (run manually upon infrastructure changes): check if there are changes to `template`, if so asks for confirmation and applies them before going on with `update`. Regenerates both `context` and `template`.

I think it's clear now that this case of subscribing/unsubscribing should go only in `update_template` rather than being executed at every build (e.g. it wouldn't have made the `prod-observer` pipeline fail). I'll try to move that later, as we'll also need to make `update_template` do the SQS updates.

First step: `setup_sqs` as part of `update` should read the new context and apply it.

giorgiosironi:
With the latest commits I can add and remove subscriptions in a few seconds, which proves `unsub_sqs` works very well if given the right input context. It follows that is only needs the current context rather than needing to regenerate it, as `update_template` will have already done that every time.

src/buildercore/utils.py
```diff
@@ -36,6 +36,19 @@ def dictfilter(func, ddict):
 def dictmap(fn, ddict):
     return {key: fn(key, val) for key, val in ddict.items()}
 
+def nested_dictmap(fn, ddict):
```
giorgiosironi: where is this function used? I can't find a call to it apart from tests

giorgiosironi:
Going to get the tests green, than working on the removal of the leftover permissions.

src/buildercore/utils.py
lsh-0: it was part of the project vars interpolation that I reverted. I opted to keep it because it was useful. I plan to extract all my little utilities into their own module so I don't keep duplicating them across projects. Will open a ticket and talk to Sean about it.

lsh-0:
yup, sorry I didn't address this PR earlier, I meant to Weds but two other things ended up eating up my whole day. I'll take a look first thing

src/buildercore/bootstrap.py
```diff
@@ -164,24 +164,42 @@ def is_resourcing():
...
-        return
-    context = context_handler.load_context(stackname)
-    setup_sqs(stackname, context.get('sqs', {}), pdata['aws']['region'])
+    These statements are created by boto's subscribe_sqs_queue()"""
 
-def unsub_sqs(stackname, context_sqs, region, dry_run=False):
+    def for_unsubbed_topic(statement):
+        return statement.get('Condition', {}).get('StringLike', {}).get('aws:SourceArn') in topic_arns
+
+    policy['Statement'] = list(filter(lambda s: not for_unsubbed_topic(s), policy['Statement']))
```
lsh-0: python2 map/filter/reduce type are not lazy, no need to realise result via `list`. 2to3 will 'helpfully' clean these up for you when builder gets upgraded. no change necessary, just sayin'.

src/buildercore/bootstrap.py
```diff
@@ -164,24 +164,42 @@ def is_resourcing():
...
         sns = core.boto_sns_conn(region)
-        [sns.unsubscribe(sub['TopicArn']) for sub in utils.shallow_flatten(struct.values())]
-    return struct
+        for sub in utils.shallow_flatten(unsub_map.values()):
+            LOG.info("Unsubscribing %s from %s", sub['Topic'], stackname)
+            sns.unsubscribe(sub['SubscriptionArn'])
+        sqs = core.boto_sqs_conn(region)
+
+        for queue_name, topic_arns in permission_map.items():
+            # not an atomic update, but there's no other way to do it
```
lsh-0: :+1:

lsh-0: I wonder why the sqs policy isn't removed when unsubscribed? this is obviously what I was missing

src/buildercore/cfngen.py
lsh-0: it's entirely possible pylint does catch this, but doesn't consider it an error. I would have thought pyflakes could catch this though.

lsh-0:
your changes look fine, @giorgiosironi , sorry for holding you up. please merge

src/buildercore/bootstrap.py
giorgiosironi: `Topic`, `Queue` and `Subscription` are three different objects. When unsubscribing, we delete the `Subscription` so things like `RawMessageDelivery that are attribute of the `Subscriptions` are removed as well. However, the `Policy` is a global property of the `Queue` so if you unsubscribe from just N out of M `Topics` it shouldn't be removed; rather, edited to remove references to these N topics but leaving alone the other statements referring to still-subscribed-to `Topics`.

giorgiosironi:
Running `bldr update_template:end2end` and `bldr update_template:observer--prod`.

giorgiosironi:
No problem, this PR was not blocking other tasks.

PR #244, opened 2018-02-19
upgraded linting to support built-in shadowing checks

lsh-0:


lsh-0:
eeeehhh - I know I should, deduplication and all, but I'd be duplicating many exceptions for each project or doing a massive lint operation to shorten every line to 80 chars etc. And I really don't agree with all the rules. It feels like make-work after a certain point.

Also, the python-sdk code is heavily OO which is perfect Luke-repellent - I probably won't be contributing to the code unless absolutely required to do so. I'm going to leave these as-is for the time being.

giorgiosironi:
I get the banter about OO, but:
- we are talking about https://github.com/elifesciences/proofreader-python which does not contain classes
- `boto`, `requests` and most of builder's dependencies are OO already

lsh-0:
there's more than one? I'll take a look at proofreader today and see what can be done.

> boto, requests and most of builder's dependencies are OO already

oh cmon - you don't think I don't know that?? 

that was a rhetorical question, I think you're just teasing me ;)

PR #246, opened 2018-02-20
disabled unsubbing in setup_sqs (for now)

lsh-0:
is crashing deploys to observer--prod. disabling here and will enable in PR #243


giorgiosironi:
:+1: good feature-flagging for the time being

PR #247, opened 2018-02-21
[wip] Feat run strip coverletter remotely

lsh-0:


giorgiosironi:
Seems suspicuously specific, but no big deal if it's isolated into a namespace of tasks (e.g. `./bldr elife_bot:...` or `./bldr continuum:...`)

lsh-0:
The task was very helpful but if we switch the strip-coverletter to a docker container it won't be necessary. I've left it as a [WIP] for the time being

PR #249, opened 2018-02-26
Remove minion key from master on destroy

giorgiosironi:
The intended behavior is that a stack can be recreated after `destroy`, without having to manually delete the key from the master-server's `/etc/salt/pki/...` folder.

However, needs to cater for:
- masterless instances (may just fail silently?)
- master servers themselves (can't destroy them?)

### DIFF
--- a/src/buildercore/utils.py
+++ b/src/buildercore/utils.py
@@ -65,6 +65,14 @@ def complement(pred):
 def splitfilter(func, data):
     return filter(func, data), filter(complement(func), data)
 
+def mkidx(fn, lst):
+    groups = {}
+    for v in lst:
+        key = fn(v)
+        grp = groups.get(key, [])
+        grp.append(v)
+        groups[key] = grp
+    return groups
 
 """
 # NOTE: works, unused.
--- a/src/buildercore/bootstrap.py
+++ b/src/buildercore/bootstrap.py
@@ -608,11 +608,42 @@ def delete_stack_file(stackname):
     return dict(zip(paths, map(_unlink, paths)))
 
 def remove_minion_key(stackname):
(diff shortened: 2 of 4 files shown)


lsh-0:
regular users can't access the master server for security reasons, right?  if they attempt to destroy a stack this step won't work

giorgiosironi:
Maybe it's better to keep it as a separate task then.

lsh-0:
I think so. Also wondering if there is a maintenance task that could be run periodically on master to detect destroyed minions and remove their keys. It could fetch the list of cfn templates and remove keys for those that have no match. That, or a secure way of sending events to the master like we do with github and updating formulas

giorgiosironi:
`update_master` is the task you're looking for (then someone needs to execute it, but it's easy to do from Alfred).

lsh-0:
added the logic, I'll let you do the honours and actually delete the keys ;)

src/buildercore/utils.py
```diff
@@ -65,6 +65,14 @@ def wrapper(*args, **kwargs):
 def splitfilter(func, data):
     return filter(func, data), filter(complement(func), data)
 
+def mkidx(fn, lst):
```
giorgiosironi: is there a Pylint check for too low percentage of vowels in a row? :nerd_face:

giorgiosironi:
Wow, they were more than 300.

src/buildercore/utils.py
lsh-0: > too low percentage of vowels in a row

probably, but I also don't consider it a problem. we should spend more time addressing low test coverage or low quality tests than linting. linting after a certain point is just nest feathering and self gratification.

PR #252, opened 2018-03-02
ELPP-3481 Add exeter-adapter project

giorgiosironi:
This project will be expanded with something reading from the queue, but now we just want to collect data.

Solves also a bug on `unsub_sqs` called on a queue not subscribed to anything. The `Policy` attribute cannot be empty, so it needs to be removed by setting an empty string when there are no topics to subscribe to.

Incidentally, even if a queue subscribed to topics, it will still run `unsub_sqs` at its creation when it's not subscribed to anything yet, running into the problem.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1725,6 +1725,8 @@ exeter-adapter:
     intdomain: False
     aws:
         ec2: false
+        sqs:
+            exeter-adapter--{instance}: {}
     aws-alt:
         continuumtest:
             sqs:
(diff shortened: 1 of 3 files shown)


giorgiosironi:
Don't know why I have to override the inherited `aws-alt` configuration using `ec2: false`, but only for this project.

giorgiosironi:
This is more urgent than I thought: I think the formulas of projects with SQS queues cannot be built until this solution is merged in.

lsh-0:
I'm going to merge this and give the failing observer build a kick, hope you don't mind

giorgiosironi:
:+1:

PR #253, opened 2018-03-09
node support for `download_file`

lsh-0:
a node parameter can now be specified to download_file to d/l from a specific node

### DIFF
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -258,10 +258,9 @@ def stack_conn(stackname, username=config.DEPLOY_USER, node=None, **kwargs):
 
     data = stack_data(stackname, ensure_single_instance=False)
     ensure(len(data) == 1 or node, "stack is clustered with %s nodes and no specific node provided" % len(data))
-    if node:
-        ensure(utils.isint(node) and int(node) > 0, "given node must be an integer and greater than zero")
-        didx = int(node) - 1 # decrement to a zero-based value
-    data = data[didx or 0] # data is ordered by node
+    node and ensure(utils.isint(node) and int(node) > 0, "given node must be an integer and greater than zero")
+    didx = int(node) - 1 if node else 0 # decrement to a zero-based value
+    data = data[didx] # data is ordered by node
     public_ip = data['ip_address']
     params = _ec2_connection_params(stackname, username, host_string=public_ip)
 
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -182,6 +182,10 @@ def find_ec2_instances(stackname, state='running', node_ids=None, allow_empty=Fa
         ec2_instances = [i for i in all_ec2_instances if i.state == state]
     else:
         ec2_instances = all_ec2_instances
+
+    # multiple instances are sorted by node asc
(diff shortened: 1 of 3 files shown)


src/cfn.py
```diff
@@ -295,7 +295,7 @@ def _interactive_ssh(command):
 
 @task
 @requires_aws_stack
-def download_file(stackname, path, destination='.', allow_missing="False", use_bootstrap_user="False"):
+def download_file(stackname, path, destination='.', node=None, allow_missing="False", use_bootstrap_user="False"):
```
giorgiosironi: seems consistent with `cmd`

src/buildercore/utils.py
```diff
@@ -11,6 +11,13 @@
 import logging
 LOG = logging.getLogger(__name__)
 
+def isint(v):
+    try:
+        int(v)
```
giorgiosironi: silent conversion:
```
>>> int(4.5)
4
```
You probably want `str.isdigit`?

src/buildercore/core.py
```diff
@@ -248,11 +252,16 @@ def _ec2_connection_params(stackname, username, **kwargs):
...
 
 @contextmanager
-def stack_conn(stackname, username=config.DEPLOY_USER, **kwargs):
+def stack_conn(stackname, username=config.DEPLOY_USER, node=None, **kwargs):
     if 'user' in kwargs:
         LOG.warn("found key 'user' in given kwargs - did you mean 'username' ??")
 
-    data = stack_data(stackname, ensure_single_instance=True)[0]
+    data = stack_data(stackname, ensure_single_instance=False)
+    ensure(len(data) == 1 or node, "stack is clustered and no node specified")
```
giorgiosironi: error message of this could include the value of `len(data)` `clustered (3 nodes)`

src/buildercore/core.py
```diff
@@ -248,11 +252,16 @@ def _ec2_connection_params(stackname, username, **kwargs):
...
+    if node:
+        ensure(utils.isint(node) and int(node) > 0, "given node must be an integer and greater than zero")
+        node = int(node) - 1 # decrement to a zero-based value
```
giorgiosironi: would not overwrite a method parameter with a value carrying a different meaning, `data_index` or `node_index` could be created with a default of `0`

src/buildercore/core.py
```diff
@@ -182,6 +182,10 @@ def find_ec2_instances(stackname, state='running', node_ids=None, allow_empty=Fa
         ec2_instances = [i for i in all_ec2_instances if i.state == state]
     else:
         ec2_instances = all_ec2_instances
+
+    # multiple instances are sorted by node asc
+    ec2_instances = sorted(ec2_instances, key=lambda ec2inst: ec2inst.tags.get('Node', 0))
```
giorgiosironi: this tag should be present everywhere by now, as we updated all tags for cost tracking

src/buildercore/core.py
lsh-0: this is a function, not a method. I can rename 'node' during decrement to something else

PR #254, opened 2018-03-12
new rules added to pylint's warning whitelist. linted

lsh-0:
added some more rules to the pylint warning whitelist while searching for a particular error I thought should be getting raised (referenced-before-assignment).

### DIFF
--- a/src/buildercore/trop.py
+++ b/src/buildercore/trop.py
@@ -522,8 +522,7 @@ def _add_bucket_policy(template, bucket_title, bucket_name):
 def _elb_protocols(context):
     if isinstance(context['elb']['protocol'], str):
         return [context['elb']['protocol']]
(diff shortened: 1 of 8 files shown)


src/buildercore/trop.py
```diff
@@ -912,16 +911,15 @@ def entry(hostname, i):
                     GetAtt(ELB_TITLE, "DNSName")
                 )
             )
-        else:
```
giorgiosironi: no `else` can contain a `return` essentially? No strong opinion, but if it pushes for simpler functions it can be a good thing

lsh-0: I think that's the jist of it and I agree - it's minutely simpler. 

It also gives an official stamp to one of my own pet peeves ;)

PR #256, opened 2018-03-14
requirements bump, switch to Fabric3

lsh-0:
Fabric3 is the unofficial py2/3 fork of Fabric.

Reading the Fabric roadmap and their issue tracker, I'm quite convinced they're yak shaving and what they do release will be a different product altogether. 

The unofficial fork is carrying the torch here and aims to be a drop-in replacement. I've dropped it in and it passes tests locally and I can connect to remote hosts etc.

This branch is an attempt to upgrade builder to python3


giorgiosironi:
The changes seem benign so far. Good to know there is a way out for Fabric.

lsh-0:
Quite benign. I don't think there is anything that we can't overcome here. I prevented `2to3` doing that thing where it wraps everything in `list(...)` so I still need to do a pass and check for unintended laziness but I'm optimisitic

lsh-0:
:tada:

.activate-venv.sh
```diff
@@ -13,12 +13,18 @@ else
     echo "* the no-delete-venv flag is set. preserving venv"
 fi
 
-if [ ! -d venv ]; then
-    # build venv if one doesn't exist
-    virtualenv --python=`which python2` venv
+python=/usr/bin/python3.5
```
giorgiosironi: Alfred seems to have no problem with Python 3 https://github.com/elifesciences/builder-private/blob/master/salt/top.sls#L263

.activate-venv.sh
```diff
@@ -13,12 +13,18 @@ else
...
+py=${python##*/} # ll: python3.5
+
+# build venv if one doesn't exist OR 
+# venv exists but the right python isn't installed
+if [ ! -e "venv/bin/$py" ]; then
+    echo "could not find venv/bin/$py, recreating venv"
+    rm -rf venv
+    $python -m venv venv
```
giorgiosironi: is this substituting `virtualenv --python=... venv`?

src/tests/test_buildercore_cfngen.py
```diff
@@ -122,27 +124,27 @@ def test_template_delta_includes_parts_of_cloudfront(self):
...
 
     def test_template_delta_includes_elb_security_group(self):
         "for consistency with EC2 security groups"
         context = self._base_context('project-with-cluster')
         context['elb']['protocol'] = 'https'
         context['elb']['certificate'] = 'DUMMY_CERTIFICATE'
         (delta_plus, delta_edit, delta_minus) = cfngen.template_delta(context)
-        self.assertEqual(delta_edit['Resources'].keys(), ['ElasticLoadBalancer', 'ELBSecurityGroup'])
-        self.assertEqual(delta_edit['Outputs'].keys(), [])
+        self.assertCountEqual(list(delta_edit['Resources'].keys()), ['ElasticLoadBalancer', 'ELBSecurityGroup'])
```
giorgiosironi: assertion method that wins the worst name in the year :)

.activate-venv.sh
lsh-0: yup, virtualenv is incorporated into python3 so no need to install it separately. 

python3 ships with a `venv` tool, so you could just do `venv myvenv`, but that has been deprecated so it's now as you see above: `python -m venv myvenv`

https://docs.python.org/3/library/venv.html

src/tests/test_buildercore_cfngen.py
lsh-0: tell me about it.

lsh-0:
I feel like there may be a few hidden gotchas involving laziness still hiding, so I'm going to merge this Monday rather than ruin everybody's Friday in case it's catastrophic.

Today I'll just merge in master and resolve any problems there.

After this it's the boto cleanup and common utils pruning.

lsh-0:
oh, and +1 for the integration tests - they take a while but it revealed a pernicious little bug

src/buildercore/core.py
```diff
@@ -157,16 +157,16 @@ def boto_s3_conn(region):
...
 @cached
 def connect_aws_with_pname(pname, service, with_boto3=False):
-    "convenience"
+    "convenience. returns a boto client for a service in same region as given project"
     pdata = project.project_data(pname)
     region = pdata['aws']['region']
     LOG.debug('connecting to a %s instance in region %s', pname, region)
     if with_boto3:
-        return boto3.client('rds', region)
+        return boto3.client('rds', region) # BUG: this should be `service`, right?
```
lsh-0: @giorgiosironi ?

giorgiosironi: Yes, probably it was the only value used so far

giorgiosironi:
I think I saw a failure of `bldr update_master` today related to encode/decode. Can you test if that has any problems for you?

lsh-0:
will do

PR #257, opened 2018-03-19
SQS notifications from silent-corrections bucket

giorgiosironi:
For https://github.com/elifesciences/elife-bot/pull/664#discussion_r175234319 and cannot be applied before that is merged.

This seems implemented in https://github.com/elifesciences/builder/blob/master/src/buildercore/bootstrap.py#L273 not through CloudFormation, which probably doesn't support if for already existing queues, but through Boto directly.


giorgiosironi:
Just pointing out that the notification is on `ObjectCreate` so new silent corrections mean creating new files, not modifying the existing ones.

giorgiosironi:
Applied everywhere.

PR #258, opened 2018-03-23
Feat python 3 get put

lsh-0:
doing some adhoc testing of the python3 branch revealed some more issues with uploading and downloading data.

this introduces a simpler interface for uploading and downloading files/data to/from a remote host and containing all the encoding wrangling.

### DIFF
--- a/src/buildercore/utils.py
+++ b/src/buildercore/utils.py
@@ -1,6 +1,6 @@
 import pytz
 import os, sys, copy, json, time, random, string
-from io import StringIO
+from io import StringIO, BytesIO
 from functools import wraps
 from datetime import datetime
 import yaml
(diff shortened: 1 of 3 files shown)


src/buildercore/utils.py
```diff
@@ -308,3 +309,33 @@ def hasallkeys(ddict, key_list):
...
+    "wrapper around fabric.operations.put"
+    label = label or local_path
+    msg = "uploading %s to %s" % (label, remote_path)
+    LOG.info(msg)
+    put(local_path=local_path, remote_path=remote_path, use_sudo=use_sudo)
+    return remote_path
+
+def fab_put_data(data, remote_path, use_sudo=False):
+    ensure(isinstance(data, bytes) or isinstance(data, str), "data must be bytes or a string that can be encoded to bytes")
+    data = data if isinstance(data, bytes) else data.encode()
```
giorgiosironi: maybe specify `utf-8` as an argument to `encode()` to be unambiguous?

lsh-0: I think as we get more familiar with python3 the `.encode` and `.decode` calls will become very familiar.

PR #259, opened 2018-03-27
Bugfix: yaml_dumps always returns a string

lsh-0:
ticket: https://elifesciences.atlassian.net/browse/ELPP-3567

please merge if you're happy with this @giorgiosironi

### DIFF
--- a/src/buildercore/project/files.py
+++ b/src/buildercore/project/files.py
@@ -18,6 +18,7 @@ def update_project_file(path, value, pdata, project_file):
     utils.updatein(pdata, path, value, create=True)
     return pdata
 
+# TODO: deletion candidate
 @testme
 def write_project_file(new_project_data, project_file):
     data = utils.yaml_dumps(new_project_data)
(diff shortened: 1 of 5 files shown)


src/buildercore/project/files.py
```diff
@@ -18,6 +18,7 @@ def update_project_file(path, value, pdata, project_file):
     utils.updatein(pdata, path, value, create=True)
     return pdata
 
+# TODO: deletion candidate
 @testme
```
giorgiosironi: I agree, we should not use a file both as a source and of data as a target of generation

lsh-0: it's from the earlier days of the builder where the projects file could be post-processed and updated. Comments weren't retained but I could always go back through the diff and restore them.

I'm adding the comments as I see a code removal branch in builder's near future to trim it down a bit

giorgiosironi: I see that thanks to DST-magic we cross over a bit now

lsh-0: It's also 7pm here and I can't seem to stop working today ;)

Stopping now. I need food.

PR #260, opened 2018-03-27
ELPP-3520 Create Fastly CDN through Terraform

giorgiosironi:
Expands the output target of builder from CloudFormation to include Terraform. Uses the Fastly Terraform provider to build a minimal Fastly CDN pointing to the full hostname of a stack.

Current status: held together with duct tape


projects/elife.yaml
```diff
@@ -333,7 +333,9 @@ api-gateway:
             - 443
             # - 8000: # don't expose this to public. Kong uses this to proxy requests
             # - 8001: # don't ever expose to public. Kong uses this for API admin
-
+        fastly:
+            subdomains:
+                - "{instance}--fastly-gateway"
```
giorgiosironi: Automatically points to `full_hostname`. More than one should be supported, also it's useful to have `subdomains-without-dns` too which will allow migrations (where both CloudFront and Fastly CDNs are in place and the DNS is switched from one to another during the transition period).

requirements.txt
```diff
@@ -14,6 +14,7 @@ pyflakes==1.6.0
 pylint==1.8.2
 python-dateutil==2.7.0
 python-slugify==1.2.4
+git+https://github.com/beelit94/python-terraform.git@develop#egg=python-terraform
```
giorgiosironi: The stablest Python library for Terraform I could find, will pin it to a commit or to the latest release when `0.10.0` is out.

src/buildercore/bootstrap.py
```diff
@@ -96,6 +97,8 @@ def _create_generic_stack(stackname, parameters=None, on_start=_noop, on_error=_
         context = context_handler.load_context(stackname)
         # setup various resources after creation, where necessary
         setup_ec2(stackname, context['ec2'])
+        # since setup_ec2 doesn't actually set up, this may have to be moved
```
giorgiosironi: there is no reason to wait until EC2 is up, so can be done as soon as `_wait_until_in_progress` has confirmed CloudFormation has completed the creation of the stack. I wouldn't run Terraform and CloudFormation in parallel for simplicity.

src/buildercore/bootstrap.py
```diff
@@ -139,6 +142,7 @@ def is_updating(stackname):
            "Failed to create stack: %s.\nEvents: %s" % (final_stack.stack_status, pformat(events)))
 
 def setup_ec2(stackname, context_ec2):
+    # TODO: misleading name, doesn't set up anything but only waits
```
giorgiosironi: actually it does setup something at the end of the method... ignore this comment.

src/buildercore/bootstrap.py
```diff
@@ -162,6 +166,18 @@ def is_resourcing():
 
     stack_all_ec2_nodes(stackname, _setup_ec2_node, username=BOOTSTRAP_USER)
 
+def setup_terraform(stackname, context):
+    if not context.get('fastly'):
```
giorgiosironi: these are feature flags that prevent any stack not using Fastly from going through this unstable code for now

src/buildercore/bootstrap.py
```diff
@@ -162,6 +166,18 @@ def is_resourcing():
...
+        return
+    if context.get('fastly'):
+        if 'FASTLY_API_KEY' not in os.environ:
```
giorgiosironi: this is configured out-of-band like AWS credentials, so should be mostly transparent for builder.

src/buildercore/bootstrap.py
```diff
@@ -162,6 +166,18 @@ def is_resourcing():
...
+            raise ConfigurationException("You must provide a FASTLY_API_KEY environment variable to provision Fastly resources")
+    
+    working_dir = join(TERRAFORM_DIR, stackname)
+    t = Terraform(working_dir=working_dir)
```
giorgiosironi: logging what is happening needs work. Now I think it's just printing it out

src/buildercore/bootstrap.py
```diff
@@ -256,6 +272,14 @@ def update_sqs_stack(stackname, **kwargs):
         # stack has/had queues that may need to be created/destroyed
         setup_sqs(stackname, current_context_sqs, current_context['project']['aws']['region'])
 
+def update_terraform_stack(stackname, **kwargs):
+    if 'FASTLY_API_KEY' not in os.environ:
+        raise ConfigurationException("You must provide a FASTLY_API_KEY environment variable to provision Fastly resources")
+    
+    working_dir = join(TERRAFORM_DIR, stackname)
+    t = Terraform(working_dir=working_dir)
+    t.apply(input=False, capture_output=False, raise_on_error=True)
```
giorgiosironi: `apply()` calls require a confirmation from the user typing `yes`. There is a lot of potential around using `plan` and `apply` in two separate phases, but it's still unclear how to integrate that for now.

src/buildercore/bootstrap.py
```diff
@@ -256,6 +272,14 @@ def update_sqs_stack(stackname, **kwargs):
         # stack has/had queues that may need to be created/destroyed
         setup_sqs(stackname, current_context_sqs, current_context['project']['aws']['region'])
 
+def update_terraform_stack(stackname, **kwargs):
+    if 'FASTLY_API_KEY' not in os.environ:
+        raise ConfigurationException("You must provide a FASTLY_API_KEY environment variable to provision Fastly resources")
+    
+    working_dir = join(TERRAFORM_DIR, stackname)
```
giorgiosironi: every stack has its own `.cfn/terraform/api-gateway--prod/` directory for maximum isolation. May require some duplication of values and/or some basic stacks like `bus` to be created in general, but avoids a `terraform destroy` on a testing stack accidentally taking down production

giorgiosironi: Also, everything in that directory is local. It should use a remote storage like the `elife-builder` S3 bucket to share what has been created between builder users (even Alfred)

src/buildercore/bootstrap.py
```diff
@@ -457,7 +481,8 @@ def update_stack(stackname, service_list=None, **kwargs):
     service_update_fns = OrderedDict([
         ('ec2', update_ec2_stack),
         ('s3', update_s3_stack),
-        ('sqs', update_sqs_stack)
+        ('sqs', update_sqs_stack),
+        ('fastly', update_terraform_stack)
```
giorgiosironi: This can't be `('fastly', update_fastly_stack)` because if we had Fastly and, say, Github managed by Terraform we would still have to `apply` them together. Or maybe there's a way to separate them. In any case, not an hard decision until we have some future additional service there.

src/buildercore/cfngen.py
```diff
@@ -237,6 +238,18 @@ def build_subdomain(x):
...
         context['cloudfront'] = False
 
+def build_context_fastly(context, parameterize):
+    def build_subdomain(x):
+        return complete_domain(parameterize(x), context['domain'])
+    if 'fastly' in context['project']['aws']:
+        context['fastly'] = {
+            'subdomains': [build_subdomain(x) for x in context['project']['aws']['fastly']['subdomains']],
+            # future use
+            'subdomains-without-dns': [],
```
giorgiosironi: probably lots of stuff will appear here, with Fastly terminology. We don't do a translation from abstract terms, e.g. everything under `cloudfront` is pretty much taken from CloudFront/CloudFormation documentation. I expect `fastly` to be similar, otherwise the user trying to configure it has to learn yet another terminology on top of the CloudFront/Fastly one. Portability of the same configuration across cloud providers or vendors is out of scope, don't think even Terraform does that as AWS stuff is specified through the [aws provider](https://www.terraform.io/docs/providers/aws/).

`subdomains` is a rare example of an abstract key.

src/buildercore/cfngen.py
```diff
@@ -279,6 +292,15 @@ def write_template(stackname, contents):
     open(output_fname, 'w').write(contents)
     return output_fname
 
+def write_terraform_template(stackname, contents):
+    "optionally, store a terraform configuration file for the stack"
+    if contents:
```
giorgiosironi: `if` allows other stacks not to be affected by what is going on here, making the Terraform files effectively optional

src/buildercore/cfngen.py
```diff
@@ -350,34 +372,46 @@ def generate_stack(pname, **more_context):
     """given a project name and any context overrides, generates a Cloudformation
     stack file, writes it to file and returns a pair of (context, stackfilename)"""
     context = build_context(pname, **more_context)
-    template = render_template(context)
+    cloudformation_template = render_template(context)
```
giorgiosironi: can see the asymmetry here: should be `cloudformation_template = cloudformation.render(context)`. I plan to extract a `cloudformation` module from this, as `cfngen` has grown with more than just those specific concerns. This should become a `context generator` rather than a `cloudformation generator` if that was the original meaning (but I am *very* hesitant to rename modules, have to live with it for a while).

src/buildercore/cfngen.py
```diff
@@ -350,34 +372,46 @@ def generate_stack(pname, **more_context):
...
+    terraform_template = terraform.render(context)
     stackname = context['stackname']
-    out_fname = write_template(stackname, template)
+
     context_handler.write_context(stackname, context)
-    return context, out_fname
+    cloudformation_template_file = write_template(stackname, cloudformation_template)
+    terraform_template_file = write_terraform_template(stackname, terraform_template)
+    return context, cloudformation_template_file, terraform_template_file
```
giorgiosironi: should become `context, data_structure` where `data_structure` could be a list of two elements or any way to keep these files together. `namedtuple` maybe

src/buildercore/cfngen.py
```diff
@@ -350,34 +372,46 @@ def generate_stack(pname, **more_context):
...
 # update existing template
 #
 
 
 # can't add ExtDNS: it changes dynamically when we start/stop instances and should not be touched after creation
-UPDATABLE_TITLE_PATTERNS = ['^CloudFront.*', '^ElasticLoadBalancer.*', '^EC2Instance.*', '.*Bucket$', '.*BucketPolicy', '^StackSecurityGroup$', '^ELBSecurityGroup$', '^CnameDNS.+$', '^AttachedDB$', '^AttachedDBSubnet$', '^ExtraStorage.+$', '^MountPoint.+$', '^IntDNS.*$', '^ElastiCache.*$']
+UPDATABLE_TITLE_PATTERNS = ['^CloudFront.*', '^ElasticLoadBalancer.*', '^EC2Instance.*', '.*Bucket$', '.*BucketPolicy', '^StackSecurityGroup$', '^ELBSecurityGroup$', '^CnameDNS.+$', 'FastlyDNS\\d+$', '^AttachedDB$', '^AttachedDBSubnet$', '^ExtraStorage.+$', '^MountPoint.+$', '^IntDNS.*$', '^ElastiCache.*$']
 
-REMOVABLE_TITLE_PATTERNS = ['^CnameDNS\\d+$', '^ExtDNS$', '^ExtraStorage.+$', '^MountPoint.+$', '^.+Queue$', '^EC2Instance.+$', '^IntDNS.*$', '^ElastiCache.*$', '^.+Topic$']
+REMOVABLE_TITLE_PATTERNS = ['^CnameDNS\\d+$', 'FastlyDNS\\d+$', '^ExtDNS$', '^ExtraStorage.+$', '^MountPoint.+$', '^.+Queue$', '^EC2Instance.+$', '^IntDNS.*$', '^ElastiCache.*$', '^.+Topic$']
```
giorgiosironi: Just the usual whitelisting saying we can update the CNAME DNS entries of Fastly that are managed through CloudFormation.

src/buildercore/cfngen.py
```diff
@@ -350,34 +372,46 @@ def generate_stack(pname, **more_context):
...
 EC2_NOT_UPDATABLE_PROPERTIES = ['ImageId', 'Tags', 'UserData']
 
-class Delta(namedtuple('Delta', ['plus', 'edit', 'minus'])):
+class Delta(namedtuple('Delta', ['plus', 'edit', 'minus', 'terraform'])):
```
giorgiosironi: CloudFormation is nicely chopped up into:

- what to add
- what to modify
- what to remove

What we see here is the new Terraform `generated.tf` file, containing all resources (just Fastly so far). We can do a diff with the current one which would already be an improvement, but ultimately the source of truth is changing it and running a `terraform plan` to see proposed changes. We should however roll it back if the user doesn't confirm, so this will need some machinery (I would prefer never to modify it in the first place if possible, don't know enough about Terraform workflow yet.)

src/buildercore/terraform.py
```diff
@@ -0,0 +1,31 @@
...
+                    'name': context['stackname'],
+                    'domain': {
+                        'name': context['fastly']['subdomains'][0],
+                    },
+                    'backend': {
+                        'address': context['full_hostname'],
+                        'name': context['stackname'],
+                        'port': 443,
+                        'use_ssl': True,
+                        'ssl_check_cert': False
```
giorgiosironi: bad option, but it's for minimal fuss. Before we start customizing this, a lot of the risk to be tackled is integrating everything together with a good lifecycle for adding, modifying and removing CDNs that point to CloudFormation-managed resources.

src/buildercore/trop.py
```diff
@@ -49,6 +51,8 @@
 
 KEYPAIR = "KeyName"
 
+FASTLY_GLOBAL = 'nonssl.global.fastly.net.'
```
giorgiosironi: The other options is a hostname for just EU/US edge nodes, which costs less but has worse performance outside of EU/US.

src/buildercore/trop.py
```diff
@@ -377,6 +381,24 @@ def external_dns_cloudfront(context):
...
+def external_dns_fastly(context):
+    "additional CNAME DNS entries pointing to a Fastly CDN"
+    assert isinstance(context['domain'], str), "A 'domain' must be specified for CNAMEs to be built"
+
+    def entry(hostname, i):
+        if _is_domain_2nd_level(hostname):
+            raise ConfigurationError("2nd-level domains aliases are not supported yet by builder. See https://docs.fastly.com/guides/basic-configuration/using-fastly-with-apex-domains")
+        hostedzone = context['domain'] + "."
+        return route53.RecordSetType(
+            R53_FASTLY_TITLE % (i + 1),
```
giorgiosironi: expecting that there can be more than one entry (aliases), so numbering them immediately

src/buildercore/trop.py
```diff
@@ -377,6 +381,24 @@ def external_dns_cloudfront(context):
 
     return dns_records
 
+def external_dns_fastly(context):
+    "additional CNAME DNS entries pointing to a Fastly CDN"
+    assert isinstance(context['domain'], str), "A 'domain' must be specified for CNAMEs to be built"
+
+    def entry(hostname, i):
+        if _is_domain_2nd_level(hostname):
```
giorgiosironi: requires some contract tweak that we'll eventually need to do

src/tests/fixtures/projects/dummy-project.yaml
```diff
@@ -294,6 +294,16 @@ project-with-cloudfront-origins:
                     cookies:
                         - session_id
 
+project-with-fastly-minimal:
```
giorgiosironi: we can test up to the template generation with this, which will be useful for all kind of customizations

src/buildercore/bootstrap.py
giorgiosironi: Also, right now all these are executed during deploy. We should move the default to only update EC2 during deploy as SQS/S3/Fastly/anything else should not be modified during deploy but with `update_template` (to be renamed, accepting suggestions?) by a human operator.

src/cfn.py
```diff
@@ -82,6 +83,7 @@ def update_template(stackname):
     LOG.info("Create: %s", pformat(delta.plus))
     LOG.info("Update: %s", pformat(delta.edit))
     LOG.info("Delete: %s", pformat(delta.minus))
+    LOG.info("New Terraform generated file: %s", pformat(json.loads(delta.terraform)))
```
giorgiosironi: `update_template` is not the right name anymore as the process is integrating Terraform with its own resource files, not just CloudFormation templates. It was already expanded to also run Salt, and other AWS API calls for S3 and SQS. 

Possible names:

- `reprovision`
- `apply` (of Terraform fame)
- `re_source`
- `update_...` (update what?)
- `source_again` (crappy)

It's important to render this is a human operation, not to be performed at every deploy.

lsh-0:
just a quick comment off the bat. I appreciate you annotating the code through github, but it's making the diff difficult to read. With big rough changes like this I'd prefer blocks of comments above the code itself with explanations. Review-type feedback can then be annotated via github with less noise. What do you think?

src/buildercore/bootstrap.py
lsh-0: `.prerequisites.py` candidate then?

src/buildercore/cfngen.py
lsh-0: > Portability of the same configuration across cloud providers or vendors is out of scope,

:+1:

In the beginning with cloudformation I kept project configuration abstract and translated options into cfn features collected together in cfngen.py which was then used to render a simple template with python formatting strings. then jinja and then finally troposphere was introduced to generate the template. It did mean that while the project property you were changing was easy to do, tracing it from project config into template meant traversing through a lot of code. Gathering this config under specific headings like `fastly` or `rds` rather than `cdn` or `remote-database` means we can skip the code traversal (probably) and match the options to the docs. The next step might be a simple pass-through of options in project config into the final template.

src/buildercore/cfngen.py
lsh-0: it's fine. `render_template` is just calling trop.py, so we could rename trop.py to cloudformation.py 

if you look at `render_template` it looks like you or I built in support for other template types in the future, but we could simply shift that function into the new cloudformation.py module and have a matching one in terraform.py. or use `render_template` to dispatch calls to the other modules, I don't mind.

src/buildercore/cfngen.py
lsh-0: eh, namedtuples have their place but it isn't here, use a plain old dict

lsh-0: ... do we see ourselves writing a third provider? I think if anything I'd rather write a provider for terraform and access it that way. in that case, I'd just leave it as a triple,

src/cfn.py
lsh-0: `update_infrastructure`

it sounds necessarily big and scary

src/tests/fixtures/projects/dummy-project.yaml
lsh-0: can we do any sort of dry run or validation like we can with Cloudformation (no matter how sucky it is) ?

lsh-0:
alright, I've been through all of the changes. 

* I think you should rename trop.py
* I'm still kind of lost when it comes to your delta wrangling when doing updates. I'm not sure how I would do it differently.
* I added some of your github annotations into the code
* `update_template` definitely needs to change, that area is already a bit muddy with `update` `update_master` and `highstate`.

credentials management is getting messy as well. I wonder if we should the credentials check into a decorator:

```python
@credentials('fastly')
def update_terraform(...):
    ...

@credentials('aws')
def client_conn(service, ...):
    ...
```

perhaps leave off until a third set of secrets is required

lsh-0:
going to hold off on explicit approval until the `[wip]` disappears

thewilkybarkid:
> just a quick comment off the bat. I appreciate you annotating the code through github, but it's making the diff difficult to read. With big rough changes like this I'd prefer blocks of comments above the code itself with explanations. Review-type feedback can then be annotated via github with less noise. What do you think?

If you hide whitespace changes (add `?w=1` to the URL) it also hides comments.

lsh-0:
that's good to know, thank you.

I can't see an option for that anywhere, how did you come by this piece of arcane knowledge?

thewilkybarkid:
No idea! Don't think it is available through the UI, which is weird because it's pretty useful.

giorgiosironi:
> Review-type feedback can then be annotated via github with less noise. What do you think?

We could also use `# TODO:` comments which are durable and I usually filter with `git diff origin/master...HEAD | grep TODO` to make sure everything has been addressed before merge.

src/tests/fixtures/projects/dummy-project.yaml
giorgiosironi: `terraform plan` does some sort of validation, but interacts with real infrastructure so it's a no-go for tests (and will eventually be executed as part of the process anyway).
https://www.terraform.io/docs/commands/validate.html does not do that much, but it's a start.

src/buildercore/cfngen.py
giorgiosironi: One important thing that happens here is interpolation of strings e.g. `pattern-library-{instance}` bucket name being interpolated to `pattern-library-ci`

giorgiosironi:
Done a general cleanup, more module shifting will follow in the next PR but this is ready for final review and merge.

src/buildercore/cloudformation.py
```diff
@@ -0,0 +1,8 @@
+from . import trop
```
lsh-0: shifting contents of trop.py to cloudformation.py will come later?

PR #263, opened 2018-04-05
Feature: py2 support

lsh-0:
there are performance problems with python3 + fabric/paramiko causing builds to take a very very long time to complete.

this introduces some changes that makes builder python2 and python3 compatible while performance issues are sorted.

tests are passing on python2 and python3


giorgiosironi:
I am testing various tasks, but the main issue so far is that `Fabric` (and probably the old `paramiko`?) should be used in place of `Fabric3` when Python 2 is chosen; it's not the language version that causes slowdowns but we had suspicious about that part of the stack.

giorgiosironi:
Have to add `tox` to Alfred.

giorgiosironi:
Finally green again and running tests on both versions of Python and Fabric

Tears of joy

src/tests/test_buildercore_utils.py
```diff
@@ -39,6 +39,10 @@ def test_isint(self):
         ]
         self.assertAllTrue(utils.isint, expected_true)
 
+    def test_isstr(self):
+        expected_true = [''] if utils.gtpy2() else ['', r'', u'']
```
giorgiosironi: this was also passing `b''`, but that clearly fails in Python 3 not being an instance of `str`

src/tests/base.py
```diff
@@ -2,11 +2,12 @@
 import os
 from os.path import join
 from subprocess import check_output
-from unittest import TestCase
+# pylint: disable-msg=import-error
+from unittest2 import TestCase
```
giorgiosironi: got confused by this, but the most consistent choice seems to use `unittest2` (the backport of unittest for Python 2) for all versions, so that the methods like `assertRegex` are always perfectly the same

src/tests/test_buildercore_utils.py
lsh-0: weird - it was all passing in python2 and 3. replicating now I can see it's not.

src/tests/base.py
lsh-0: yeah, unittest2 was a little old and I think it missed assertCountEquals that I had to monkey patch, but it fixed a host of other smaller problems

PR #265, opened 2018-04-06
boto2 removal

lsh-0:
* s3
* cloudformation
* ec2


giorgiosironi:
Good idea, tag me when ready for review

giorgiosironi:
Suggest to solve conflicts and clean up the S3 part, or even part of the S3 part, to become ready for merge. Pull requests that live longer than a few days are :warning: signs nowadays

lsh-0:
> Pull requests that live longer than a few days are warning signs nowadays

why?

giorgiosironi:
A short and simple summary of unfinished work as overhead https://www.agilealliance.org/what-is-waste/

src/buildercore/s3.py
```diff
@@ -1,46 +1,59 @@
...
+    except ClientError as err:
+        LOG.error("unhandled error attempting to find S3 bucket %r in region %r", nom, region,
+                  extra={'bucket': nom, 'region': region, 'error': err.message})
         raise
 
 def exists(key):
-    return builder_bucket().get_key(key) is not None
+    "predicate, returns True if given key in configured bucket+region exists"
+    try:
+        builder_bucket().Object(key).load()
```
giorgiosironi: I thought it could be checked without actually loading the value of the key (like an underlying `HEAD` request) but for this use case it shouldn't be a problem

giorgiosironi: consider also https://stackoverflow.com/a/38376288/91590 if you haven't already excluded it

lsh-0: already considered it ;) see docs: http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Object.load

lsh-0:
this is still a `wip`, but I'll outline how I'm tackling this, because it requires a lot of delicate work.

buildercore.core has a bunch of connection accessors and convenience functions that are mixing boto2 and boto3 and within boto3 there is mixing of 'client' vs 'resource' abstractions.

I'm reducing these accessors down to just three: `boto_conn`, `boto_client` and `boto_resource`.

boto3 client code looks very similar to boto2 code but still may behave differently to it's boto2 counterpart even if two methods are named the same.

I was planning on switching as much as I could to the higher level boto3 'resource' abstraction, but after some more work on it today (after completing the s3 module previously) I think this will take too long and I'll just take the path of least resistance so long as it's boto3.

I've encountered several dead functions that I've removed and some old fabric tasks that could be stripped out.

Tests and linting are passing locally but I'm a little distrustful and I hope to write some tests as this chunk of work nears the end.

giorgiosironi:
> Tests and linting are passing locally but I'm a little distrustful and I hope to write some tests as this chunk of work nears the end.

Ok, good idea, I'll wait for the un-wip of this PR and then review.

lsh-0:
the branch was only getting old because you poked the salt upgrade branch and I switched to that ;)

I don't think managing workload via git instead of jira is the right way to go about things (not to imply you did this intentionally, but it's what I did). Questioning branches/PRs that are getting long in the tooth is fair.

lsh-0:
once you bend your head around the quirks of the boto3 docs it's actually not so bad. the boto2 docs are quite unfriendly.

giorgiosironi:
> the branch was only getting old because you poked the salt upgrade branch and I switched to that ;)

Not saying we shouldn't have more than one theme running at a time, as there are physiological delays involved (review from the other side of the world, feedback from builds and production deploys that may happen down the line). However, breaking down each theme into [many pull requests that get merged](https://trunkbaseddevelopment.com/short-lived-feature-branches/) sooner avoids conflicts and divergence, letting people enjoy value sooner (in this case possible performance, robustness or maintenance improvements of Boto3+S3).

lsh-0:
I'm going to stop with any more code changes in this branch and just work on getting the tests to pass now, it's large enough as it is.

lsh-0:
I have the (regular) tests passing locally and my own 'integration' tests that borrowed liberally from yours, @giorgiosironi. It's something I've wanted to do for a long time and your own integration tests module was good enough for a long time. Nothing too special, it brings up an instance and after the tests have finished running it brings it down again, but I've added options to let me persist the instance so I can go in and debug. It's been really helpful in ironing out these last few bugs.

src/integration_tests/test_with_instance.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # to re-use an existing stack, ensure self.reuse_existing_stack is True
+        # this will read the instance name from a temporary file (if it exists) and
+        # look for that, creating it if doesn't exist yet
+        # also ensure self.cleanup is False so the instance isn't destroyed after tests complete
+        self.reuse_existing_stack = False
+        self.cleanup = True
+
+        self.stacknames = []
+        self.environment = base.generate_environment_name()
```
giorgiosironi: this is key to allow isolation of the test suite and having multiple builds running simultaneously: each `stackname` has a random suffix and different test suites don't see each other

lsh-0: I get that - I don't think this undermines that. `generate_environment_name` had to be shifted out of the class is all.

giorgiosironi: Was a positive comment for anyone reading here in the future

lsh-0: ah - sorry, I was a bit unsure. I'm also feeling slightly defensive as I prepare to take responsibility for builder breaking again so soon after the salt upgrade. let's hope it doesn't break ;) the last of the bugs are being ironed out and the tests are doing a good job of unearthing things. I didn't realize just how deeply boto2-isms had infiltrated the code.

giorgiosironi: I have a feeling of why  :grinning:

src/buildercore/cloudformation.py
```diff
@@ -15,8 +15,8 @@ def render_template(context):
 def _log_backoff(event):
     LOG.warn("Backing off in validating project %s", event['args'][0])
 
-@backoff.on_exception(backoff.expo, boto.connection.BotoServerError, on_backoff=_log_backoff, max_time=30)
+@backoff.on_exception(backoff.expo, botocore.exceptions.ClientError, on_backoff=_log_backoff, max_time=30)
```
giorgiosironi: will check this is what comes out on throttling

src/buildercore/bootstrap.py
```diff
@@ -77,22 +78,27 @@ def create_stack(stackname):
...
+        parameters.append({'ParameterKey': 'KeyName', 'ParameterValue': stackname})
         on_start = lambda: keypair.create_keypair(stackname)
         on_error = lambda: keypair.delete_keypair(stackname)
 
     return _create_generic_stack(stackname, parameters, on_start, on_error)
 
 def _create_generic_stack(stackname, parameters=None, on_start=_noop, on_error=_noop):
     "transforms templates stored on the filesystem into real resources (AWS via CloudFormation, Terraform)"
+    if core.stack_is_active(stackname):
+        LOG.info("stack exists") # avoid on_start handler
```
giorgiosironi: I assumed the code shouldn't get here if the stack already exists?

giorgiosironi: But idempotence is welcome

src/buildercore/bootstrap.py
```diff
@@ -104,7 +110,9 @@ def _create_generic_stack(stackname, parameters=None, on_start=_noop, on_error=_
         LOG.info("Stack taking a long time to complete: %s", err)
         raise
 
-    except BotoServerError as err:
+    except botocore.exceptions.ClientError as err:
+        # TODO: is this ' already exists' check different from the other 'does not exist' check?
+        # if so, can we get a list of these messages somewhere?
         if err.message.endswith(' already exists'):
```
giorgiosironi: this has changed to:
```
>>> conn.create_stack(StackName='medium--ci', TemplateBody='{}')
INFO - boto3.resources.action - Calling cloudformation:create_stack with {'StackName': 'medium--ci', 'TemplateBody': '{}'}
2018-04-20 16:09:59,376 botocore.vendored.requests.packages.urllib3.connectionpool [INFO] Starting new HTTPS connection (1): cloudformation.us-east-1.amazonaws.com
INFO - botocore.vendored.requests.packages.urllib3.connectionpool - Starting new HTTPS connection (1): cloudformation.us-east-1.amazonaws.com
  File "<stdin>", line 1, in <module>
  File "/home/giorgio/code/builder/venv/local/lib/python2.7/site-packages/boto3/resources/factory.py", line 520, in do_action
    response = action(self, *args, **kwargs)
  File "/home/giorgio/code/builder/venv/local/lib/python2.7/site-packages/boto3/resources/action.py", line 83, in __call__
    response = getattr(parent.meta.client, operation_name)(**params)
  File "/home/giorgio/code/builder/venv/local/lib/python2.7/site-packages/botocore/client.py", line 310, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/giorgio/code/builder/venv/local/lib/python2.7/site-packages/botocore/client.py", line 599, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.AlreadyExistsException: An error occurred (AlreadyExistsException) when calling the CreateStack operation: Stack [medium--ci] already exists
```

giorgiosironi: so it should be easier to catch

src/buildercore/bootstrap.py
```diff
@@ -135,9 +143,11 @@ def is_updating(stackname):
         update_msg='Waiting for AWS to finish creating stack ...',
         exception_class=StackTakingALongTimeToComplete
     )
+
     final_stack = core.describe_stack(stackname)
-    events = [(e.resource_status, e.resource_status_reason) for e in final_stack.describe_events()]
-    ensure(final_stack.stack_status in ['CREATE_COMPLETE'],
+    # NOTE: stack.events.all|filter|limit can take 5+ seconds to complete regardless of events returned
```
giorgiosironi: I think it goes through soem pagination

src/buildercore/bootstrap.py
```diff
@@ -370,55 +379,63 @@ def _sqs_notification_configuration(queue_arn, notification_specification):
...
+        parameters.append({'ParameterKey': 'KeyName', 'ParameterValue': stackname})
     try:
-        conn.update_stack(stackname, json.dumps(template), parameters=parameters)
-    except BotoServerError as ex:
-        if ex.message == 'No updates are to be performed.':
+        conn = core.describe_stack(stackname)
+        conn.update(TemplateBody=json.dumps(template), Parameters=parameters)
+    except botocore.exceptions.ClientError as ex:
+        # ex.response ll: {'ResponseMetadata': {'RetryAttempts': 0, 'HTTPStatusCode': 400, 'RequestId': 'dc28fd8f-4456-11e8-8851-d9346a742012', 'HTTPHeaders': {'x-amzn-requestid': 'dc28fd8f-4456-11e8-8851-d9346a742012', 'date': 'Fri, 20 Apr 2018 04:54:08 GMT', 'content-length': '288', 'content-type': 'text/xml', 'connection': 'close'}}, 'Error': {'Message': 'No updates are to be performed.', 'Code': 'ValidationError', 'Type': 'Sender'}}
+        if ex.response['Error']['Message'] == 'No updates are to be performed.':
```
giorgiosironi: `LOG.info` would be helpful here

src/buildercore/bootstrap.py
```diff
@@ -370,55 +379,63 @@ def _sqs_notification_configuration(queue_arn, notification_specification):
...
 def template_info(stackname):
     "returns some useful information about the given stackname as a map"
-    data = core.describe_stack(stackname).__dict__
-    data['outputs'] = reduce(utils.conj, map(lambda o: {o.key: o.value}, data['outputs']))
-    return utils.exsubdict(data, ['connection', 'parameters'])
+    # data = core.describe_stack(stackname).__dict__ # original boto2 approach, never officially supported
+    data = core.describe_stack(stackname).meta.data # boto3
+
+    # looking at the entire set of formulas, all usage is cfn.outputs, cfn.stack_id and cfn.stack_name
+    # in the interests of being explicit on what is supported, I'm changing what this now returns
```
giorgiosironi: oh yeah :+1:

src/buildercore/bootstrap.py
```diff
@@ -654,21 +673,22 @@ def delete_stack(stackname):
...
-        # remove_minion_key(stackname) # don't do this. prevents regular users deleting stacks
+        # don't do this. requires master server access and would prevent regular users deleting stacks
+        # remove_minion_key(stackname)
         keypair.delete_keypair(stackname) # deletes the keypair wherever it can find it (locally, remotely)
         delete_stack_file(stackname) # deletes the local cloudformation template
         delete_dns(stackname)
-
         LOG.info("stack %r deleted", stackname)
-    except BotoServerError as err:
+
```
giorgiosironi: it's interesting how the errors turned from server faults to client faults in a new version :)

src/integration_tests/test_with_instance.py
```diff
@@ -0,0 +1,128 @@
...
+class One(base.BaseCase):
+    @classmethod
+    def setUpClass(self): # cls, not self
+        super(One, self).setUpClass()
+
+        # to re-use an existing stack, ensure self.reuse_existing_stack is True
+        # this will read the instance name from a temporary file (if it exists) and
+        # look for that, creating it if doesn't exist yet
+        # also ensure self.cleanup is False so the instance isn't destroyed after tests complete
+        self.reuse_existing_stack = False
```
giorgiosironi: if this read an environment variable it could be set for local usage without changing the code

src/integration_tests/test_with_instance.py
```diff
@@ -0,0 +1,128 @@
...
+        self.cleanup = True
+
+        self.stacknames = []
+        self.environment = base.generate_environment_name()
+        # self.temp_dir, self.rm_temp_dir = utils.tempdir()
+
+        # debugging only, where we keep an instance up between processes
+        self.state, self.statefile = {}, '/tmp/.open-test-instances.txt'
```
giorgiosironi: this files is not written outside of local usage, so should keep `ci` isolated :+1:

src/integration_tests/test_with_instance.py
```diff
@@ -0,0 +1,128 @@
...
+        self.assertEqual(len(core.find_ec2_instances(self.stackname)), 1) # 1 node is running
+        self.assertEqual(len(core.find_ec2_instances(self.stackname, state='stopped', allow_empty=True)), 0) # 0 nodes are stopped
+
+    def test_bootstrap_template_info(self):
+        resp = bootstrap.template_info(self.stackname)
+        expecting = {
+            'stack_name': self.stackname,
+            'stack_id': 'arn:aws:cloudformation:us-east-1:512686554592:stack/dummy1--luke-20180420025657-458624/afc10200-4459-11e8-8c52-500c2864aa35',
+            'outputs': {
+                'PublicIP1': '34.229.140.136',
```
giorgiosironi: ? this doesn't look stable between builds? But could still check the keys and a regex on the values

src/integration_tests/test_with_instance.py
```diff
@@ -0,0 +1,128 @@
...
+                'AZ1': 'us-east-1d',
+                'InstanceId1': 'i-027c0ae8fd9054cff',
+                'PrivateIP1': '10.0.2.237'
+            }
+        }
```
giorgiosironi: ah, this is the example

src/integration_tests/test_with_instance.py
```diff
@@ -0,0 +1,128 @@
...
+    def test_bootstrap_write_environment_info(self):
+        with core.stack_conn(self.stackname, username=BOOTSTRAP_USER):
+            bootstrap.write_environment_info(self.stackname, overwrite=False)
+            bootstrap.write_environment_info(self.stackname, overwrite=True)
+
+    def test_delete_stack_file_while_still_active(self):
+        self.assertEqual(None, bootstrap.delete_stack_file(self.stackname))
+
+    def test_core_describe_stack(self):
+        core.describe_stack(self.stackname)
```
giorgiosironi: all these calls are very valuable as it would very difficult to exercise that code with stubs/mocks

src/integration_tests/test_with_instance.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def test_core_stack_is(self):
+        self.assertTrue(core.stack_is_active(self.stackname))
+        self.assertFalse(core.stack_is(self.stackname, ['fubar']))
+        self.assertRaises(RuntimeError, core.stack_is, self.stackname, ['fubar'], core.ACTIVE_CFN_STATUS)
+
+    def test_core_active_aws_stacks(self):
+        self.assertTrue(len(core.active_aws_stacks(self.region)) >= 1)
```
giorgiosironi: these assertions are a bit weak, we could check for the presence of the stack in the list instead

src/tasks.py
```diff
@@ -189,8 +60,3 @@ def repair_context(stackname):
 @requires_aws_stack
 def remove_minion_key(stackname):
     bootstrap.remove_minion_key(stackname)
-
```
giorgiosironi: :+1: on deletions

src/integration_tests/test_with_instance.py
lsh-0: heh.

src/buildercore/bootstrap.py
lsh-0: it does actually, the only thing stopping it is a check in `cfn.launch`. without it, the `on_start` succeeds the first time in creating the keypair but fails the next time with a 'file exists' type error.

src/buildercore/bootstrap.py
lsh-0: yeah, I was testing on new stacks and old stacks and regardless of the number of the number of events or how I limited the number returned and returned per-page it was a very slow call. Thought I should point it out.

src/buildercore/bootstrap.py
lsh-0: I hadn't noticed, but yeah, now that you point it out ... And there is so little documentation around botocore and it's exceptions.

src/integration_tests/test_with_instance.py
lsh-0: agreed, I'll do that

lsh-0:
thanks for your feedback, @giorgiosironi . I've just gone through all of my changes after a weekend off and I'm glad to see the changes are uncontroversial. Build is still failing but it looks to be a `NetworkError` rather than a code error. I'll tweak the few things you mentioned and get this build passing.

What remains is sns and sqs and these should be much more discrete after laying the groundwork here.

src/integration_tests/test_with_instance.py
lsh-0: I've been looking at a friend's codebase on my Thursdays recently, trying to get him on board the CI train but he was complaining about tests being hard to do and slow etc etc. I have his tests working fast now but debugging his work he's doing something similar to what I'm doing here: setting up a single huge scenario and running assertions against it. Unfortunately this approach sees the tests mutate the scenario's state so how deterministic they can be I don't know, but honestly - some tests are far better than no tests at all.

src/buildercore/bootstrap.py
lsh-0: it has some issues: https://github.com/boto/boto3/issues/1195

it says to use botocore.exceptions but it doesn't exist there either. It is bypassing the adhoc 'exists' check on the message string rather than returning false, so good catch

src/integration_tests/test_with_instance.py
lsh-0: looking at the output of active_aws_stacks, this test is especially weak and embarrassing. for some reason I must have assumed the test stack was the only one being brought up

lsh-0:
very interesting - there is a difference in the cloudformation template generated by python2 vs one generated by python3. after switching to a py3 environment to replicate these alfred errors, the test is causing an update to occur

lsh-0:
my new test suite will be run with the integration tests, but you can now also run it like this:

`BLDR_TWI_REUSE_STACK=1 BLDR_TWI_CLEANUP=0 ./.run-tests.sh integration_tests.test_with_instance`

the above will preserve a stack to test against and skip tearing it down afterwards. When you're done, just change `BLDR_TWI_CLEANUP=0` to `1`.

lsh-0:
it's green it's green it's green it's green :tada:

src/integration_tests/test_with_instance.py
giorgiosironi: isolation is in a trade off with speed here, unless tests can be perfectly run in parallel which would both isolate them onto different stacks and allow the slowest test to be the bottleneck rather than summing the execution times of them all

giorgiosironi:
Plan to merge at 15:00 UTC after we have finished with Fastly changes for the day.

lsh-0:
cheers, @giorgiosironi - I know you've been working on builder too + fastly and I didn't want to step on anyone's toes

PR #267, opened 2018-04-09
ELPP-3448 Pin master server in build vars

giorgiosironi:
`update_template` (soon to be renamed in `update_infrastructure`) pins the new master ip, which then is preferred to the one passed from builder.

`update` or other deploys don't change the build vars instead.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -45,6 +45,8 @@ defaults:
             # ami: ami-92002785   # elife 'basebox.2016-11-03'
             # use a master server or go ronin?
             masterless: false
+            # optional: pin a master server for all new instances of a project
+            master_ip: 10.0.2.198
         region: us-east-1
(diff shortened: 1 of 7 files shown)


lsh-0:
I can see what you've done and approve, this approach will let us gradually migrate master servers. If we go down this route we should give the master server an elastic IP or the minions a domain name instead of an IP. I hadn't thought of it before, but if the master server is restarted and comes up with a different IP, like it would right now, all of the minions would need their config updated.

I'm going to approve this, but will look into providing the minion with an internal domain name for the master server while debugging the crm formula.

giorgiosironi:
>  if the master server is restarted and comes up with a different IP

This is a private ip so it should be preserved during reboots (it's public IPs that are scarce)

> will look into providing the minion with an internal domain name for the master server

But an internal domain name will be useful, because when you look at a minion configuration you see `2017-...elife.internal` which gives more information than an anonymous ip address.

lsh-0:
ah, didn't realize the private IP was preserved. good to know

PR #270, opened 2018-04-10
bugfix for erronous terraform check that happens during cfn.launch

lsh-0:
quick fix for now, but I would make all these update_* fns consistent in their signature and checks. Probably add a decorator to inject the context if it isn't provided and skip calling the function if the service isn't provided.


lsh-0:
turns out this is actually breaking builds, good to get this in quick

PR #275, opened 2018-04-12
Backoff in CloudFormation validation calls

giorgiosironi:
This integration test often fails with a `BotoServerError` showing a `400 Bad Request` with `Code` `Throttling`. Rather than blindly waiting on every call, we can backoff our calls only when they fail.

### DIFF
--- a/src/cfn.py
+++ b/src/cfn.py
@@ -1,7 +1,9 @@
 from distutils.util import strtobool as _strtobool  # pylint: disable=import-error,no-name-in-module
 import json
 from pprint import pformat
+import backoff
 from fabric.api import task, local, run, sudo, put, get, abort, settings
+import fabric.exceptions
 import fabric.state
 from fabric.contrib import files
 import aws, utils, buildvars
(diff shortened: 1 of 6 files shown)


giorgiosironi:
I'll fix the next `NetworkError: Timed out trying to connect to 54.146.161.208 (tried 1 time)` that has come up.

src/cfn.py
```diff
@@ -313,11 +315,15 @@ def download_file(stackname, path, destination='.', node=None, allow_missing="Fa
     Boolean arguments are expressed as strings as this is the idiomatic way of passing them from the command line.
     """
     allow_missing, use_bootstrap_user = lmap(strtobool, [allow_missing, use_bootstrap_user])
-    with stack_conn(stackname, username=BOOTSTRAP_USER if use_bootstrap_user else DEPLOY_USER, node=node):
-        if allow_missing and not files.exists(path):
-            return # skip download
-        get(path, destination, use_sudo=True)
 
+    @backoff.on_exception(backoff.expo, fabric.exceptions.NetworkError, max_time=60)
```
lsh-0: can't hurt I suppose ...

giorgiosironi: While I was testing this fix, this other error on `download_file` in the `integration_test.test_provisioning` module came up. If a newly created instance doesn't let you access it through SSH yet... never give up, never surrender.

PR #276, opened 2018-04-12
ELPP-3520 Switch api-gateway CDN traffic to Fastly

giorgiosironi:
Already applied to every `api-gateway` where CloudFront was used.

The best pattern to do so seems to be to

1. switch the `subdomains` of `cloudfront` to `subdomains-without-dns` as that is just a DNS deletion rather than a tens-of-minutes CloudFront update. 
1. Switch the `subdomains` of `fastly` from a fake value to the real one, that had been freed by CloudFront.

The process takes a few minutes. If someone still has a cached DNS value, it will end up on CloudFront for the time being, until we retire it (should have zero traffic at that point).

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -338,9 +338,6 @@ api-gateway:
             - 443
             # - 8000: # don't expose this to public. Kong uses this to proxy requests
             # - 8001: # don't ever expose to public. Kong uses this for API admin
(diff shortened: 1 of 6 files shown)


giorgiosironi:
Added the HTTPS certificate check of the backend when accessed by Fastly servers.

PR #278, opened 2018-04-13
Use parallelized elifeLocalTests()

giorgiosironi:
Parallelizes some of the checks and make them independent commit statuses for faster feedback:

Basically everything in `.ci` will be executed in parallel. Paves the way for more of these checks to become parallel, as this part was only related to the Jenkins side.
(too bad `shellcheck` is not present on Alfred so one of this steps gets skipped; we'll install it to allow the check to happen)

Also fixes the execution of linters, which was erroneously skipped when the `tox` environment had to be rolledback to serial rather than parallel.


giorgiosironi:
`continuous-integration/jenkins/static_checking.sh` will become separate checks.

giorgiosironi:
Going to eat my own advice and make this a small, easy to merge PR, to follow up with more work on a new branch.

PR #280, opened 2018-04-17
removes the temporary patch in the highstate script

lsh-0:
that removes the public key from the minion

fyi @giorgiosironi


giorgiosironi:
:+1: 
This operation should be accomplished in the remastering task?

lsh-0:
looks like it is

PR #281, opened 2018-04-18
fixes circular dependency issue with aws.py

lsh-0:
aws.find_region has been shifted into utils.
aws.py is probably a deletion candidate

a tiny peace offering in lieu of the gigantic boto3 port coming up.

### DIFF
--- a/src/aws.py
+++ b/src/aws.py
@@ -1,34 +1,8 @@
 from buildercore import core
 import utils
 from decorators import requires_aws_stack, debugtask, echo_output
-from buildercore.core import MultipleRegionsError
-
 import logging
 LOG = logging.getLogger(__name__)
-
-def find_region(stackname=None):
-    """tries to find the region, but falls back to user input if there are multiple regions available.
-
-    Uses stackname, if provided, to filter the available regions"""
-    try:
-        return core.find_region(stackname)
-    except MultipleRegionsError as e:
-        print("many possible regions found!")
-        return utils._pick('region', e.regions())
-
-def stack_list(region=None):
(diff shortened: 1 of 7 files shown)


src/utils.py
```diff
@@ -118,3 +117,12 @@ def table(rows, keys):
...
     return "\n".join(lines)
+
+def find_region(stackname=None):
+    """tries to find the region, but falls back to user input if there are multiple regions available.
+    Uses stackname, if provided, to filter the available regions"""
+    try:
+        return core.find_region(stackname)
+    except core.MultipleRegionsError as e:
+        print("many possible regions found!")
+        return _pick('region', e.regions())
```
giorgiosironi: feels like the right place to deal with user input

src/aws.py
```diff
@@ -1,35 +1,9 @@
...
-    """tries to find the region, but falls back to user input if there are multiple regions available.
-
-    Uses stackname, if provided, to filter the available regions"""
-    try:
-        return core.find_region(stackname)
-    except MultipleRegionsError as e:
-        print("many possible regions found!")
-        return utils._pick('region', e.regions())
-
-def stack_list(region=None):
```
giorgiosironi: this method was dead code, good to remove it

lsh-0: yeah, always a toss up between marking it as 'dead' or just outright deleting it.

src/aws.py
```diff
@@ -1,35 +1,9 @@
...
-    return core.active_stack_names(region)
-
-#
-#
-#
-
-#
-#
-#
-
```
giorgiosironi: the remaining tasks here do not add much, but seem harmless

PR #282, opened 2018-04-18
Add terraform to prerequisites

giorgiosironi:
Better to install a single binary and not need it that getting a failure halfway through an update of production infrastructure and.


giorgiosironi:
I'll leave this also for @lsh-0 to review since it will start asking for Terraform installation during `./update.sh`.

PR #284, opened 2018-04-18
Allow Alfred to run master.update

giorgiosironi:
In https://alfred.elifesciences.org/job/process-master-server/5/console which was run on-demand before but could become a daily/weekly thing.

### DIFF
--- a/src/master.py
+++ b/src/master.py
@@ -13,7 +13,6 @@ import logging
 
 LOG = logging.getLogger(__name__)
 
-@debugtask
 def update(master_stackname=None):
     "same as `cfn.update` but also removes any orphaned minion keys"
     master_stackname = master_stackname or core.find_master(utils.find_region())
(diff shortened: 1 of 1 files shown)


lsh-0:
don't quite get why you're removing a task decorator here

giorgiosironi:
Ah, it has to be substituted with @task rather than removed. Alfred isn't allowed to run `@debugtask`s.

lsh-0:
ah, no problems.

I wonder if I didn't get too obscure with those decorators. Ian was getting on my case about the output of ./bldr being filled with things he (or anyone looking at builder) shouldn't be using so I made the 'debugtask' for admin type stuff. Alfred can run those jobs with the envvar `BLDR_ROLE=admin`

giorgiosironi:
It is true some things require a savvy human (`update_infrastructure` comes to mind), so I'll leave it incapable of running @debugtask.

PR #285, opened 2018-04-18
Add fastly to iiif, incremental rollout

giorgiosironi:
Will only affect `iiif--end2end`, adding an unused Fastly CDN alongside the current CloudFront one.
First attempt failed because the instance was turned off (unrelated CloudFormation woes), will do it correctly tomorrow with locking.

### DIFF
--- a/src/tests/test_buildercore_terraform.py
+++ b/src/tests/test_buildercore_terraform.py
@@ -1,5 +1,6 @@
 import json  # , yaml
 import os
+import yaml
 from os.path import join
 from . import base
 from buildercore import cfngen, terraform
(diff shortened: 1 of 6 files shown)


src/tests/test_buildercore_terraform.py
```diff
@@ -82,3 +85,7 @@ def test_fastly_template_complex(self):
             },
             data
         )
+
+    def _parse_template(self, terraform_template):
+        """use yaml to load JSON due to https://stackoverflow.com/a/16373377/91590"""
```
lsh-0: due to url? I know what you're getting at, I'd like an actual explanation with the url as reference please

giorgiosironi:
Run into https://github.com/terraform-providers/terraform-provider-fastly/issues/74 while applying this, but the issue is under control.

PR #287, opened 2018-04-23
[WIP] ELPP-3598 Enable Fastly GZIPping

thewilkybarkid:
Need to test whether https://github.com/terraform-providers/terraform-provider-fastly/issues/66 happens.

The API also has lots of custom content types (eg `application/vnd.elife.article-vor+json;version=1`) which won't be included currently, but according https://docs.fastly.com/guides/basic-configuration/enabling-automatic-gzipping we can't use regex here. Ideally anything `+json` would be included.

/cc @giorgiosironi

### DIFF
--- a/src/buildercore/terraform.py
+++ b/src/buildercore/terraform.py
@@ -31,6 +31,9 @@ def render(context):
                         'ssl_cert_hostname': context['full_hostname'],
                         'ssl_check_cert': True,
                     },
+                    'gzip': {
+                        'name': context['stackname'],
+                    },
                     'force_destroy': True
                 }
             }
--- a/src/tests/test_buildercore_terraform.py
+++ b/src/tests/test_buildercore_terraform.py
@@ -37,6 +37,9 @@ class TestBuildercoreTerraform(base.BaseCase):
                                 'ssl_cert_hostname': 'prod--www.example.org',
                                 'ssl_check_cert': True,
                             },
+                            'gzip': {
+                                'name': 'project-with-fastly-minimal--prod',
+                            },
                             'force_destroy': True
                         }
                     }
@@ -74,6 +77,9 @@ class TestBuildercoreTerraform(base.BaseCase):
                                 'use_ssl': True,
                                 'ssl_cert_hostname': 'prod--www.example.org',
                                 'ssl_check_cert': True
+                            },
+                            'gzip': {
+                                'name': 'project-with-fastly-complex--prod',
                             },
                             'force_destroy': True
                         }


thewilkybarkid:
GZIP can also be enable at the origin.

giorgiosironi:
You can create a test stack with:
`./bldr launch:api-gateway,test`
That should make https://test--cdn-gateway.elifesciences.org/ping` available to see what compression is used.

thewilkybarkid:
Suffix-based GZIPping requires custom VCL, will do later.

thewilkybarkid:
Confirmed that https://github.com/terraform-providers/terraform-provider-fastly/issues/66 happens.

src/buildercore/terraform.py
```diff
@@ -10,6 +10,14 @@
 RESOURCE_TYPE_FASTLY = 'fastly_service_v1'
 RESOURCE_NAME_FASTLY = 'fastly-cdn'
 
+FASTLY_GZIP_TYPES = ['text/html', 'application/x-javascript', 'text/css', 'application/javascript',
+                     'text/javascript', 'application/json', 'application/vnd.ms-fontobject',
+                     'application/x-font-opentype', 'application/x-font-truetype',
+                     'application/x-font-ttf', 'application/xml', 'font/eot', 'font/opentype',
+                     'font/otf', 'image/svg+xml', 'image/vnd.microsoft.icon', 'text/plain',
+                     'text/xml']
+FASTLY_GZIP_EXTENSIONS = ['css', 'js', 'html', 'eot', 'ico', 'otf', 'ttf', 'json']
```
giorgiosironi: is there a standard way to sort these, or should we do it in the code, so that the templates stays stable?

thewilkybarkid: This is the order on https://docs.fastly.com/guides/basic-configuration/enabling-automatic-gzipping.

src/buildercore/terraform.py
```diff
@@ -31,6 +39,13 @@ def render(context):
                         'ssl_cert_hostname': context['full_hostname'],
                         'ssl_check_cert': True,
                     },
+                    'gzip': {
+                        'name': context['stackname'],
```
giorgiosironi: not sure what this `name` is used for, seems arbitrary from the documentation. Not a big deal to change if needed

thewilkybarkid: Tempted to just call it 'default'.

src/buildercore/terraform.py
giorgiosironi: We could sort in https://github.com/elifesciences/builder/pull/287/files#diff-c266af1366f88d308f6fcb68d77552bcR46 so that doing a diff in the future is easier?

thewilkybarkid: `sorted()`

giorgiosironi:
Applied to `api-gateway` stacks.

PR #288, opened 2018-04-23
ELPP-3598 Force SSL in Fastly

thewilkybarkid:
Seems to also set `xff` to `append`, probably the Fastly default too.


thewilkybarkid:
Hmm, seems similar to #287, things being set that shouldn't be.

thewilkybarkid:
Starting to apply.

thewilkybarkid:
Applied everywhere.

PR #289, opened 2018-04-23
Read CUSTOM_SSH_KEY also for EC2 access

giorgiosironi:
Would have put it into `settings.yml` as a user setting to specify only once, but that isn't read by the Vagrantfile which uses this environment variable.

### DIFF
--- a/docs/ssh-key.md
+++ b/docs/ssh-key.md
@@ -1,12 +1,32 @@
 # custom ssh key
 
+## Default
+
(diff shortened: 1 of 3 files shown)


lsh-0:
builder relies/relied on a bunch of very common defaults, some of which have come back to bite it. The AWS region used to be hardcoded to us-east-1 and that was fun re-integrating it. AWS credentials are another that haven't really been addressed yet and of course this one, ssh credentials. 

Fix em as needs must.

PR #290, opened 2018-04-23
Run tests in parallel

giorgiosironi:
~If this works and assuming CloudFormation rate limiting has been correctly solved, this could shave off 6-7 minutes in each build.~

Final result is 13-14 minutes rather than 30 per build.

### DIFF
--- a/.run-tests-ci.sh
+++ b/.run-tests-ci.sh
@@ -12,7 +12,6 @@ fi
 echo "Running tests"
 export PYTHONPATH="src"
 pytest \
-    --cov=src \
     -n 4 \
     -s \
     --junitxml=build/pytest-$envname.xml \
--- a/.run-tests-ci.sh
+++ b/.run-tests-ci.sh
@@ -13,6 +13,7 @@ echo "Running tests"
 export PYTHONPATH="src"
 pytest \
     -n 4 \
+    --dist=loadscope \
     -s \
     --junitxml=build/pytest-$envname.xml \
     src/tests src/integration_tests
--- a/.run-tests-ci.sh
+++ b/.run-tests-ci.sh
@@ -11,15 +11,20 @@ fi
 
 echo "Running tests"
 export PYTHONPATH="src"
+if [ "$envname" = "py27" ]; then
+    coverage_options="--cov-config=.coveragerc --cov-report= --cov=src"
+else
+    coverage_options=
+fi
 pytest \
-    --cov-config=.coveragerc \
-    --cov-report= \
-    --cov=src \
+    $coverage_options
     -n 4 \
     --dist=loadscope \
     -s \
-    --junitxml=build/pytest-$envname.xml \
+    --junitxml="build/pytest-$envname.xml" \
     src/tests src/integration_tests
 
-echo "Checking coverage report"
-coverage report --fail-under=50
+if [ ! -z "$coverage_options" ]; then
+    echo "Checking coverage report"
+    coverage report --fail-under=50
+fi
--- a/.run-tests-ci.sh
+++ b/.run-tests-ci.sh
@@ -26,5 +26,5 @@ pytest \
 
 if [ ! -z "$coverage_options" ]; then
     echo "Checking coverage report"
-    coverage report --fail-under=50
+    coverage report --fail-under=69
 fi
--- a/.run-tests-ci.sh
+++ b/.run-tests-ci.sh
@@ -9,6 +9,7 @@ if [ -z "$envname" ]; then
     exit 1
 fi
 
+echo "Running tests"
 export PYTHONPATH="src"
 pytest \
     --cov=src \
@@ -17,4 +18,5 @@ pytest \
     --junitxml=build/pytest-$envname.xml \
     src/tests src/integration_tests
 
+echo "Checking coverage report"
 coverage report --fail-under=67
--- a/.run-tests-ci.sh
+++ b/.run-tests-ci.sh
@@ -12,7 +12,7 @@ fi
 export PYTHONPATH="src"
 pytest \
     --cov=src \
-    -n 2 \
+    -n 4 \
     -s \
     --junitxml=build/pytest-$envname.xml \
     src/tests src/integration_tests
--- a/.run-tests-ci.sh
+++ b/.run-tests-ci.sh
@@ -12,6 +12,9 @@ fi
 echo "Running tests"
 export PYTHONPATH="src"
 pytest \
+    --cov-config=.coveragerc \
+    --cov-report= \
+    --cov=src \
     -n 4 \
     --dist=loadscope \
     -s \
--- a/.run-tests-ci.sh
+++ b/.run-tests-ci.sh
@@ -18,4 +18,4 @@ pytest \
     src/tests src/integration_tests
 
 echo "Checking coverage report"
-coverage report --fail-under=67
+coverage report --fail-under=50
--- a/.run-tests-ci.sh
+++ b/.run-tests-ci.sh
@@ -17,7 +17,7 @@ else
     coverage_options=
 fi
 pytest \
-    $coverage_options
+    $coverage_options \
     -n 4 \
     --dist=loadscope \
     -s \
--- a/src/buildercore/config.py
+++ b/src/buildercore/config.py
@@ -97,7 +97,8 @@ ROOTLOG.addHandler(FILE_HANDLER)
 
 LOG = logging.getLogger(__name__)
 logging.getLogger('paramiko.transport').setLevel(logging.ERROR)
-logging.getLogger('botocore.vendored').setLevel(logging.ERROR)
+# TODO: leave on for FILE_HANDLER but not for CONSOLE_HANDLER
+#logging.getLogger('botocore.vendored').setLevel(logging.ERROR)
 
 def get_logger(name):
     "ensures logging is setup before handing out a Logger object to use"
--- a/src/buildercore/config.py
+++ b/src/buildercore/config.py
@@ -81,19 +81,19 @@ ROOTLOG = logging.getLogger() # important! this is the *root LOG*
 ROOTLOG.setLevel(logging.DEBUG) # *default* output level for all LOGs
 
 # StreamHandler sends to stderr by default
-H1 = logging.StreamHandler()
-H1.setLevel(logging.INFO) # output level for *this handler*
-H1.setFormatter(CONSOLE_FORMAT)
+CONSOLE_HANDLER = logging.StreamHandler()
+CONSOLE_HANDLER.setLevel(logging.INFO) # output level for *this handler*
+CONSOLE_HANDLER.setFormatter(CONSOLE_FORMAT)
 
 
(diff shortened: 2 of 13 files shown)


giorgiosironi:
Even the backoff gave up after some time
```
_common.py                  93 ERROR    Giving up validate_template(...) after 7 tries (BotoServerError: BotoServerError: 400 Bad Request
```

giorgiosironi:
https://console.aws.amazon.com/support/v1?region=us-east-1#/case/?displayId=5011815901&language=en

giorgiosironi:
I think we have race conditions between the multiple copies of the test suite running together, but the rate limiting seems to have improved.

giorgiosironi:
The test artifacts need to avoid conflicts, as https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Fbuilder/detail/PR-290/2/tests only shows the latest run I think.

From the output of the failed parallel stage, I cannot understand which test(s) failed:
https://alfred.elifesciences.org/blue/rest/organizations/jenkins/pipelines/pull-requests-projects/pipelines/builder/branches/PR-290/runs/2/nodes/62/steps/70/log/?start=0

giorgiosironi:
Can finally zoom in on failures: https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Fbuilder/detail/PR-290/4/tests

giorgiosironi:
It seems `test_with_many_nodes` is the main culprit for the long test suite:
```
$ tail -n 10000 -f logs/app.log | grep TestWithManyNodes
2018-10-22 16:08:15,560 - INFO - MainProcess - conftest - Executing <class 'integration_tests.test_with_many_nodes.TestWithManyNodes'>::test_restart_all_started
2018-10-22 16:13:06,893 - INFO - MainProcess - conftest - Setting up <class 'integration_tests.test_with_many_nodes.TestWithManyNodes'>::test_restart_all_started
2018-10-22 16:17:11,305 - INFO - MainProcess - conftest - Tearing down up <class 'integration_tests.test_with_many_nodes.TestWithManyNodes'>::test_restart_all_started
2018-10-22 16:17:11,320 - INFO - MainProcess - conftest - Setting up <class 'integration_tests.test_with_many_nodes.TestWithManyNodes'>::test_restart_all_stopped
...
```
Parallelizing at the class level is not enough for this case, as it would still be a bottleneck; we need to make the 3 tests in there run all in parallel on different infrastructure.

src/integration_tests/test_with_instance.py
```diff
@@ -16,69 +12,16 @@
 # * run any logic against it that needs a running ec2 instance
 # * optionally leave the instance running while debugging happens
 
-class One(base.BaseCase):
+class TestWithInstance(base.BaseIntegrationCase):
     @classmethod
-    def setUpClass(self): # cls, not self
```
giorgiosironi: setup/teardown code was identical across two classes

src/tests/__init__.py
```diff
@@ -1,2 +1,5 @@
 import logging
-logging.disable(logging.CRITICAL)
+
+from buildercore.config import CONSOLE_HANDLER
+
+CONSOLE_HANDLER.setLevel(logging.CRITICAL)
```
giorgiosironi: we still log to the file, which is useful to debug strange test failures

lsh-0:
measuring coverage while running the tests in parallel has always been supported by the various tools, but I found they gave never gave consistent results across multiple runs and compared to the serial tests.

parallelism in python feels so clunky I almost want to return to raw threads and locks.

there is definitely something weird in the setup/teardown stage of builder tests. I have two tests that consistently fail while running locally but run fine individually.

src/buildercore/config.py
```diff
@@ -81,22 +81,27 @@ class ConfigurationError(Exception):
...
+FILE_HANDLER.setFormatter(FORMAT)
 
-ROOTLOG.addHandler(H1)
-ROOTLOG.addHandler(H2)
+ROOTLOG.addHandler(CONSOLE_HANDLER)
+ROOTLOG.addHandler(FILE_HANDLER)
 
 LOG = logging.getLogger(__name__)
 logging.getLogger('paramiko.transport').setLevel(logging.ERROR)
+logging.getLogger('botocore.vendored').setLevel(logging.ERROR)
```
lsh-0: what is the effect of this? 

I closed a similar PR recently https://github.com/elifesciences/builder/pull/339/files because the effect on the output would have been a bit radical

src/buildercore/lifecycle.py
```diff
@@ -50,30 +51,34 @@ def restart(stackname):
...
+            call_while(
+                lambda: _some_node_is_not_ready(stackname, node=node_id, concurrency='serial'),
+                interval=2,
+                update_msg="waiting for nodes to be networked",
+                done_msg="all nodes have public ips"
+            )
+        update_dns(stackname)
+        return history
+    except Exception:
+        LOG.info("Partial restart history of %s: %s", stackname, pformat(history))
```
lsh-0: :+1: 

there is a known bug attempting to restart an instance with a multi-az RDS instance - they can't be restarted apparently :(

src/integration_tests/test_provisioning.py
```diff
@@ -49,6 +49,7 @@ def test_create(self):
             buildvars.switch_revision(stackname, 'master')
             buildvars.force(stackname, 'answer', 'forty-two')
 
+            # TODO: remove as it is tested separately?
```
lsh-0: sure, if it's out of scope for the test.

src/buildercore/lifecycle.py
giorgiosironi: Yes, it's not supported by the API. In this case I stumbled upon a timeout while running a test, and wanted more information next time it happens.

giorgiosironi:
> there is definitely something weird in the setup/teardown stage of builder tests. I have two tests that consistently fail while running locally but run fine individually.

Which ones? I can try to reproduce.

.run-tests-ci.sh
```diff
@@ -9,12 +9,22 @@ if [ -z "$envname" ]; then
     exit 1
 fi
 
+echo "Running tests"
 export PYTHONPATH="src"
+if [ "$envname" = "py27" ]; then
+    coverage_options="--cov-config=.coveragerc --cov-report= --cov=src"
```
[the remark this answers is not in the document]
de-code: I'm not quite familiar with the setup for different python versions. Would you ever test python 2.7 and one python 3 version?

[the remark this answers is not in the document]
giorgiosironi: `builder` currently supports both `2.7` and `3.5`, as the `3.x` version has some performance problems in the libraries. Ideally we would only upgrade to `3.x` and support that. Still, if all the tests are executed in both versions, not very useful to have two code coverage reports when they should practically be identical.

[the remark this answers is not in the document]
de-code: Okay, this is just to test the builder project itself?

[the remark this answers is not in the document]
giorgiosironi: Yes, creating some fictional resources in AWS and similar, updating them, destroying them, etc.

src/buildercore/config.py
```diff
@@ -81,22 +81,28 @@ class ConfigurationError(Exception):
 ROOTLOG.setLevel(logging.DEBUG) # *default* output level for all LOGs
 
 # StreamHandler sends to stderr by default
-H1 = logging.StreamHandler()
-H1.setLevel(logging.INFO) # output level for *this handler*
-H1.setFormatter(CONSOLE_FORMAT)
+CONSOLE_HANDLER = logging.StreamHandler()
```
de-code: I know this has been like this before. But side effects like changing the logging configuration should ideally not happen by just importing a module.

src/tests/__init__.py
giorgiosironi: > side effects don't belong into a __init__.py file I think

> BTW by default pytest will only show the logging or console output for failing tests. In that case I've usually done the opposite, set the default logging to DEBUG

Essentially there is too much output, so we disable most of the `CONSOLE_HANDLER` logs and leave `FILE_HANDLER` in place so that the logs can be seen in a file after the test run.

Yes, it could be, although maybe `src/tests/conftest.py` rather than the main one.

> BTW by default pytest will only show the logging or console output for failing tests. In that case I've usually done the opposite, set the default logging to DEBUG

> I saw pytest further up there, is that not what we are using?

Yes, we are exclusively running tests with `pytest` now

src/buildercore/config.py
```diff
@@ -81,22 +81,28 @@ class ConfigurationError(Exception):
...
+ROOTLOG.addHandler(CONSOLE_HANDLER)
+ROOTLOG.addHandler(FILE_HANDLER)
 
 LOG = logging.getLogger(__name__)
 logging.getLogger('paramiko.transport').setLevel(logging.ERROR)
+# TODO: leave on for FILE_HANDLER but not for CONSOLE_HANDLER
+# logging.getLogger('botocore.vendored').setLevel(logging.ERROR)
+
+def get_logger(name):
+    "ensures logging is setup before handing out a Logger object to use"
```
de-code: Apparently this is not necessary in general. Is that just so that for example tests import this module which then configures the logging? I would recommend to move the setup of the logging to a method something like `configure_logging` and call it from a fixture in `conftest` and from the main entry point wherever it seems sensible (usually I would add after `if __name__ == "__main__"`).

(I only found a `get_logger` necessary in combination with Apache Beam which serialises and deserialises the pipeline)

de-code:
Don't know enough about `builder` but seems to make sense and if it works.
(Just made minor comments about style regarding logging, feel free to ignore, just don't think it follows best practices)

src/buildercore/config.py
giorgiosironi: There is more than one entry point to cover, but logging configuration always stuck out as a global state problem. I'll try this in a separate PR to see how it may work.

src/buildercore/config.py
giorgiosironi: I agree because it's global state, but previously found some ramifications that make it difficult to avoid.

src/tests/__init__.py
de-code: > Essentially there is too much output, so we disable most of the `CONSOLE_HANDLER` logs and leave `FILE_HANDLER` in place so that the logs can be seen in a file after the test run.

When you say too much output, you mean too much shown or was it making it too slow? Because if it's failing, it seems good to have a lot of output. Usually it shouldn't fail and have no output. The file log however will have everything in it, for passing and failing tests.
If you see console output for passing tests, then there may be something wrong with the setup.

> Yes, it could be, although maybe `src/tests/conftest.py` rather than the main one.

By that you mean the default logging level DEBUG? Yes, for tests I meant (i.e. in conftest). When running the application the default logging level would usually be INFO.
Or do you mean another `conftest.py`?

src/buildercore/config.py
lsh-0: don't get me started on logging ...

PR #292, opened 2018-04-24
Feature: boto2 removal remaining outliers

lsh-0:
cleans up any adhoc boto2 things

### DIFF
--- a/src/aws.py
+++ b/src/aws.py
@@ -6,15 +6,17 @@ LOG = logging.getLogger(__name__)
 
 @debugtask
 @requires_aws_stack
-@echo_output
 def rds_snapshots(stackname):
-    from boto import rds
-    conn = rds.RDSConnection()
-    instance = conn.get_all_dbinstances(instance_id=stackname)[0]
-    # all snapshots order by creation time
-    objdata = conn.get_all_dbsnapshots(instance_id=instance.id)
-    data = sorted(map(lambda ss: ss.__dict__, objdata), key=lambda i: i['snapshot_create_time'])
-    return data
+    "prints all snapshots for given stack, order by creation time"
+    inst = core.find_rds_instances(stackname)[0]
+    conn = core.boto_conn(stackname, 'rds', client=True)
+    snapshots = conn.describe_db_snapshots(**{
+        'DBInstanceIdentifier': inst['DBInstanceIdentifier'],
+    })['DBSnapshots']
+    data = [(ss['DBSnapshotIdentifier'], ss['SnapshotType'], ss['SnapshotCreateTime']) for ss in snapshots]
+    data = sorted(data, key=lambda triple: triple[2])
+    for row in data:
+        print(row)
 
 @debugtask
 @echo_output
--- a/src/buildercore/lifecycle.py
+++ b/src/buildercore/lifecycle.py
@@ -8,7 +8,7 @@ import re
 from fabric.contrib import files
 import fabric.exceptions as fabric_exceptions
 from . import config, core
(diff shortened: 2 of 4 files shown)


src/aws.py
```diff
@@ -6,15 +6,17 @@
...
+    "prints all snapshots for given stack, order by creation time"
+    inst = core.find_rds_instances(stackname)[0]
+    conn = core.boto_conn(stackname, 'rds', client=True)
+    snapshots = conn.describe_db_snapshots(**{
+        'DBInstanceIdentifier': inst['DBInstanceIdentifier'],
+    })['DBSnapshots']
+    data = [(ss['DBSnapshotIdentifier'], ss['SnapshotType'], ss['SnapshotCreateTime']) for ss in snapshots]
+    data = sorted(data, key=lambda triple: triple[2])
+    for row in data:
+        print(row)
```
giorgiosironi: this is low-risk, only used to list snapshots in a manual process

src/buildercore/lifecycle.py
```diff
@@ -283,3 +281,9 @@ def _ec2_connection(stackname):
 
 def _rds_connection(stackname):
     return boto_conn(stackname, 'rds')
+
+def route53conn():
+    """returns a boto2 route53 connection.
+    route53 for boto3 is *very* poor and much too low-level with no 'resource' construct (yet?). It should be avoided"""
+    import boto # will only ever be imported once
```
giorgiosironi: I agree with the isolation but I wouldn't hide `boto` inside a method, imports should be visible at the top of the file?

giorgiosironi: In the future, this could be ported to https://boto3.readthedocs.io/en/latest/reference/services/route53.html#Route53.Client.change_resource_record_sets

src/aws.py
lsh-0: yeah, I was going to get rid of it entirely but felt it might serve as a demonstration for anyone looking at tasks and how to do one

src/buildercore/lifecycle.py
lsh-0: as per my comment, I don't think the boto3 route53 interface is up to scratch and `change_resource_record_sets` seems far too low level to be switching *to*. I spent some time trying to port the code to boto3 but the translation was just ridiculous. You can't select a specific zone for example, only list all zones with a preference as to which one should be first, if it exists at all. So instead of `get_zone(zone_name)` you need to fetch a list of summaries, ensure the first record has a zone value that matches the one given, and then another call to `get_hosted_zone` using the `Id` from the summary result you just fetched. And then this zone dict has to be pulled apart to get the resource records, that exhibit similar behaviour (no `get_a` methods here) in order to build up an elaborate record for `change_resource_record_sets`.

I reverted most of the code I had and stuck with the boto2 version. It's stable, tested and the interface is convenient.

giorgiosironi: :man_shrugging: sometimes you eat the SDK and sometimes the SDK eats you. I think later on I'll try to move this code into a `buildercore.route53` module which could become the only module including boto2.

giorgiosironi:
This seems safe to merge now. I'll do it later.

src/buildercore/lifecycle.py
lsh-0: (ha)

I've been thinking about this actually and it was the reason I created a `sns.py` module. `core.py` and `bootstrap.py` could have big chunks of logic moved out into discrete AWS modules like `sns` or `ec2` or `cloudformation`. The more generic functions I've probably labelled with a TODO.

As core and bootstrap get larger because we're accommodating more stack features it might be worthwhile to do an experimental refactor and see how well it copes.

giorgiosironi:
YOLO

lsh-0:
move fast and **BREAK EVERYTHING**

kaboom

giorgiosironi:
Just a minor `update_infrastructure` use case in the end :grinning:

PR #294, opened 2018-04-24
ELPP-3521 Correct Terraform file extension

giorgiosironi:
Following https://www.terraform.io/docs/configuration/index.html although the `.tf` extension has worked so far. Separates all the handling of files into the `buildercore.terraform` module.

### DIFF
--- a/src/tests/test_buildercore_terraform.py
+++ b/src/tests/test_buildercore_terraform.py
@@ -1,6 +1,7 @@
 import os
+import shutil
 import yaml
-from os.path import join
+from os.path import exists, join
 from . import base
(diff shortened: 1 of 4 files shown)


src/tests/test_buildercore_terraform.py
```diff
@@ -128,6 +132,11 @@ def test_fastly_template_complex(self):
             data
         )
 
+    def test_write_template(self):
```
thewilkybarkid: Also tests read?

giorgiosironi: Indeed, test names should be scenarios, not components

PR #295, opened 2018-04-24
ELPP-3521 Extend Fastly Terraform template to include GCS logging

giorgiosironi:
Like CloudFront, a CDN has its own servers so if you want access logs to analyze, you need to configure it to dump them somewhere.

CloudFront did this to an S3 bucket. Fastly can do the same but with many different targets, one that could be useful for us is Google Cloud Storage which could then be imported into Google Big Query.

Credentials manipulation will be needed - Fastly will need a service account (like we do with AWS IAM users called `lax`) which I created, and I'm now figuring out the correct permissions to give it. This PR forces the user to configure two environment variables before being able to use the `fastly.gcslogging` feature.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -369,7 +369,7 @@ api-gateway:
                 gcslogging:
                     bucket: end2end-elife-fastly
                     path: api-gateway--test/
-                    period: 120
+                    period: 3600
     vagrant:
         box: ubuntu/trusty64 # Ubuntu 14.04, deprecated
         ports:
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -133,6 +133,7 @@ defaults:
             subdomains-without-dns: []
             dns:
                 cname: u2.shared.global.fastly.net
+            gcslogging: false
         subdomains: []
         elasticache:
             # elasticache defaults only used if an `rds` section present in project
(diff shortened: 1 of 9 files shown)


src/buildercore/terraform.py
```diff
@@ -61,6 +61,16 @@ def render(context):
             }
         },
     }
+    if context['fastly']['gcslogging']:
+        gcslogging = context['fastly']['gcslogging']
+        # TODO: require FASTLY_GCS_EMAIL env variable
+        # TODO: require FASTLY_GCS_SECRET env variable
```
giorgiosironi: no one wants to copy arbitrary credentials on a laptop to be able to run `update_infrastructure`, I think the recently revamped S3 module that manages EC2 SSH keys can be extended to store also these credentials (they are not keypairs, but similar) /cc @lsh-0

lsh-0: are they files that are generated for us by Fastly?

giorgiosironi: The credentials are generated by our Google Cloud console, and are needed for Fastly to write to Google Cloud Storage; means we need to push them into the Fastly Terraform template so that they are available to their servers. I think Terraform must have some standard way of handling credentials so I'll look for some standard pattern there first.

projects/elife.yaml
```diff
@@ -361,6 +362,14 @@ api-gateway:
                 subdomains:
                     - "{instance}--cdn-gateway"
                     - api
+        test:
+            fastly:
+                subdomains:
+                    - "{instance}--cdn-gateway"
+                gcslogging:
+                    bucket: end2end-elife-fastly
```
giorgiosironi: see https://console.cloud.google.com/storage/browser/end2end-elife-fastly/api-gateway--test/?project=elife-fastly&authuser=1&folder=true&organizationId=567577722038

giorgiosironi: This is using the default format for logs, but I will integrate this:
https://docs.fastly.com/guides/streaming-logs/log-streaming-google-bigquery#example-format

PR #296, opened 2018-04-25
fixes issue where salt-minion fails forever when it's config is switched

lsh-0:
out while it's running


lsh-0:
please merge if all good

scripts/init-vagrant-formulas.sh
```diff
@@ -10,6 +11,9 @@ set -xv  # output the scripts and interpolated steps
 
 minion_id=$1
 
+# ELPP-3601: Vagrant machines run masterless and don't need salt-minion running
+systemctl stop salt-minion 2> /dev/null || service salt-minion stop
```
giorgiosironi: ah, nice to know

PR #297, opened 2018-04-25
ELPP-3242 Update bldr stop_if_running_for

giorgiosironi:
The task was relying on a string being returned in the field describing how long an instance has been running, this is now a `datetime` object.

/cc @lsh-0

### DIFF
--- a/src/buildercore/lifecycle.py
+++ b/src/buildercore/lifecycle.py
@@ -114,7 +114,8 @@ def _last_ec2_start_time(stackname):
     nodes = find_ec2_instances(stackname, allow_empty=True)
 
     def _parse_datetime(value):
-        return datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%fZ")
+        assert value.tzname() == 'UTC', 'datetime object returned by the EC2 API is not UTC, needs timezone conversion'
+        return value.replace(tzinfo=None)
     return {node.id: _parse_datetime(node.launch_time) for node in nodes}
 
 def _stop(stackname, ec2_to_be_stopped, rds_to_be_stopped):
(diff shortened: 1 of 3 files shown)


giorgiosironi:
```
E       AttributeError: 'str' object has no attribute 'tzname'
```
So sometimes it's a string, sometimes it's a `datetime`? More likely there is some environmental difference in Alfred.

giorgiosironi:
Environment on Alfred is the same as me
```
boto==2.46.1
boto3==1.4.4
botocore==1.5.95
```

giorgiosironi:
It was a unit test failing, which stubbed that field.

src/buildercore/lifecycle.py
```diff
@@ -114,7 +114,8 @@ def _last_ec2_start_time(stackname):
     nodes = find_ec2_instances(stackname, allow_empty=True)
 
     def _parse_datetime(value):
-        return datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%fZ")
+        assert value.tzname() == 'UTC', 'datetime object returned by the EC2 API is not UTC, needs timezone conversion'
+        return value.replace(tzinfo=None)
```
lsh-0: Stripping timezone information results in a 'naive' datetime object, which is not good a state to be in. Your tests are also specifying `utc`, so it's strange.

giorgiosironi: This was a quick fix that did the job of letting the code run, but it's open for improvement

lsh-0: I'm certain my future self will find it, wrinkle his nose and feel compelled to fix it

PR #291, opened 2018-04-24
Feature: boto2 removal of sqs, sns

lsh-0:
* sqs
* sns

PR #300, opened 2018-04-26
Add bugfix  for Queue object creation

giorgiosironi:
Follow-up to https://github.com/elifesciences/builder/pull/291

/cc @lsh-0


giorgiosironi:
Yes, we could add it, after I am done working with AWS support to improve our CloudFormation API call quota (so that we can run more stuff in parallel)

PR #301, opened 2018-04-27
Feature: reboot multi nodes serially

lsh-0:
tickets:
https://elifesciences.atlassian.net/browse/ELPP-3374
https://elifesciences.atlassian.net/browse/ELPP-3021

### DIFF
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -227,13 +227,38 @@ def rds_dbname(stackname, context=None):
     return context.get('rds_dbname') or slugify(stackname, separator="") # *must* use 'or' here
 
 def rds_iid(stackname):
-    return slugify(stackname)
+    max_rds_iid = 63 # https://docs.aws.amazon.com/cli/latest/reference/rds/create-db-instance.html#options
+    # the rds iid needs to be deterministic, or, we need to find an attached rds db without knowing it's name
+    slug = slugify(stackname)
+    ensure(len(slug) <= max_rds_iid, "a database instance identifier must be less than 64 characters")
+    return slug
 
 def find_rds_instances(stackname, state='available'):
     "This uses boto3 because it allows to start/stop instances"
-    conn = boto_conn(stackname, 'rds', client=True) # RDS has no 'resource'
-    rid = rds_iid(stackname)
-    return conn.describe_db_instances(DBInstanceIdentifier=rid)['DBInstances']
+    try:
+        conn = boto_conn(stackname, 'rds', client=True) # RDS has no 'resource'
+        rid = rds_iid(stackname)
+        if rid:
+            # TODO: return the first (and only) result of DBInstances
+            return conn.describe_db_instances(DBInstanceIdentifier=rid)['DBInstances']
+    except AssertionError:
+        # invalid dbid. RDS doesn't exist because stack couldn't have been created with this ID
+        return []
+
+    except botocore.exceptions.ClientError as err:
+        LOG.info(err.response)
+        msg = err.response['Error']['Message']
+        invalid_dbid = "Invalid database identifier"
+        if msg.startswith(invalid_dbid):
+            # what we asked for isn't a valid db id, we probably made a mistake
+            # we definitely couldn't have created a db with that id
+            return []
+
+        db_notfound = "DBInstanceNotFound"
+        if msg.startswith(db_notfound):
+            return []
+
+        raise err
 
 #
 #
@@ -291,7 +316,7 @@ def stack_all_ec2_nodes(stackname, workfn, username=config.DEPLOY_USER, concurre
     public_ips = {ec2['InstanceId']: ec2['PublicIpAddress'] for ec2 in data}
     nodes = {ec2['InstanceId']: int(tags2dict(ec2['Tags'])['Node']) if 'Node' in tags2dict(ec2['Tags']) else 1 for ec2 in data}
     if node:
-        nodes = {k: v for k, v in nodes.items() if v == node}
+        nodes = {k: v for k, v in nodes.items() if v == int(node)}
         public_ips = {k: v for k, v in public_ips.items() if k in nodes.keys()}
 
     params = _ec2_connection_params(stackname, username)
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -14,6 +14,7 @@ from contextlib import contextmanager
 from fabric.api import settings, execute, env, parallel, serial, hide, run, sudo
 from fabric.exceptions import NetworkError
 from fabric.state import output
+from slugify import slugify
 import importlib
 import logging
 from kids.cache import cache as cached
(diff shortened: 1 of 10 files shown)


src/buildercore/lifecycle.py
```diff
@@ -10,11 +10,37 @@
...
+    return lmap(_start, rds_to_be_started.keys())
+
+def restart(stackname):
+    "for each ec2 node in given stack, ensure ec2 node is stopped, then start it, then repeat. rds is started if stopped but otherwise not affected"
+    start_rds_nodes(core.find_rds_nodes(stackname))
+    for node in ec2_nodes(stackname):
+        node.stop()
+        node.wait_until_stopped()
+        node.start()
+        node.wait_until_running()
```
giorgiosironi: DNS may need to be updated (although if there is more than one node, nothing needs to be done), see https://github.com/elifesciences/builder/pull/301/files#diff-e26cdd0bfc8a5ff0ffb7859c59362a18L195

giorgiosironi: Also a running state is usually no guarantee that services have started, see https://github.com/elifesciences/builder/pull/301/files#diff-e26cdd0bfc8a5ff0ffb7859c59362a18R196 that waits for a standard file that signals the boot has been completed

lsh-0:
ready to be merged

src/buildercore/lifecycle.py
```diff
@@ -10,11 +10,74 @@
...
 from .context_handler import load_context, download_from_s3
 
 LOG = logging.getLogger(__name__)
 
+def _node_id(node):
+    name = core.tags2dict(node.tags)['Name']
+    nid = core.parse_stackname(name, all_bits=True, idx=True).get('cluster_id', 1)
+    return int(nid)
+
+def ec2_nodes(stackname):
```
giorgiosironi: needs a qualifier in the name? `ec2_nodes` is very generic and doesn't convey the filtering or ordering being done here

src/buildercore/lifecycle.py
```diff
@@ -10,11 +10,74 @@
...
+    "returns all non-terminated nodes associated with stackname, ordered by node-id"
+    nodes = find_ec2_instances(stackname, state='pending|running|shutting-down|stopping|stopped')
```
giorgiosironi: if you are excluding `terminated`, I think also `shutting-down` should be excluded as a node like that is beyond being used anymore. It's a corner case

giorgiosironi: (`shutting-down` is going towards being `terminated` without further intervention: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html)

src/integration_tests/test_with_many_nodes.py
```diff
@@ -0,0 +1,145 @@
...
+            [(1, 'stopped'), (2, 'running')],
+            [(1, 'pending'), (2, 'running')],
+            [(1, 'running'), (2, 'running')],
+
+            # node2, stopped -> running
+            [(1, 'running'), (2, 'stopping')],
+            [(1, 'running'), (2, 'stopped')],
+            [(1, 'running'), (2, 'pending')],
+            [(1, 'running'), (2, 'running')],
+        ]
```
giorgiosironi: the history is a nice way to assert on what has happened

giorgiosironi:
The tests for restarts are necessary, but we have to find a way to run at least the `integration_tests` them in parallel. https://pypi.org/project/pytest-xdist/ is what I use in elife-spectrum. The build time of `master` is ~20 minutes at the moment and this PR doubles it to >40 (https://alfred.elifesciences.org/job/pull-requests-projects/job/builder/job/PR-301/)

src/buildercore/core.py
```diff
@@ -221,6 +217,53 @@ def _all_nodes_filter(stackname, node_ids):
...
+    max_rds_iid = 63 # https://docs.aws.amazon.com/cli/latest/reference/rds/create-db-instance.html#options
+    # the rds iid needs to be deterministic, or, we need to find an attached rds db without knowing it's name
+    slug = slugify(stackname)
+    ensure(len(slug) <= max_rds_iid, "a database instance identifier must be less than 64 characters")
+    return slug
+
+def find_rds_instances(stackname, state='available'):
+    "This uses boto3 because it allows to start/stop instances"
+    try:
+        conn = boto_conn(stackname, 'rds', client=True) # RDS has no 'resource'
```
giorgiosironi: I think the failure in https://alfred.elifesciences.org/job/process-stop-stacks-long-term/57/console is related. Will leave it there to avoid conflicts

src/integration_tests/test_with_many_nodes.py
lsh-0: when I get frustrated with code and want to tear it all down (which is like, always) I describe exactly what I want and then figure out how to get there. In this case it was pleasantly easy to get to this output.

lsh-0:
> The tests for restarts are necessary, but we have to find a way to run at least the integration_tests them in parallel.

can we have jenkins do something like: 
run unit tests|run integration tests suite1.py|run suite2.py|run suite3.py

(where each pipe is a separate process)

It looks like the Jenkins file has some code in it, so I think the individual integration suites could be enumerated easily.

lsh-0:
hmm .. is it the template validation that is hurting us? or the polling while bringing up instances?

giorgiosironi:
See the ticket linked in #290, they don't know yet but it may be applying to the while set of CloudFormation API calls, which are used by all integration tests.

giorgiosironi:
Caught up in Vault today, but I'll amend this Jenkinsfile tomorrow to try to run some of the stuff in parallel. Given the possible instability, I'll do it as a PR to this branch, if it doesn't work it can be cleanly discarded.

giorgiosironi:
Attempting to run tests in 2 parallel processes to speed it up. `src/tests` went from 40 seconds to 25, `src/integration_tests` may improve or break horribly.

giorgiosironi:
The first test suite for Python 2.7 passed in ~13 minutes (rather than ~20), but coverage broke. It's understandable as the separate processes may require their own instrumentation, let's check out how to integrate both.

giorgiosironi:
Seems to work in https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Fbuilder/detail/PR-301/12/pipeline/49 coverage at 73%; the target can probably be raised a bit.

giorgiosironi:
Two tests failed in the second suite, probably by chance:

`src.tests.test_buildercore_cfngen.TestUpdates.test_template_delta_includes_parts_of_rds`
```
botocore.exceptions.ClientError: An error occurred (BadDigest) when calling the PutObject operation (reached max retries: 4): The Content-MD5 you specified did not match what we received
```
Probably we have to decouple S3 here. However it points to a race condition, possibly multiple tests using the same S3 objects.

`src.integration_tests.test_with_many_nodes.One.test_restart_all_started` is a valid run:
```
E         [[(1, 'running'), (2, 'running')],
E       -  [(1, 'stopping'), (2, 'running')],
E          [(1, 'stopped'), (2, 'running')],
```
The instance was faster than the test in shutting down and the `stopping` state was not observed.

giorgiosironi:
The first test class stability should improve now, the second one remains to be updated. Several runs need to finish now, will take a while and times won't be indicative because there are some other `builder` builds running using the lock.

lsh-0:
yes, code coverage gets dodgy when running in parallel. coverage has support for merging results but those results too are inconsistent and error prone. I wouldn't trust that 73%

PR #304, opened 2018-05-01
Double space available to xpub containers

giorgiosironi:
Despite running docker-prune often, we had several cases of builds failing due to lack of disk space.

Currently appling to `ci` and then `demo`. However their disk seems to be `magnetic` so can't be expanded in place, will have to recreate it (probably the whole server).

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1718,7 +1718,7 @@ elife-xpub:
             - 80
         ext:
             # external hdd
-            size: 10 # GB
+            size: 20 # GB
             device: /dev/sdh
     aws-alt: {}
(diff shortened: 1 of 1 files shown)


giorgiosironi:
First attempt: delete the disk, readd-it.

giorgiosironi:
No luck, the original disk was in use and now the instance can't boot because it's trying to mount a disk that is detached (or maybe destroyed). Recreating `elife-xpub--ci`.

giorgiosironi:
Now recreating `elife-xpub--demo`. This time the disks are SSD, so they should be enlargeable in place if needed.

PR #305, opened 2018-05-03
ELPP-3598 Configuration mechanicm for custom VCL

giorgiosironi:
The only thing we can allow through Terraform at the moment is to fully specify a custom VCL. We don't strictly need to use this option for now, as snippets may be easier to integrate if they get implemented in time.

So, not going the whole mile but only keeping the space open in the configuration to specify VCL.


src/buildercore/terraform.py
```diff
@@ -52,6 +52,29 @@
...
 FASTLY_LOG_LINE_PREFIX = 'blank' # no prefix
 
+# at the moment VCL snippets are unsupported, this can be worked
+# around by using a full VCL
+# https://github.com/terraform-providers/terraform-provider-fastly/issues/7 tracks when snippets could become available in Terraform
+FASTLY_MAIN_VCL_KEY = 'main'
+FASTLY_CUSTOM_VCL = {
+    # take from https://manage.fastly.com/configure/services/4Wswrt0KnqbtpcohzrIDDx/versions/7/vcl for example
+    # not pasting it here at the moment as needs customization of values related to customer, hostnames, etc. plus maintenance
+    FASTLY_MAIN_VCL_KEY: """to be defined""",
```
thewilkybarkid: https://docs.fastly.com/guides/vcl/mixing-and-matching-fastly-vcl-with-custom-vcl#fastlys-vcl-boilerplate?

giorgiosironi: I can paste that here if that solves the problem. Duplication seems removed through the `#FASTLY` macros

giorgiosironi:
Some problem with the new VCL:
```
* fastly_service_v1.fastly-cdn: [ERR] Invalid configuration for Fastly Service (sCzIYxt2LzokGGLeQnWhq): Syntax error: Expected one of: `acl
`, `callback`, `sub`, `backend`, `director`, `import`, `table`, or `pragma`
at: (input Line 41 Pos 1)
.cfn/terraform/api-gateway--vcltest/main.vcl
#-------------------------------------------
```
which is this segment
```
      if (beresp.http.Expires || beresp.http.Surrogate-Control ~ "max-age" || beresp.http.Cache-Control ~ "(s-maxage|max-age)") {
        # keep the ttl here
      } else {
        # apply the default ttl 
        set beresp.ttl = 3600s;
      }  
```

giorgiosironi:
Fixed now, example at https://manage.fastly.com/configure/services/1jssdFTux7x4TEME7dw7JQ/versions/1/vcl
It seems there is a begin section:
```
pragma optional_param default_ssl_check_cert 1;
pragma optional_param max_backends 5;
pragma optional_param customer_id "4lxcAJz4L0ev2SPzwlFVTt";
C!
W!
# Backends
backend F_api_gateway__vcltest {
    ...
}
```
and an end section:
```
sub vcl_pipe {
    ...
}
sub vcl_hash {
    ...
}
```
there are inserted regardless of this `main` template.

The `gzip-by-regex` template is currently unused, as expected; it's a library step not referenced by `main`.

thewilkybarkid:
Just did a diff, and it's the same apart from whitespace and ordering changes.

thewilkybarkid:
To activate the custom VCL we'll need to add `include "gzip-by-regex"` inside `vcl_fetch`, or just generate a single file (might be easier to diff?).

giorgiosironi:
Implemented the addition of `include` statements, and applied to a test CDN.

It appears that in the final Fastly-side version available for viewing on the panel, VCL `include`s are resolved and pasted inside the main configuration:
```
    sub vcl_fetch {
        if ((beresp.status == 200 || beresp.status == 404) && (beresp.http.content-type ~ "(\+json)\s*($|;)" || req.url ~ "\.(css|js|html|eot|ico|otf|ttf|json|svg)($|\?)" ) ) {
          # always set vary to make sure uncompressed versions dont always win
```
https://manage.fastly.com/configure/services/1jssdFTux7x4TEME7dw7JQ/versions/2/vcl

src/buildercore/terraform.py
```diff
@@ -56,6 +58,167 @@
 # see https://docs.fastly.com/guides/streaming-logs/changing-log-line-formats#available-message-formats
 FASTLY_LOG_LINE_PREFIX = 'blank' # no prefix
 
+# at the moment VCL snippets are unsupported, this can be worked
```
giorgiosironi: all this new stuff should go into a `fastly` module, where the older Fastly code can be moved too later

src/buildercore/terraform.py
```diff
@@ -56,6 +58,167 @@
...
+        lines = list(self._lines)
+        lines.insert(section_start + 1, '')
+        lines.insert(
+            section_start + 1,
+            '%s  %s' % (indentation, statement)
+        )
+        return FastlyVCL(lines)
+
+    def _find_section_start(self, section):
+        lookup = r"(?P<indentation> +)sub vcl_%s {" % section
```
giorgiosironi: not sure whether the snippets should be included before or after the #FASTLY macro. I embrace change

giorgiosironi:
Doesn't seem to work at the moment as `Vary` is not modified:
```
$ curl -v https://vcltest--gateway.elifesciences.org/articles -o /dev/null 2>&1 | grep Vary
< Vary: Accept
< Vary: Authorization
```

thewilkybarkid:
Make a few changes. In the future we might want to run snippets before or after Fastly's default, but for now after is fine.

src/buildercore/fastly.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def insert(self, section, statements):
+        section_start = self._find_section_start(section)
+        lines = list(self._lines)
+        lines[section_start + 1:section_start + 1] = ['  %s' % s for s in statements]
+        lines.insert(section_start + 1, '')
+        return FastlyVCL(lines)
+
+    def _find_section_start(self, section):
+        lookup = r"^( *)#FASTLY %s" % section
```
giorgiosironi: maybe this will emerge as a configurable matcher later

PR #306, opened 2018-05-03
Execute pyflakes, pylint, and scrub in parallel

giorgiosironi:
Separate commit statuses give faster feedback for these early failures.

Preserved `.lint.sh` for local usage.


giorgiosironi:
Separate commit statuses give faster feedback for these early failures.

lsh-0:
Scrub script modifies files so it shouldn't be done in parallel to to the linting. I haven't seen it happen in a very long time, but a change by the scrub script can cause a pylint issue. This was back when I was letting pylint be comprehensive in it's checking (now it's almost entirely errors we look for).

I'm also not sure what the overlap is between pyflakes and pylint. Pylint has always been slower but more thorough but I'm not sure if Pylint would fail to find anything pyflakes wouldn't. Pyflakes is so fast the overhead in parallelism might not be worth it.

For CI I would have the pipeline go:

`commit -> scrub -> pyflakes|lint|cc-check -> test`

not sure if builder has a complexity checker (could of sworn it did), but it doesn't modify files either

giorgiosironi:
> For CI I would have the pipeline go:
> commit -> scrub -> lint|cc-check -> test

I'll extract scrub so that it is done first, still with a separate commit status that tells you it has failed.

PR #309, opened 2018-05-04
Show diff of Terraform template when updating

giorgiosironi:
Currently the whole new file was printed. Not perfect but it gives an overview of what we are going to change:
```
INFO - cfn - Create: {'Outputs': {}, 'Resources': {}}
INFO - cfn - Update: {'Outputs': {}, 'Resources': {}}
INFO - cfn - Delete: {'Outputs': {}, 'Resources': {}}
INFO - cfn - Terraform delta: {'dictionary_item_removed': set(["root['resource']['fastly_service_v1']['fastly-cdn']['backend']['healthcheck']",
 'values_changed': {"root['resource']['fastly_service_v1']['fastly-cdn']['gcslogging']['period']": {'new_value': 900,
                                 'old_value': 1800}}}
```


projects/elife.yaml
```diff
@@ -369,7 +369,7 @@ api-gateway:
                 gcslogging:
                     bucket: end2end-elife-fastly
                     path: api-gateway--test/
-                    period: 1800
+                    period: 900
```
giorgiosironi: this is just to exemplify what would happen when running `update_infrastructure:api-gateway--delta`

src/buildercore/cfngen.py
```diff
@@ -418,14 +419,23 @@ def non_empty(self):
 _empty_cloudformation_dictionary = {'Resources': {}, 'Outputs': {}}
 Delta.__new__.__defaults__ = (_empty_cloudformation_dictionary, _empty_cloudformation_dictionary, _empty_cloudformation_dictionary, None)
 
+class TerraformDelta(namedtuple('TerraformDelta', ['old_contents', 'new_contents'])):
```
lsh-0: I have never seen this before. Please add a comment or a link to just how this is working.

src/buildercore/cfngen.py
```diff
@@ -418,14 +419,23 @@ def non_empty(self):
...
+    def diff(self):
+        return deepdiff.DeepDiff(json.loads(self.old_contents), json.loads(self.new_contents))
+
 def template_delta(context):
     """given an already existing template, regenerates it and produces a delta containing only the new resources.
 
     Some the existing resources are treated as immutable and not put in the delta. Most that support non-destructive updates like CloudFront are instead included"""
     old_template = read_template(context['stackname'])
     template = json.loads(cloudformation.render_template(context))
+    old_terraform_template_file = '{}'
```
lsh-0: weird magic literal, it's beginning to smell a lot like eval. or php. A simple `empty_template = '{}'` and then `old_terraform_template_file = empty_template` would do

lsh-0:
`deepdiff` looks fine, I'm sure we'll find a use for it elsewhere as well. Fix up those two things and she'll be right for merging.

src/buildercore/cfngen.py
giorgiosironi: `TerraformDelta`? Was just introduced here as a new class, adding a description.

lsh-0: no - I mean, it looks like you're defining a class but inheriting from an instantiated object? It's as alien as: `class Foo({"bar?": "baz"}): ...` is to me.

giorgiosironi: `namedtuple()` returns a class, that part seems standard:
https://docs.python.org/2/library/collections.html#namedtuple-factory-function-for-tuples-with-named-fields
Then I needed to add methods to that class, so I extended it with inheritance. No need for storing the original `namedtuple`-created class as only the extended one should be used. I'll add some comments about this pattern.

PR #310, opened 2018-05-04
ELPP-3598 Healthcheck real world feedback

giorgiosironi:
New Relic probes from Brazil and Japan fail completely due to the local Fastly servers not reaching the origin within the healthcheck timeouts.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -374,7 +374,7 @@ api-gateway:
                 healthcheck:
                     path: /ping-fastly
                     check-interval: 30000
-                    timeout: 10000
+                    timeout: 2000
     vagrant:
         box: ubuntu/trusty64 # Ubuntu 14.04, deprecated
         ports:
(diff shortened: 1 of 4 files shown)


giorgiosironi:
Much better now from all locations:

projects/elife.yaml
```diff
@@ -370,9 +370,11 @@ api-gateway:
                 gcslogging:
                     bucket: end2end-elife-fastly
                     path: api-gateway--test/
-                    period: 1800
+                    period: 600
                 healthcheck:
                     path: /ping-fastly
+                    check-interval: 30000
+                    timeout: 10000
```
thewilkybarkid: 10 seconds seems a bit long, looks like it was 0.5 before?

giorgiosironi: can be tuned, this is a large precautionary value that makes the POPs from around the world happy

giorgiosironi:
Just applied, https://synthetics.newrelic.com/accounts/1451451/monitors/375649ae-c489-4ea6-8ddf-bbe7f30f9a2b will tell us the result.

giorgiosironi:
New Relic shows no failures in the last 3 hours, so we can merge this.

giorgiosironi:
For the record, the slowest result is from Mumbai with 1.67 seconds for the NewRelic->FastlyPOP->AWS path.

PR #313, opened 2018-05-08
ELPP-3623 Integrate Vault secrets in builder

giorgiosironi:
Terraform is able to use Vault through, of course, the Vault provider. Authentication to Vault must be dealt with by the user, similarly to the case of AWS credentials; however, Vault should then start to provide secrets so that no other logins are needed.

- Integrate a Vault data resource into Fastly GCS logging implementation
- Add Vault configuration to projects (`address`)
- Add documentation for logging in with Vault
- Add `vault` to prerequisites

### DIFF
--- a/src/tests/fixtures/dummy1-project.json
+++ b/src/tests/fixtures/dummy1-project.json
@@ -22,6 +22,9 @@
         }, 
         "type": "t2.micro", 
         "region": "us-east-1", 
+        "vault": {
+            "address": "https://vault.example.org:8200"
+        },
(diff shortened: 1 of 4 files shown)


giorgiosironi:
Alfred's missing a `vault` client, I'll exclude it from checks as test should not use it.

giorgiosironi:
(not sure if it will pass all tests yet)

lsh-0:
my only quibble were the docs, so I opened a PR: https://github.com/elifesciences/builder/pull/314

PR #314, opened 2018-05-09
tweaks to vault docs

lsh-0:


src/vault.py
```diff
@@ -0,0 +1,41 @@
+from fabric.api import task, local
```
giorgiosironi: ah, the `ssh` style

src/vault.py
```diff
@@ -0,0 +1,41 @@
...
+    local(cmd)
+
+@task
+def lookup(token):
+    cmd = "VAULT_ADDR=%s VAULT_TOKEN=%s vault token lookup" % (vault_addr(), token)
+    local(cmd)
+
+@task
+def create_token():
+    fname = utils.get_input('first name: ')
```
giorgiosironi: a display name seems to have no preferred structure, so I just stuck to our NameSurname convention but this could be used also for applications e.g. `bot` or `search` or `journal`. I'd leave a single custom value as the input.

I'd also allow for the parameter to be passed in as a non-interactive argument, to remain automation-friendly.

src/buildercore/project/__init__.py
```diff
@@ -66,6 +66,16 @@ def find_project(project_location_triple):
...
     return fnmap[protocol](path, hostname)
 
+def raw_project_map(project_locations_list=None):
+    "returns an unprocessed list of maps of project data"
+    if not project_locations_list:
+        project_locations_list = config.app()['project-locations']
+
+    struct = {files.project_file_name(path): files.all_projects(path) for _, _, path in project_locations_list}
+    if len(struct) == 1:
+        return struct.values()[0]
```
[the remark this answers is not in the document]
lsh-0: it is a bit iffy, but we came to the same conclusion: cross the issue of multiple project files when we come to it.

docs/vault.md
```diff
@@ -8,39 +8,28 @@ builder assumes a logged-in Vault client, probably running on the master server.
...
+Terraform, wrapped by builder, will read the ~/.vault-token file during its operations.
 
 ## Creating new tokens
 
 After logging in with a root token, run:
 
-```
-VAULT_ADDR=<ADDRESS> vault token create -policy=builder-user -display-name=NameSurname
-```
+`./bldr vault.create_token`
```
giorgiosironi: Being this a layer of indirection upon the binary `vault`, I'd stick to the convention it has introduced: `vault.token.create` (or `vault.token_create` if that's not possible)

lsh-0:
(grr, you've broken my `.lint.sh` script)

giorgiosironi:
> broken my .lint.sh script

I run this often, when in the virtualenv, without particular problems. What's broken?

src/vault.py
```diff
@@ -0,0 +1,39 @@
...
+    cmd = "VAULT_ADDR=%s VAULT_TOKEN=%s vault token lookup" % (vault_addr(), token)
+    local(cmd)
+
+@task
+def token_create():
+    token = utils.get_input('token: ')
+    if not token or not token.strip():
+        print("a token name is required")
+        sys.exit(1)
+    cmd = "VAULT_ADDR=%s vault token create -policy=builder-user -display-name=%s" % (vault_addr(), token)
```
giorgiosironi: `builder-user` is a magic value that eLife's Vault has defined, should probably be a constant. I think we'll later push it up to the configurations, even

lsh-0:
I use ./.lint.sh often across all the python projects, but with builder it's behaviour has changed.  I run it before commits and sometimes before testing typically, but now it dies if the repo has changes in it and prints a diff.

The files prefixed with a `.` are supposed to be relatively atomic and assume a virtualenv.

giorgiosironi:
That should improve this behavior?

giorgiosironi:
Added some tweaks to merge this.

PR #315, opened 2018-05-09
fixes a bug in lifecycle.py where an rds resource is returned instead…

lsh-0:
… of the expected client

### DIFF
--- a/src/buildercore/lifecycle.py
+++ b/src/buildercore/lifecycle.py
@@ -282,7 +282,7 @@ def _ec2_connection(stackname):
     return boto_conn(stackname, 'ec2')
 
 def _rds_connection(stackname):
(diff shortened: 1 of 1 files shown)


giorgiosironi:
This is the problem in https://alfred.elifesciences.org/job/process-stop-stacks-long-term/ ?

lsh-0:
looks like it

lsh-0:
I suppose it's not covered in any test because of attached rds instance

giorgiosironi:
It takes a long time (>10 minutes sometimes) to stop one of those instances, making automated tests prohibitively slow.

giorgiosironi:
Tested in https://alfred.elifesciences.org/job/process-stop-stacks-long-term/60/console which seems to be running fine for now.

PR #319, opened 2018-05-14
ELPP-3598 Setup harcoded response object and condition for each error code

giorgiosironi:
Contrived example at https://test--cdn-gateway.elifesciences.org/not-found where the sample gateway (which shouldn't even return non-JSON responses) returns the HTML error page.

Implemented using https://www.terraform.io/docs/providers/http/data_source.html to make Terraform load the URLs and dynamically pass them in.

### DIFF
--- a/src/buildercore/terraform.py
+++ b/src/buildercore/terraform.py
@@ -66,6 +66,10 @@ FASTLY_LOG_LINE_PREFIX = 'blank' # no prefix
 # https://github.com/terraform-providers/terraform-provider-fastly/issues/7 tracks when snippets could become available in Terraform
 FASTLY_MAIN_VCL_KEY = 'main'
 
+FASTLY_ERRORS_RESPONSE = {
+    404: 'Not Found',
+    503: 'Service Unavailable',
+}
 
 def render(context):
     if not context['fastly']:
@@ -137,9 +141,9 @@ def render(context):
             response_objects.append({
                 'name': 'error-%s' % code,
                 'status': int(code),
-                # TODO: replace
+                'response': FASTLY_ERRORS_RESPONSE[int(code)],
                 'content': '${data.http.error-page-%s.body}' % code,
-                'content_type': 'text/html; charset=iso-8859-1',
+                'content_type': 'text/html; charset=us-ascii',
                 'cache_condition': cache_condition['name'],
             })
             data[DATA_TYPE_HTTP]['error-page-%d' % code] = {
--- a/src/buildercore/terraform.py
+++ b/src/buildercore/terraform.py
@@ -106,6 +106,7 @@ def render(context):
                 }
             }
         },
+        'data': {},
     }
 
     if context['fastly']['healthcheck']:
@@ -122,6 +123,7 @@ def render(context):
         errors = context['fastly']['errors']
         response_objects = []
         cache_conditions = []
+        tf_file['data']['http'] = {}
         for code, path in errors['codes'].items():
             cache_condition = {
                 'name': 'condition-%s' % code,
@@ -133,9 +135,12 @@ def render(context):
                 'name': 'error-%s' % code,
                 'status': int(code),
                 # TODO: replace
-                'content': 'Error HTML for %s' % code,
+                'content': '${data.http.error-page-%s.body}' % code,
                 'cache_condition': cache_condition['name'],
             })
+            tf_file['data']['http']['error-page-%d' % code] = {
+                'url': '%s%s' % (errors['url'], path),
+            }
         tf_file['resource'][RESOURCE_TYPE_FASTLY][RESOURCE_NAME_FASTLY]['response_object'] = response_objects
         tf_file['resource'][RESOURCE_TYPE_FASTLY][RESOURCE_NAME_FASTLY]['condition'] = cache_conditions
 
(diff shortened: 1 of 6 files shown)


giorgiosironi:
```
* fastly_service_v1.fastly-cdn: [ERR] Invalid configuration for Fastly Service (7KLc7eHd4LZ406ckBryHw): ERROR: Non-ASCII characters found:
(input Line 417 Pos 49)
    <p>Something has gone seriously wrong.<br>We’ve logged the problem and will look into it.</p>
```
:eyeglasses: looking for them

giorgiosironi:
The apostrophe in `We’ve` is a [right single quotation mark](https://www.fileformat.info/info/unicode/char/2019/index.htm). It is still valid UTF-8 so we may be missing some charset/encoding configuration at the Fastly level.

giorgiosironi:
https://www.fastly.com/blog/how-solve-anything-vcl-part-2-so-routing-and-non-ascii says we need to do some encoding of non-ASCII characters into HTML entities, if they end up in VCL:
```
synthetic {"Responsive
Nikkei&#12450;&#12523;&#12501;&#12449;&#12503;&#12525;&#12464;&#
12521;&#12512;&#12398;&#12513;&#12531;&#12496;&#12540;&#12398;&#
30342;&#27096;&#12289;&#12450;&#12523;&#12501;&#12449;&#12496;&#
12540;&#12472;&#12519;&#12531;&#12398;&#12469;&#12452;&#12488;&#
12395;&#12450;&#12463;&#12475;&#12473;&#12391;&#12365;&#12394;&#
12356;&#22580;&#21512;&#12289;rnfeedback@nex.nikkei.co.jp
&#12414;&#12391;&#12381;&#12398;&#26088;&#36899;&#32097;&#12367;
&#12384;&#12373;&#12356;&#12290;"};
```

giorgiosironi:
Since we don't want to reimplement Terraform pulling down HTML responses to encode them, changed 3 characters in https://github.com/elifesciences/pattern-library/pull/866

giorgiosironi:
Working only with the 404 example for now as it's the only one without non-ASCII characters:
https://test--cdn-gateway.elifesciences.org/does-not-exist
We want only 503 in reality (according to what we do with CloudFront now).

thewilkybarkid:
> We want only 503 in reality (according to what we do with CloudFront now).

With the API we'll want it to turn invalid errors (like https://api.elifesciences.org/articles/foo/bar and https://api.elifesciences.org/blog-articles/foo/bar) into proper errors.

thewilkybarkid:
```

**Comment by thewilkybarkid** [comment]
```
set resp.status = 404;
set resp.response = "OK";
```

🤔

File: src/tests/test_buildercore_terraform.py [hunk 1]
```diff
@@ -146,6 +153,22 @@ def test_fastly_template_complex(self):
...
+                                    'statement': 'beresp.status == 503',
+                                    'type': 'CACHE',
+                                },
+                            ],
+                            'response_object': [
+                                {
+                                    'name': 'error-503',
+                                    'status': 503,
+                                    'content': '${data.http.error-page-503.body}',
+                                    'content_type': 'text/html; charset=iso-8859-1',
```
thewilkybarkid: `us-ascii`?

thewilkybarkid:
> Since we don't want to reimplement Terraform pulling down HTML responses to encode them

Could we update Terraform to do this?

thewilkybarkid:
Thought: how would you do non-ASCII in non-HTML? Would you have to `base64` it?

giorgiosironi:
```
set resp.response = "OK";
```
I think that's the default and has to be specified explicitly for each status code. Not sure if the browser actually sees it

src/buildercore/terraform.py
```diff
@@ -118,6 +126,32 @@ def render(context):
...
+            cache_condition = {
+                'name': 'condition-%s' % code,
+                'statement': 'beresp.status == %d' % code,
+                'type': 'CACHE',
+            }
+            cache_conditions.append(cache_condition)
+            response_objects.append({
+                'name': 'error-%s' % code,
+                'status': int(code),
+                'response': FASTLY_ERRORS_RESPONSE[int(code)],
```
thewilkybarkid: `http.client.responses[int(code)]`?

giorgiosironi:
> though you meant the Fastly provider, but see it's Terraform itself, so probably a no-go

What would be needed is a PHP-style `htmlentities` built-in function in https://www.terraform.io/docs/configuration/interpolation.html but seeing no way to load it from the user side (would have to be written in Go and statically linked in, I guess)

src/buildercore/terraform.py
giorgiosironi: Python-3 specific, but I'm looking for a portable alternative

giorgiosironi:
This PR contains support for `conditions`, which will be used in https://elifesciences.atlassian.net/browse/ELPP-3522 hence the need to merge it to allow multiple backends (S3 buckets). Implementation can always be revisited, no production service affected.

src/buildercore/terraform.py
```diff
@@ -63,6 +66,8 @@
 # https://github.com/terraform-providers/terraform-provider-fastly/issues/7 tracks when snippets could become available in Terraform
 FASTLY_MAIN_VCL_KEY = 'main'
 
+# e.g. 404: 'Not Found'
+FASTLY_RESPONSE_REASON_PHRASES = http_responses()
```
thewilkybarkid: Is `FASTLY_RESPONSE_REASON_PHRASES` really needed?

giorgiosironi: It's a little abstraction layer but I'll put a docstring on the original function instead

giorgiosironi:
Merging as-is and rolling out configuration of error-pages separately. Actually, we don't have error-pages to configure because one CDN is serving JSON and the other images. But we should still remove the one of `api-gateway--test`.

PR #320, opened 2018-05-15
Use terraform plan to produce a diff of what is going to happen

giorgiosironi:
At first, we tried using a computed difference of the `.tf` files as a stopgap, but it falls short of different people running `update_infrastructure` on different laptops, as they have no way to get at the old `.tf` files.

Terraform has first-class support for generating a diff of specification and actual infrastructure, so let's use `terraform plan` to do it. This will take into account the actual state of remote resources rather than just the local files.

### DIFF
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -403,14 +403,16 @@ EC2_NOT_UPDATABLE_PROPERTIES = ['ImageId', 'Tags', 'UserData']
 # * what to add
 # * what to modify
 # * what to remove
-# What we see here is the new Terraform generated.tf file, containing all resources (just Fastly so far).
-# We can do a diff with the current one which would already be an improvement, but ultimately the source of truth
-# is changing it and running a terraform plan to see proposed changes. We should however roll it back if the user
-# doesn't confirm.
 class Delta(namedtuple('Delta', ['plus', 'edit', 'minus', 'terraform'])):
-    """represents a delta between and old and new CloudFormation generated template, showing which resources are being added, updated, or removed
+    @classmethod
+    def from_cloudformation_and_terraform(cls, cloud_formation_delta, terraform_delta):
+        return cls(
+            cloud_formation_delta.plus,
+            cloud_formation_delta.edit,
+            cloud_formation_delta.minus,
+            terraform_delta
+        )
 
-    Extends the namedtuple-generated class to add custom methods."""
     @property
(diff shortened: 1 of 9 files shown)


src/buildercore/cloudformation.py
```diff
@@ -20,3 +21,29 @@ def validate_template(pname, rendered_template):
     "remote cloudformation template checks."
     conn = core.boto_conn(pname, 'cloudformation', client=True)
     return conn.validate_template(TemplateBody=rendered_template)
+
+# CloudFormation is nicely chopped up into:
```
giorgiosironi: just moving things progressively here from `cfngen.py`, no code change apart from the class name

src/buildercore/cfngen.py
```diff
@@ -404,14 +403,16 @@ def generate_stack(pname, **more_context):
...
 # * what to modify
 # * what to remove
-# What we see here is the new Terraform generated.tf file, containing all resources (just Fastly so far).
-# We can do a diff with the current one which would already be an improvement, but ultimately the source of truth
-# is changing it and running a terraform plan to see proposed changes. We should however roll it back if the user
-# doesn't confirm.
-"""represents a delta between and old and new CloudFormation generated template, showing which resources are being added, updated, or removed
-
-Extends the namedtuple-generated class to add custom methods."""
 class Delta(namedtuple('Delta', ['plus', 'edit', 'minus', 'terraform'])):
```
giorgiosironi: still has to be a 4-ple for backward compatibility

src/buildercore/terraform.py
```diff
@@ -194,6 +196,34 @@ def read_template(stackname):
...
+
+def generate_delta(context, new_template):
+    # simplification: unless Fastly is involved, the TerraformDelta will be empty
+    # this should eventually be removed, for example after test_buildercore_cfngen tests have been ported to test_buildercore_cloudformation
+    if not context['fastly']:
+        return None
+
+    write_template(context['stackname'], new_template)
+    terraform = init(context['stackname'], context)
+    terraform.plan(input=False, no_color=IsFlagged, capture_output=False, raise_on_error=True, detailed_exitcode=IsNotFlagged, out='out.plan')
```
giorgiosironi: the fair of random command line arguments

giorgiosironi:
One of the integration tests was failing because of transient states, had to assert less to make it reliable.

PR #321, opened 2018-05-17
ELPP-3522 Multiple Fastly backends

giorgiosironi:
Theory: use a default and a custom backend, adding a `request_condition` to select a particular pattern like `articles/` for the custom one.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -732,19 +732,29 @@ generic-cdn:
     intdomain: null
     aws:
         ec2: false
-        cloudfront:
+        #cloudfront:
+        #    subdomains:
+        #        - "{instance}-cdn"
+        #    origins:
+        #        # first is default
+        #        default:
+        #            hostname: elife-cdn.s3.amazonaws.com
+        #        articles:
+        #            hostname: "{instance}-elife-published.s3.amazonaws.com"
+        #            pattern: articles/*
+        #    default-ttl: 86400 # seconds
+        #    headers:
+        #        - Origin
+        fastly:
             subdomains:
-                - "{instance}-cdn"
-            origins:
+                - "{instance}-fastly"
+            backends:
                 # first is default
                 default:
                     hostname: elife-cdn.s3.amazonaws.com
                 articles:
                     hostname: "{instance}-elife-published.s3.amazonaws.com"
-                    pattern: articles/*
-            default-ttl: 86400 # seconds
-            headers:
-                - Origin
+                    condition: req.url ~ "^/articles" 
 
 journal-cms:
     subdomain: journal-cms # journal-cms.elifesciences.org
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -253,12 +253,19 @@ def build_context_cloudfront(context, parameterize):
         context['cloudfront'] = False
 
 def build_context_fastly(context, parameterize):
-    def build_subdomain(x):
+    def _build_subdomain(x):
         return complete_domain(parameterize(x), context['domain'])
+    def _parameterize_hostname(b):
+        b['hostname'] = parameterize(b['hostname'])
+        return b
+
     if 'fastly' in context['project']['aws']:
+        backends = context['project']['aws']['fastly'].get('backends', {})
+        print backends
         context['fastly'] = {
-            'subdomains': [build_subdomain(x) for x in context['project']['aws']['fastly']['subdomains']],
-            'subdomains-without-dns': [build_subdomain(x) for x in context['project']['aws']['fastly']['subdomains-without-dns']],
+            'backends': {n: _parameterize_hostname(b) for n, b in backends.items()},
+            'subdomains': [_build_subdomain(x) for x in context['project']['aws']['fastly']['subdomains']],
+            'subdomains-without-dns': [_build_subdomain(x) for x in context['project']['aws']['fastly']['subdomains-without-dns']],
             'dns': context['project']['aws']['fastly']['dns'],
             'healthcheck': context['project']['aws']['fastly']['healthcheck'],
             'errors': context['project']['aws']['fastly']['errors'],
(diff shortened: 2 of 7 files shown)


giorgiosironi:
```
curl -v https://test2-fastly.elifesciences.org/ping.txt
```
returns 
```
<Error><Code>NoSuchBucket</Code><Message>The specified bucket does not exist</Message><BucketName>test2-fastly.elifesciences.org</BucketName><RequestId>15435208C081CCD3</RequestId><HostId>mK/perxnFqpkM38v4nFrAU6KXP4SZEEBPT7/5chYEb3jvLOn5QZV+iL7EHFBEm97pmvvtYN1sSU=</HostId></Error>
```
which means we are incorrectly specifying the backend's `Host`. AWS servers need to know which bucket to serve.

giorgiosironi:
That bucket name is not in the VCL, meaning it's probably just forwarding `Host`:
https://manage.fastly.com/configure/services/1rf3CiPgOFRTFKlcoP9HiS/versions/1/vcl

giorgiosironi:
Best solution seems to set up a `header` with the same `request_condition`:
https://www.terraform.io/docs/providers/fastly/r/service_v1.html#default_host-1

ping
```diff
@@ -0,0 +1 @@
+pong
```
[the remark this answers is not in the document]
giorgiosironi: I accidentally https://cdn.elifesciences.org/ping.txt

giorgiosironi:
`settings.default_host` is the best option now, with the caveat of using N of them, with N-1 having a `request_condition` to select them. Avoids adding a hand-rolled `Host` header.

giorgiosironi:
Applying to `generic-cdn--continuumtest` to test their equivalence on a real CDN. There were some pending microconfiguration changes to it, so it will take the usual tens of minutes to update the CloudFront part before the Fastly one can be created.

giorgiosironi:
Ok, compare:
https://continuumtest-cdn.elifesciences.org/ping.txt (`elife-cdn` bucket)
https://continuumtest-fastly.elifesciences.org/ping.txt (`elife-cdn` bucket)
https://continuumtest-cdn.elifesciences.org/articles/00666/elife-00666-fig1-v10.jpg (`continuumtest-elife-published` bucket)
https://continuumtest-fastly.elifesciences.org/articles/00666/elife-00666-fig1-v10.jpg (`continuumtest-elife-published` bucket)

giorgiosironi:
Tests pass again, on Python 2 at least, therefore ready for review. Applied to `end2end` and `continuumtest`.

projects/elife.yaml
```diff
@@ -745,6 +746,17 @@ generic-cdn:
...
+        fastly:
+            subdomains:
+                - "{instance}-fastly"
+            backends:
+                # first is default
+                default:
+                    hostname: elife-cdn.s3.amazonaws.com
+                articles:
+                    hostname: "{instance}-elife-published.s3.amazonaws.com"
+                    condition: req.url ~ "^/articles"
```
thewilkybarkid: Should this have a trailing slash?

giorgiosironi: Can be added, works but can be made more specific. It's pure VCL

giorgiosironi:
This is ready for review and has the `generic-cdn` instances provisioning Fastly with a different DNS. Follow-up will be the invalidation.

projects/elife.yaml
```diff
@@ -745,6 +747,57 @@ generic-cdn:
...
+                articles:
+                    hostname: "{instance}-elife-published.s3.amazonaws.com"
+                    condition: req.url ~ "^/articles/"
+            default-ttl: 86400 # seconds
+            surrogate-keys:
+                article-id:
+                    url: "^/articles/([0-9]+)/(.+)$"
+                    value: "articles/\\1"
+    aws-alt:
+        # have to specify this because otherwise configuration such as `fresh` inherited from defaults won't see the ec2: false in `aws:` and test_validation.py will fail for all of them?
```
lsh-0: urgh. at some point we should revisit the rough edges of project wrangling.

src/buildercore/cfngen.py
```diff
@@ -253,18 +253,27 @@ def build_subdomain(x):
...
+    def _parameterize_hostname(b):
+        b['hostname'] = parameterize(b['hostname'])
+        return b
+
     if 'fastly' in context['project']['aws']:
+        backends = context['project']['aws']['fastly'].get('backends', OrderedDict({}))
         context['fastly'] = {
-            'subdomains': [build_subdomain(x) for x in context['project']['aws']['fastly']['subdomains']],
-            'subdomains-without-dns': [build_subdomain(x) for x in context['project']['aws']['fastly']['subdomains-without-dns']],
+            'backends': OrderedDict([(n, _parameterize_hostname(b)) for n, b in backends.items()]),
```
lsh-0: `.items()` won't have a guaranteed order (unless `backends` is also an `OrderedDict`?)

projects/elife.yaml
```diff
@@ -745,6 +747,57 @@ generic-cdn:
...
+                default:
+                    hostname: elife-cdn.s3.amazonaws.com
+                articles:
+                    hostname: "{instance}-elife-published.s3.amazonaws.com"
+                    condition: req.url ~ "^/articles/"
+            default-ttl: 86400 # seconds
+            surrogate-keys:
+                article-id:
+                    url: "^/articles/([0-9]+)/(.+)$"
+                    value: "articles/\\1"
```
thewilkybarkid: Should use `article/12345`, as we use that format elsewhere. Means it's a little different to the original URIs though (here, and our API).

giorgiosironi: Where? Here I ported the original CDN paths but they seemed in line with https://elifesciences.org/articles/10627

projects/elife.yaml
giorgiosironi: Yes, in this specific case maybe the inherited `defaults` could be merged with the project's `aws`, but who knows what that could break.

src/buildercore/cfngen.py
giorgiosironi: It is https://github.com/elifesciences/builder/pull/321/files/de7dc59419d52adeb6daa9006fdf396f2c31571e#diff-11aa22c619b41fcc4ac5151f83decddcR264
Was causing a random test failure, which then was resolved

PR #324, opened 2018-05-23
Fix broken db 404

lsh-0:


src/buildercore/core.py
```diff
@@ -218,15 +218,13 @@ def find_rds_instances(stackname, state='available'):
 
     except botocore.exceptions.ClientError as err:
         LOG.info(err.response)
-        msg = err.response['Error']['Message']
         invalid_dbid = "Invalid database identifier"
```
giorgiosironi: could be renamed to `invalid_dbid_message_prefix` or similar, since the value of this variable is not a db ID

lsh-0:
> Fix is clear enough, would be interesting to know which task ran into the problem

I tried using my newly merged `restart` feature and it broke :(

it turns out you have all your calls to the same function wrapped in a check I didn't feel was necessary. If it doesn't find any RDS instances, it should return an empty structure - *not* vomit and die.

PR #326, opened 2018-05-24
Integration test stability

giorgiosironi:
- Switch in test settings in recent integration tests. Allows them to run in isolation, found out after a failure of `test_validation` made them all fail because it wasn't switching back to the test settings
- Separate `TestValidationElife` to its own class
- `TestValidationElife` runs 43 separate tests rather than a single one
- `TestValidationElife` can filter one of the 43 separate tests rather than always running all of them

Last two features require `pytest`, but greatly help in debugging these cases.

Started from a build where there was one single project problem (`generic-cdn` `aws-alt` validation failures), and that produced this test result:

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -46,3 +46,11 @@ or:
 or:
 
     PYTHONPATH=src green tests.test_buildercore_trop.TestBuildercoreTrop.test_method_name
+
+## Filter single integration tests
+
+Some integration tests run through all the defined projects for regression testing. To filter only a single project, for example to reproduce a failure, you can run:
+
+```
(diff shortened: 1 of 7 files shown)


lsh-0:
I've been eyeing that area of code for a while. It was introduced before any familiarity with `unittest.mock.patch`. I'm still suspect of it after your change, perpetuating `settings.CONFIG = ...` just strikes me as wrong, but I'm glad it's had a scrub.

giorgiosironi:
In an ideal world there will be no global `settings` but it would be injected where needed; that would incidentally make testing straightforward because it would be a function/method parameter; a story for another time as it has architectural implications.

lsh-0:
absolutely

PR #327, opened 2018-05-29
ELPP-3624 Support for Google Cloud Platform buckets

giorgiosironi:
We use GCP buckets (Google Cloud Storage) to aggregate access logs from Fastly. To avoid configuration drift, this has to be captured in builder.

Cleans up CloudFormation `bootstrap`, `destroy` and `update_infrastructure` scenarios into its own module. 

Some `TODO` are intentionally left in for now, to avoid a monumental diff. Will be tackled by a separate pull request after this one.


giorgiosironi:
Have to do some local runs of integration tests to make them green, `test_with_instance` was until a couple of commits ago.

PR #328, opened 2018-05-30
ELPP-3665 Define surrogate keys for iiif

giorgiosironi:
Also refine them for generic-cdn. Will allow the bot to invalidate a single article or a single version.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -773,21 +773,21 @@ generic-cdn:
             surrogate-keys:
                 article-assets:
                     url: "^/articles/([0-9]+)/elife-([0-9]+)-([a-z0-9]+)-v([0-9]+)\\.(.+)$"
-                    value: "articles/\\1 article/\\1 article/\\1/version/\\4"
+                    value: "articles/\\1 article/\\1 article/\\1v\\4"
                     samples:
                         article-asset:
                             path: /articles/10627/elife-10627-fig1-v1.tif
-                            expected: "articles/10627 article/10627 article/10627/version/1"
+                            expected: "articles/10627 article/10627 article/10627v1"
                         # legacy, not supported: /articles/10627/elife-10627-fig1-v1-download.jpg
                         # legacy, not supported: /articles/10627/elife-10627-fig1-v1-972w.jpg
 
                 article-pdf-xml:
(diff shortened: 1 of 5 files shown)


projects/elife.yaml
```diff
@@ -773,7 +773,7 @@ generic-cdn:
             surrogate-keys:
                 article-id:
                     url: "^/articles/([0-9]+)/(.+)$"
-                    value: "articles/\\1"
+                    value: "articles/\\1 article/\\1"
```
giorgiosironi: Will define `version/...` here too, but there are file like `/articles/10627/elife-10627-v1-download.pdf` that makes it a bit more convoluted. Are those still used?

projects/elife.yaml
```diff
@@ -1536,6 +1536,17 @@ iiif:
...
+            subdomains:
+                - "{instance}--cdn-iiif"
+            healthcheck:
+                path: /ping-fastly
+                check-interval: 30000
+                timeout: 2000
+            surrogate-keys:
+                article-id-version:
+                    url: "^/lax:([0-9]+)/elife-([0-9]+)-([a-z0-9]+)-v([0-9]+)\\.(.+)$"
+                    value: "article/\\1 article/\\1/version/\\4"
```
thewilkybarkid: `article/1234v1` or `article/1234/v1` maybe?

projects/elife.yaml
thewilkybarkid: The `-download` files aren't, as they're stream through Journal instead.

src/buildercore/cfngen.py
```diff
@@ -260,7 +260,7 @@ def _parameterize_hostname(b):
         b['hostname'] = parameterize(b['hostname'])
         return b
 
-    if 'fastly' in context['project']['aws']:
+    if 'fastly' in context['project']['aws'] and context['project']['aws']['fastly']:
```
thewilkybarkid: Could be simplified with a `.get()`?

giorgiosironi: Yes, it should

giorgiosironi:
Applying everywhere while the build runs.

giorgiosironi:
And, purging all from `generic-cdn` and `iiif` to allow for the new surrogate keys to be used.

PR #331, opened 2018-05-31
ELPP-3522 Expand logs to investigate 400 errors

giorgiosironi:
- `Host` has been added, but it has already been overwritten by the Fastly VCL
- `X-Forwarded-For` shows nothing strange so far, but more monitoring is needed to get a 400 while this configuration is on


giorgiosironi:
Applied to `generic-cdn--continuumtest` and `generic-cdn--prod`.

giorgiosironi:
Retaining this for now.

PR #332, opened 2018-05-31
ELPP-3665 Expand surrogate keys

giorgiosironi:
For iiif and generic-cdn with bot test cases. These are automatically tested when provisioning since they feature an `expected` field.


gnott:
Thanks, I added another figure example, and I expanded the video matching. I didn't realise the tests take so long otherwise maybe I would have done them in a single commit. Hopefully they all pass.

giorgiosironi:
Thank you, will get this applied everywhere and unblock the bot's invalidations.

giorgiosironi:
Applied all, purged all, merging on green.

PR #333, opened 2018-06-01
Return 200 to Office making WebDAV requests

thewilkybarkid:
As mentioned in ELPP-3522.

Not tested anywhere yet.

PR #334, opened 2018-06-01
ELPP-3624 Log original hostname

giorgiosironi:
`cdn.elifesciences.org` rather than the backend it is pointing to.

Waiting on https://github.com/elifesciences/builder/pull/333#pullrequestreview-125162931 to avoid clashes

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -364,6 +364,10 @@ api-gateway:
             vcl:
                 - "original-host"
                 - "gzip-by-content-type-suffix"
+            gcslogging:
+                bucket: "{instance}-elife-fastly"
+                path: "api-gateway--{instance}/"
+                period: 600
     aws-alt:
         standalone:
             # for now a copy of standalone1404
@@ -382,10 +386,6 @@ api-gateway:
                     - api
         test:
             fastly:
-                gcslogging:
-                    bucket: end2end-elife-fastly
-                    path: api-gateway--test/
-                    period: 600
                 errors:
                     url: https://prod-elife-error-pages.s3.amazonaws.com/
                     codes:
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1537,6 +1537,10 @@ iiif:
         fastly:
             subdomains:
                 - "{instance}--cdn-iiif"
+            gcslogging:
+                bucket: "{instance}-elife-fastly"
+                path: "iiif--{instance}/"
+                period: 600
             healthcheck:
                 path: /ping-fastly
                 check-interval: 30000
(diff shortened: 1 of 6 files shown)


giorgiosironi:
Timeout in this stack trace:
```
src/buildercore/bootstrap.py:80: in create_stack
setup_ec2(stackname, context)
src/buildercore/context_handler.py:73: in decorated_with_only_if
return fn(stackname, context, **kwargs)
src/buildercore/bootstrap.py:107: in setup_ec2
stack_all_ec2_nodes(stackname, _setup_ec2_node, username=BOOTSTRAP_USER)
src/buildercore/core.py:334: in stack_all_ec2_nodes
return parallel_work(single_node_work, params)
src/buildercore/core.py:347: in parallel_work
return execute(parallel(single_node_work), hosts=list(params['public_ips'].values()))
.tox/py35/lib/python3.5/site-packages/fabric/tasks.py:419: in execute
error(err, exception=d['results'])
.tox/py35/lib/python3.5/site-packages/fabric/utils.py:359: in error
return func(message)
```
Can probably insert some retry along the way...

giorgiosironi:
Happened in other builds too, so `create_stack` may have lost some robustness during refactoring.

giorgiosironi:
It starts when running with `docker-compose up article-store`, but `docker-compose up -V` reproduces the problem for me.

giorgiosironi:
Our pattern of including VCL snippets _after_ Fastly macros is not enough here: we need to include this before `FASTLY_RECV`:
```
sub vcl_recv {
#--FASTLY RECV BEGIN
  if (req.restarts == 0) {
    if (!req.http.X-Timer) {
      set req.http.X-Timer = "S" time.start.sec "." time.start.usec_frac;
    }
    set req.http.X-Timer = req.http.X-Timer ",VS0";
  }
  # default conditions
  set req.backend = F_iiif__prod;
  if (!req.http.Fastly-SSL) {
     error 801 "Force SSL";
  }
  set req.hash_ignore_busy = false;
  set req.hash_always_miss = false;
  set req.http.host = "prod--iiif.elifesciences.org";
    # end default conditions
#--FASTLY RECV END
  // BEGIN builder original-host
set req.http.Elife-Orig-Host = req.http.host;
  // END builder original-host
```

giorgiosironi:
Better now:
```
sub vcl_recv {
  // BEGIN builder original-host
set req.http.Elife-Orig-Host = req.http.host;
  // END builder original-host
#--FASTLY RECV BEGIN
  if (req.restarts == 0) {
      ...
```

giorgiosironi:
Seems to work:
```
$ cat 2018-06-04T14\:40\:00.000-* | jq .original_host | sort -n | uniq -c
    196 "iiif.elifesciences.org"
   1234 "prod--cdn-iiif.elifesciences.org"
    653 "prod--iiif.elifesciences.org" # old incorrect value, from log lines before the change in configuration
```

giorgiosironi:
Applied everywhere.

PR #335, opened 2018-06-01
ELPP-3624 CloudFormation vs Terraform follow up

giorgiosironi:
Strictly separates more of the `update_infrastructure` process into `cloudformation` and `terraform` modules. Carries on the emptying out of `cfngen` from CloudFormation or Terraform concerns. The end result is also that `cfn` becomes simpler as `cloudformation` and `terraform` get uniform interfaces (similar function names and arguments).

### DIFF
--- a/src/integration_tests/test_with_instance.py
+++ b/src/integration_tests/test_with_instance.py
@@ -89,7 +89,7 @@ class One(base.BaseCase):
         bootstrap.setup_ec2(self.stackname, self.context)
 
     # TODO: transform this into a cloudformation module test
-    #def test_bootstrap_update_template_no_updates(self):
+    # def test_bootstrap_update_template_no_updates(self):
     #    "a template with no changes can be updated with no problems"
     #    bootstrap.update_template(self.stackname, json.load(open(self.cfn_template, 'r')))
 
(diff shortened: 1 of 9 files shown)


giorgiosironi:
Been running unit tests regularly but not sure that everything will pass at the first attempt.

src/integration_tests/test_with_instance.py
```diff
@@ -88,9 +88,10 @@ def test_bootstrap_wait_until_in_progress(self):
         cloudformation._wait_until_in_progress(self.stackname)
         bootstrap.setup_ec2(self.stackname, self.context)
 
-    def test_bootstrap_update_template_no_updates(self):
-        "a template with no changes can be updated with no problems"
-        bootstrap.update_template(self.stackname, json.load(open(self.cfn_template, 'r')))
+    # TODO: transform this into a cloudformation module test
```
giorgiosironi: about to add this

giorgiosironi:
Not sure if this will be generalizable, but the GCS buckets were already existing, hence I used this pattern:
```
bldr launch:fastly-logs,continuumtest
# generates all terraform stuff, but fails on terraform apply

cd .cfn/terraform/fastly-logs--continuumtest
terraform import google_storage_bucket.continuumtest-elife-fastly continuumtest-elife-fastly
cd -

bldr update_infrastructure:fastly-logs--continuumtest
# passes with 0 changes to make
```

lsh-0:
that's clever, well done

PR #336, opened 2018-06-04
ELPP-3624 Remove need for FASTLY_API_KEY environment variable

giorgiosironi:
A valid key is provided by Vault instead. For example, @thewilkybarkid or other builder users can make changes to the Fastly CDNs through builder without generating API tokens on the Fastly panel and maintaining them on their laptop. The only requirement is to log in into Vault, which consolidates these credentials.

To put this into context, the ideal thing would be for Vault to have a secret engine that is able to generate Fastly API keys on the fly, whenever needed. Apart from the complexity, [such an engine is not provided](https://www.vaultproject.io/docs/secrets/index.html) so we default to a static, shared secret that is kept in Vault and is temporarily downloaded by users when running `update_infrastructure`.

### DIFF
--- a/src/buildercore/terraform.py
+++ b/src/buildercore/terraform.py
@@ -22,6 +22,12 @@ DATA_TYPE_HTTP = 'http'
 DATA_NAME_VAULT_GCS_LOGGING = 'fastly-gcs-logging'
 DATA_NAME_VAULT_FASTLY_API_KEY = 'fastly'
 
+# keys to lookup in Vault
+# cannot modify these without putting new values inside Vault:
+#     VAULT_ADDR=https://...:8200 vault put secret/builder/apikey/fastly-gcs-logging email=... secret_key=@~/file.json
+VAULT_PATH_FASTLY = 'secret/builder/apikey/fastly'
+VAULT_PATH_FASTLY_GCS_LOGGING = 'secret/builder/apikey/fastly-gcs-logging'
+
 FASTLY_GZIP_TYPES = ['text/html', 'application/x-javascript', 'text/css', 'application/javascript',
                      'text/javascript', 'application/json', 'application/vnd.ms-fontobject',
                      'application/x-font-opentype', 'application/x-font-truetype',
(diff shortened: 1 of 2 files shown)


src/buildercore/terraform.py
```diff
@@ -20,6 +20,13 @@
 DATA_TYPE_VAULT_GENERIC_SECRET = 'vault_generic_secret'
 DATA_TYPE_HTTP = 'http'
 DATA_NAME_VAULT_GCS_LOGGING = 'fastly-gcs-logging'
+DATA_NAME_VAULT_FASTLY_API_KEY = 'fastly'
+
+# keys to lookup in Vault
+# cannot modify these without putting new values inside Vault:
+#     VAULT_ADDR=https://...:8200 vault put secret/builder/apikey/fastly-gcs-logging email=... secret_key=@~/file.json
+VAULT_PATH_FASTLY = 'secret/builder/apikey/fastly'
+VAULT_PATH_FASTLY_GCS_LOGGING = 'secret/builder/apikey/fastly-gcs-logging'
```
giorgiosironi: pattern:
- `secret/builder/` contains everything builder uses
- `apikey` is all the API keys for external services (chose that name some time ago)
- `fastly` or `fastly-gcs-logging` is the name for those credentials. Usually every value is a dictionary

src/buildercore/terraform.py
```diff
@@ -401,6 +408,7 @@ def init(stackname, context):
                 'fastly': {
                     # exact version constraint
                     'version': "= %s" % PROVIDER_FASTLY_VERSION,
+                    'api_key': "${data.%s.%s.data[\"api_key\"]}" % (DATA_TYPE_VAULT_GENERIC_SECRET, DATA_NAME_VAULT_FASTLY_API_KEY),
```
lsh-0: is this jsonp?

giorgiosironi: Terraform "Interpolation Syntax" https://www.terraform.io/docs/configuration/interpolation.html#attributes-of-a-data-source, the underlying data structure is [HCL](https://www.terraform.io/docs/configuration/syntax.html) which is JSON written with a different syntax.

lsh-0: ah, I should have guessed. That was the only drawback to using Terraform for me, learning this weird custom json compatible format.

diversemix: TBF: That is a common complaint about the json ... there are people I know who have written tools around transforming other more clearer formats to this format.

PR #337, opened 2018-06-05
ELPP-3633 Run `vagrant up` using a container for the Python part

giorgiosironi:
`vagrant up` runs on the host, but delegates reading project data to `.project.py`, which we can execute inside a container. The image derived from `python:2.7` allows Mac users not to install Python for this use case.

## To test

1. Install `docker` if not already installed
2. Install `vagrant` (and `VirtualBox`) if not already installed (can't easily run in Docker, I think)
3. Run `vagrant up`
4. Choose `schematron-validator` which was our candidate example
5. See the whole provisioning succeed


lsh-0:
Vagrant shouldn't depend on Docker, I would switch the executable inside Vagrant based on the presence of flag (`.use-docker.flag`).

giorgiosironi:
There is also the `.use-python-3` flag, I'll add both.

giorgiosironi:
Alfred doesn't have Docker available, at least in its master node.

Vagrantfile
```diff
@@ -25,8 +25,14 @@ def runcmd(cmd)
 end
 
 def project_cmd(argstr)
-    cmd = "/bin/bash -c \"source venv/bin/activate && ./.project.py #{argstr}\""
-    #prn(cmd)
+    flag = ".use-docker.flag"
+    if File.exists?(flag)
```
lsh-0: :+1:

giorgiosironi:
Note: the container is writing `__pycache__` folders and `.pyc` files in the source folder, mounted on it. Need to avoid that.

lsh-0:
made some suggestions here:
https://github.com/elifesciences/builder/pull/348

giorgiosironi:
Finally green.

PR #338, opened 2018-06-05
CDN path for digest images

gnott:
Making a stab at the CDN configuration described in https://github.com/elifesciences/digest-parser/issues/18.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -738,9 +738,9 @@ generic-cdn:
                 articles:
                     hostname: "{instance}-elife-published.s3.amazonaws.com"
                     condition: req.url ~ "^/articles/"
-                digest:
+                digests:
                     hostname: "{instance}-elife-published.s3.amazonaws.com"
-                    condition: req.url ~ "^/digest/"
+                    condition: req.url ~ "^/digests/"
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -738,6 +738,9 @@ generic-cdn:
                 articles:
                     hostname: "{instance}-elife-published.s3.amazonaws.com"
                     condition: req.url ~ "^/articles/"
+                digest:
```
giorgiosironi: to be consistent with `articles` this should be plural, `digests`?

giorgiosironi: not sure yet what the `surrogate-keys` would be without knowing the naming pattern, they can be added later. The default would be that they are cached for 24 hours however.

giorgiosironi:
Amended for consistency, applied to `continuumtest` and placed a strategic file for testing:
https://continuumtest-cdn.elifesciences.org/digests/ping.txt
Ready for @gnott counter-review, then I'll apply to other environments and merge.

giorgiosironi:
Closing as https://github.com/elifesciences/jira-import/issues/4299#issuecomment-402412472 substitutes this with IIIF.

PR #341, opened 2018-06-07
Allow ensure_destroyed idempotence

giorgiosironi:
Streamlines the destroy behavior to be:
- destroy CloudFormation/Terraform
- destroy leftover resources such as DNS
- finally destroy the context, both locally and from S3

The context is the first thing we check for idempotence: if it's
missing, we are sure everything has already been deleted.

https://github.com/elifesciences/journal-formula/pull/62 had some
problems to be solved with this change.

### DIFF
--- a/src/tests/test_buildercore_context_handler.py
+++ b/src/tests/test_buildercore_context_handler.py
@@ -4,7 +4,7 @@ from buildercore import cfngen, context_handler
 
 class TestBuildercoreContext(base.BaseCase):
     def setUp(self):
-        print(context_handler.delete_context_from_s3('dummy1--prod'))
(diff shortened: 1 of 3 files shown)


giorgiosironi:
Have to merge this to allow formula builds to do multiple runs, as otherwise the cleanup from run to run is incomplete.

src/tests/test_buildercore_context_handler.py
```diff
@@ -4,7 +4,7 @@
 
 class TestBuildercoreContext(base.BaseCase):
     def setUp(self):
-        print(context_handler.delete_context_from_s3('dummy1--prod'))
```
lsh-0: :+1:

PR #342, opened 2018-06-07
ELPP-3524 Fastly configuration for elifesciences.org

giorgiosironi:
First stage:

- subdomain mapped with CNAME but support for A is provided (will be needed for 2nd-level domains like `elifesciences.org`)
- health check (hits nginx through the load balancer)
- some VCL to strip all cookies but `journal`, emulating CloudFront's whitelist
- usual `gcslogging`

More will follow on the error-pages, fundamentally they don't work in all use cases now so the implementation has to be rewritten with MoreCustomVCL(TM).

### DIFF
--- a/src/buildercore/fastly/vcl/strip-non-journal-cookies.vcl
+++ b/src/buildercore/fastly/vcl/strip-non-journal-cookies.vcl
@@ -1,0 +1,7 @@
+declare local var.journal_cookie STRING;
+set var.journal_cookie = req.http.Cookie:journal;
+if (var.journal_cookie != "") {
+    set req.http.Cookie = "journal=" var.journal_cookie ";";
+} else {
+    unset req.http.Cookie;
+}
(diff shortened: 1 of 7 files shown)


src/buildercore/fastly/vcl/strip-non-journal-cookies.vcl
```diff
@@ -0,0 +1,7 @@
+declare local var.journal_cookie STRING;
+set var.journal_cookie = req.http.Cookie:journal;
```
thewilkybarkid: The cookie name is a Pillar value. But we don't currently vary it.

giorgiosironi: Was introduced before the possibility of VCL templates was implemented, so the cookie name `journal` could be passed in now - but the behavior is very specific to Journal anyway

thewilkybarkid: For now it's ok; there's a pretty standard VCL snippet for whitelisting one or more cookies (with exact or partial names), which we can always switch to if something else needs it.

PR #343, opened 2018-06-08
ELPP-3524 journal--prod test on Fastly, with new implementation of Fastly error pages

giorgiosironi:
## Test Fastly on 2nd-level domain

We use `elifejournal.org` already added to our Fastly TLS certificates. The DNS is different from the CNAME used for domains such as `cdn.elifesciences.org`, we configure A records instead.

## Error pages

The previous implementation overwrote all 503 errors, but didn't work when the origin was down. With the magic of Terraform and custom VCL we:

1. Download the `error-pages--prod` HTML as a Terraform data source
2. Put it into a template that becomes the VCL snippet `error-page-vcl-503` which is another, derived Terraform data source.
3. Include that template into the main VCL, along with any other VCL snippet we were interested in.

Incredibly, the whole pipeline works and generates a VCL that not only has correct syntax but also served the synthetic response when the origin is unreachable (easily tested by stopping nginx).

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -519,7 +519,7 @@ journal:
                     # always prod for simplicity
                     url: https://prod-elife-error-pages.s3.amazonaws.com/
                     codes:
-                        503: "5xx.html"
+                        503: "503.html"
                 vcl:
                     - "original-host"
                     - "strip-non-journal-cookies"
@@ -590,7 +590,7 @@ journal:
                     # always prod for simplicity
                     url: https://prod-elife-error-pages.s3.amazonaws.com/
                     codes:
-                        503: "5xx.html"
+                        503: "503.html"
                 vcl:
                     - "original-host"
                     - "strip-non-journal-cookies"
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -515,11 +515,11 @@ journal:
                     path: /ping-fastly
                     check-interval: 30000
                     timeout: 2000
-                #errors:
-                #    # always prod for simplicity
-                #    url: https://prod-elife-error-pages.s3.amazonaws.com/
-                #    codes:
-                #        503: "5xx.html"
+                errors:
+                    # always prod for simplicity
+                    url: https://prod-elife-error-pages.s3.amazonaws.com/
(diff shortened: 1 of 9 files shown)


src/buildercore/fastly/vcl/error-page.vcl.tpl
```diff
@@ -0,0 +1,5 @@
+if (obj.status == ${code}) {
```
thewilkybarkid: IIRC `vcl_error` is only for Varnish errors, so won't overwrite the backend returning 503.

src/buildercore/fastly/vcl/error-page.vcl.tpl
```diff
@@ -0,0 +1,5 @@
...
+   set obj.http.Content-Type = "text/html; charset=us-ascii";
```
thewilkybarkid: Doesn't look like Terraform provides response headers, so we have to set this somewhere.

projects/elife.yaml
```diff
@@ -578,6 +578,26 @@ journal:
...
+                    - "elifejournal.org"
+                healthcheck:
+                    path: /ping-fastly
+                    check-interval: 30000
+                    timeout: 2000
+                errors:
+                    # always prod for simplicity
+                    url: https://prod-elife-error-pages.s3.amazonaws.com/
+                    codes:
+                        503: "5xx.html"
```
thewilkybarkid: We have `503.html`

src/buildercore/fastly/vcl/error-page.vcl.tpl
giorgiosironi: Which would be the right behavior, because nginx has custom error pages: https://github.com/elifesciences/journal-formula/blob/master/salt/journal/config/etc-nginx-sites-available-journal.conf#L146

src/buildercore/fastly/vcl/error-page.vcl.tpl
giorgiosironi: `header` is available but depends on Fastly's request/cache/response conditions so it's not as flexible

PR #345, opened 2018-06-11
ELPP-3393 Add ELB to journal--continuumtest

giorgiosironi:
Identical to prod, and removes some headaches for integrating with CDNs in multiple environments: one fewer configuration to deal with

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -501,6 +501,13 @@ journal:
                 engine: redis
                 type: cache.t2.micro
                 clusters: 2
+            elb:
+                stickiness:
+                    type: cookie
+                    cookie-name: journal
+                protocol:
+                    - https
+                    - http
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -501,6 +501,13 @@ journal:
                 engine: redis
                 type: cache.t2.micro
                 clusters: 2
+            elb:
+                stickiness:
```
giorgiosironi: we probably don't need this anymore, but identical to the existing configurations

thewilkybarkid: We don't need this anymore, and I think it's broken anyway as the neither CDN whitelists the `AWSELB` cookie.

giorgiosironi: Will remove as a follow up.

PR #346, opened 2018-06-12
ELPP-3393 Lower Fastly healthchecks interval

giorgiosironi:
Only applying to `api-gateway--end2end` for now, but it seems given the time for the DNS changes to propagate, the CDN should be more responsive to start routing traffic.

https://elifesciences.slack.com/archives/C5VTPCH08/p1528818653000110

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -484,7 +484,7 @@ journal:
                     - "{instance}--fastly-journal"
                 healthcheck:
                     path: /ping-fastly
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Applying everywhere :wrench:

giorgiosironi:
Postponing the healthchecks on `generic-cdn` as the S3 origin is not trivial to deal with, would have to add files to the buckets in special places.

PR #347, opened 2018-06-14
ELPP-3693 Add resources for EC2 dynamic creation in Jenkins

giorgiosironi:
EC2 instances dynamically created by Jenkins shouldn't be snowflakes, so these projects capture them with configuration as code.

## Features

- specify `ports` inside `aws.ec2.security-group`, continuing to push EC2-related configuration inside the `ec2` section but functionally equivalent to the existing `aws.ports`
- specify a custom size for the `/` root volume, as `aws.ec2.root.size` (possibly also `aws.ec2.root.type` in the future, same structure as `aws.ext`)

## `elife-alfred-nodes`

This project is an empty EC2 clusters, providing a security group which the instances created by Jenkins will use. Implicitly it also provides other inherited configuration to Jenkins: the VPC and the subnet the instances should be attached to.

## `containers--...`

This is the first kind of Jenkins node that can be dynamically created. Pretty much any container-related build can run on this node. The original stack is only needed to create an AMI, Jenkins than instantiates the AMI.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -47,6 +47,8 @@ defaults:
             masterless: false
             # optional: pin a master server for all new instances of a project
             master_ip: 10.0.2.198
+            # optional: specify `ports` to be opened
+            security-group: {}
         region: us-east-1
         vpc-id: vpc-78a2071d  # vpc-id + subnet-id are peculiar to AWS account + region
         subnet-id: subnet-1d4eb46a # elife-public-subnet, us-east-1d
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1095,6 +1095,9 @@ containers:
     aws-alt:
         jenkins-plugin-ami:
             ext: false
+            ec2:
+                root:
+                    size: 20 # GB
     vagrant:
         box: ubuntu/trusty64 # Ubuntu 14.04, deprecated
         ram: 1024
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1089,6 +1089,7 @@ elife-libraries:
         ram: 1024
 
 containers:
+    description: run any Docker-based workload
     subdomain: containers
     formula-repo: https://github.com/elifesciences/containers-formula
     aws:
@@ -1099,6 +1100,9 @@ containers:
             size: 100 # GB
             type: gp2
     aws-alt:
+        # AMIs created from here used in
+        # https://alfred.elifesciences.org/configure
+        # `Cloud` section
         jenkins-plugin-ami:
             ext: false
             ec2:
@@ -1138,12 +1142,12 @@ elife-alfred:
 # used in https://alfred.elifesciences.org/configure
 # `Cloud` section
 elife-alfred-nodes:
+    description: supports dynamic creation of EC2 instances from Jenkins
     domain: false
     intdomain: false
     aws:
         ec2:
-            cluster-size: 0 # will be created automatically
-                            # by the Jenkins EC2 plugin
+            cluster-size: 0
             security-group:
                 ports:
                     - 22
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1104,11 +1104,6 @@ containers:
             ec2:
                 root:
                     size: 20 # GB
-        jenkins-plugin-test:
-            ext: false
-            ec2:
-                root:
-                    size: 20 # GB
     vagrant:
         box: ubuntu/trusty64 # Ubuntu 14.04, deprecated
         ram: 1024
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1126,6 +1126,18 @@ elife-alfred:
         ports:
             1433: 80
 
+elife-alfred-nodes:
+    domain: false
+    intdomain: false
+    aws:
+        ec2:
+            cluster-size: 0 # will be created automatically
+            security-group:
+                # preferred to aws.ports for this use case?
+                ports:
+                    - 22
+            
+
 api-dummy:
     subdomain: api-dummy
     repo: https://github.com/elifesciences/api-dummy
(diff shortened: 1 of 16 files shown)


giorgiosironi:
I've left various TODOs I'd like to address before merge but review can start.

giorgiosironi:
> these objects you're replacing functions with.

Essentially the problem spawned from merging `aws.ports` with `aws.ec2.security-group.ports` (leaving the former in for backward compatibility, but using the latter to precisely mean). At the moment, `ports` seem not to merge different definitions like the other dictionaries in builder configuration (from `defaults`, `aws` and `aws-alt`).

When looking at the data structure passed down into `trop`, several options appeared reflecting the various complexity levels with which ports can be specified:

- list `[22, 80]`: easy
- remapping ports with a dictionary `{80: 8080}`
- specifying ip ranges `OrderedDict([(80, OrderedDict([('cidr-ip', '10.0.0.0/16'), ...]))])`
- mix and match `[22, OrderedDict([(80, OrderedDict([('cidr-ip', '10.0.0.0/16'), ...]))]), ...]`

`security_group` appeared to abstract over the different ports with a custom generic function `lmap`, which is a functional approach I think. `complex_ingress` however used conditionals on the type of data structure (`list` vs `dict` and checking the dict has only 1 element); this works locally for the generation of an `troposphere.ec2.SecurityGroupRule`, but not at the higher level like during merge of multiple configurations.

`Ingress.build` then tries to rationalize the simpler `ports` data structures into the lowest common denominator, the more complex one:

```
OrderedDict([
    (80, OrderedDict([('cidr-ip', '10.0.0.0/16'), ...])),
    ...
])
```

This should be equivalent to an `ingress_rationalize: mixed -> dict[int, dict]` function followed by a `ingress_to_troposphere: dict[int, dict] -> ec2.SecurityGroupRule` function (some other names).

Then I got concerned about misuse of the data structure, as there are many `dict`s and `list`s around and it's easy to pass a `dict` that is not a `Ingress` to a function that expect one, or a `Ingress` to a function that doesn't allow one (e.g. several errors I debugged like `TypeError: 'int' object has no attribute '__getitem__'`.

In a functional language, I would define a proper type for this data structure (rusty, broken Haskell follows to demonstrate):

```
data PortProtocol = Tcp | Udp
data PortConfiguration = PortConfiguration {guest :: Integer, protocol: PortProtocol, cidr :: String }
data Ingress = Map Int PortConfiguration
ingress_rationalize definition :: Any -> Ingress
ingress_to_troposphere :: Ingress -> ec2.SecurityGroupRule
```

In a dynamic language like Python, since I don't know a way to provide tipe safety with generic data structures like `list`s and `dict`s, I tend instead to introduce wrapper classes. There is no compilation step (although I've seen `pylint` doing a good basic job of finding gross inconsistencies), but at runtime there are very clear type errors when `443` is passed where a `{`cidr`: ...}` is required. I've found these to be very helpful when debugging integration tests, as there is a lot of indirection between the high-level operation (launch a project) and the inner workings (put this string into that template).

A last note is that `troposphere` itself tends to follow two approaches, a validation one and a small-classes one, both to stop users from passing strange values for a port or an ip. The validation approach stops numbers out or range e.g. a `1000000000` port; the second one works well for dictionaries as missing keys like `FromPort` or `ToPort` are immediately caught. `troposphere` doesn't do much rationalization apart from providing default values however, that would be out of scope for a generic library.

seanwiseman:
After reading the extensive explanation justifying the use and preference for a class over stand-alone functions in this instance, it seems that basing the approval of changes on personal beliefs and preferences and a functional vs OOP debate is only going to be/has already been a waste of time. 

It's fine to point out concerns with something that may conflict with how you would have implemented it, but if something is easy to reason about for both the implementor and the reader, well tested and is not causing any detrimental effects to the project in question then something like this should not hold it up.

lsh-0:
hi guys, apologies for not addressing the explanation further earlier, Wednesday became quite busy. 

First off, both of you, I'm not pushing a functional programming agenda. The opposite of using objects is not functional. I don't pine for the glorious fields of Haskell, where the grass is always greener. My concerns about the use of objects are entire practical and stem from the long term maintainability of a codebase. Simple functions and simple datastructures are easy to test, easy to reason about and the purer the better. The patterns of applying a function to a list of data items or transforming a piece of data by passing it through several functions are dead simple to understand, universal to all languages at this level and the flow is explicit. The pattern of reducing a problem down to it's simplest case 

Secondly, @giorgiosironi quite rightly points out that the convenient higher level yaml description isn't mapping particularly well into simple lower level data structures. Tweaks and hacks as we've gone along have confused the matter. The answer is not to define an object to encapsulate the data and push your logic into private methods. The answer is to strictly define the types/shapes of data we are working with and to, at the last possible minute, create the Troposphere objects needed. 

Using objects feels nice and Python provides wonderful support for OO approaches and there is pressure to conform when most other programmers freely use objects, but data hiding and stateful objects are code smells and we should be suspicious when we see them. They make reasoning about our programs difficult and undermine our _confidence_. They're also a bastard to serialise.

Introducing objects here encourages the introduction of objects elsewhere until it's a PHP-like mash.

> it seems that basing the approval of changes on personal beliefs and preferences and a functional vs OOP debate is only going to be/has already been a waste of time.

tsk. I disagree, obviously. We're programmers, we should be concerned about the health of our programs. Debate is rarely a waste of time. For example, I now have a better idea of where your own priorities lie.

The majority of @giorgiosironi's commits are fine, this one seems to be an exception. I'll take another look today and if I can improve it, I will.

lsh-0:
>  For example, I now have a better idea of where your own priorities lie.

that was catty, I apologise.

giorgiosironi:
Priority is to make progress on giving people fast builds :running_man: so ELPP-1295 is now the place for investigating schemas and type checks that, for example could help with type safety with the same intent as classes wrapping lists/dicts do.

PR #348, opened 2018-06-18
Docker vagrant tweaks

lsh-0:


Dockerfile.py2
```diff
@@ -2,6 +2,7 @@ FROM python:2.7.15-slim-stretch
 
 RUN pip install virtualenv
 RUN mkdir /venv && virtualenv --python=python2 /venv
-COPY py2-requirements.txt /
-RUN /venv/bin/pip install -r /py2-requirements.txt
+COPY py2-requirements.txt /requirements.txt
+COPY settings.yml /settings.yml
```
giorgiosironi: the `settings.yml` file should be visible at runtime, as the whole folder is mounted as a volume. I wouldn't bake it into the image unless it was influencing the result e.g. what is in the virtualenv?

lsh-0: settings.yml is a generated file, it's absence is breaking your build.

python-docker.sh
```diff
@@ -0,0 +1,31 @@
...
+
+. .settings.sh
+
+if [ -e .use-python-3.flag ]; then
+    image=py3
+else
+    image=py2
+fi
+
+# skips building image each time
```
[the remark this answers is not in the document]
lsh-0: I had a lot of junk in my builder installation, it was copying hundreds of megs even after I removed a ~2gb file ;) I wasn't sure what your intent/workflow was. You say it should be idempotent but the Docker file contains COPY and RUN commands and the contents of the build directory will change which ensure it gets re-built each time. I get the practicalities of having a container as a sort of an almost-there development environment but definitely prefer immutable images.

Dockerfile.py2
giorgiosironi: :+1: for providing it, but as a mounted volume like the rest of the files not related to the venv

python-docker.sh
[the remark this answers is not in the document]
giorgiosironi: There are two operations that you are seeing at work I think:

- the build context for Docker being zipped up and sent to the daemon. `.dockerignore` should exclude files from being picked up during this operation, it's missing something that you have there? I remember this being a problem until `.dockerignore` was expanded to include most of my folder's contents.
- the Dockerfile commands: `COPY` should be re-executed if the files involved change, e.g. `requirements.txt`. `RUN` should be re-executed if the command itself changes, or if a previous layer changes of course.

> definitely prefer immutable images.

Me too. We're not there yet but a "virtualenv" image should definitely be immutable enough that in theory people could use builder by pulling it down rather than building it (not that we do that). `settings.yml` is a bit special because it is a configuration (hence should not go in an image) but it's really static as it never changes.

I'll try to get both these and the original PR green.

giorgiosironi:
Recap of the build failure is:
```
java.nio.file.AccessDeniedException: /var/lib/jenkins-libraries-runner/workspace/er_PR-348-PYZN25INDBZ6OIJKO5D2/logs/app.log
```
This log file is written as the container user, hence the `elife` user Jenkins is using here is not capable of deleting it to cleanup the workspace. I think we can create it before the build with the right ownership.

giorgiosironi:
All problems originated by the container running as `root`, and leaving files around. I'll get this into the branch it forks from.

PR #349, opened 2018-06-18
Remove journal session stickiness

giorgiosironi:
Applied to `end2end` for now, will monitor end2end tests today.


giorgiosironi:
https://alfred.elifesciences.org/job/test-journal/1471/ passed, applying to other environments.

giorgiosironi:
Applied everywhere now.

PR #350, opened 2018-06-19
Deploy master to Alfred

giorgiosironi:
As soon as the tests are green, the new version can be deployed without waiting for the next daily update.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,4 +45,10 @@ elifePipeline {
             actions["Test py35"]() 
         }
     }
+
+    elifeMainlineOnly {
+        stage 'Deploy to Alfred', {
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Only way to try it out is to trigger `test-builder`, the main pipeline, by merging; I'll be back with results.

giorgiosironi:
A couple of builds failed for unrelated reasons (Jenkins upgrade), still waiting for a good one that will touch this part.

giorgiosironi:
Finally got there, executed a few commands, but broke on
```
IOError: [Errno 26] Text file busy: '/ext/srv/builder/venv/bin/python2'
```
because the process was running in some other build. I fear a lock on `builder` is needed when builder operations are executed (e.g. `bldr update`), unless there is a Unix/Python level solution to substitute a file while it is running.

PR #352, opened 2018-06-20
ELPP-3693 Use tasks.create_ami from Jenkins

giorgiosironi:
Minimal extensions to make the task usable from an automated build that updates the AMI for the `containers` stack:

- fixes a bug with uncommitted files that are sometimes lost in the AMI
- put AMI ID output on a single line
- put most of the messages on stderr
- allow non-interactive confirmation through environment variable
- wait until AMI is `available`, it may be `pending` for a while
- allow a custom name to avoid clashes on `containers.2018-06-20` when running a build more than once in a day

### DIFF
--- a/scripts/clean-stack-for-ami.sh
+++ b/scripts/clean-stack-for-ami.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# runs as ROOT on ALL AWS MINIONS directly after stack creation and before AMI 
+# runs as ROOT before AMI creation
 # creation
 
 set -e # everything must pass
@@ -7,12 +7,11 @@ set -u # no unbound variables
 set -xv  # output the scripts and interpolated steps
 
 if command -v salt-minion > /dev/null; then
-    # salt is installed, probably using an AMI or creating an AMI
-    # https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.saltutil.html#salt.modules.saltutil.clear_cache
+    # salt is installed
     systemctl stop salt-minion 2> /dev/null || service salt-minion stop
 fi
 
-# remove leftover files from AMIs
+# remove credentials and stack-specific files
 rm -rf \
     /etc/cfn-info.json \
     /etc/salt/pki/minion/* \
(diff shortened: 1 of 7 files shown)


scripts/clean-stack-for-ami.sh
```diff
@@ -23,3 +22,7 @@ rm -rf \
     /etc/certificates/* \
     /root/events.log \
     /var/cache/salt/minion
+
+# commit memory buffers to disk to avoid partly written files
+# since AMIs are essentially snapshots of the root volume
+sync
```
lsh-0: nice one

src/buildercore/bakery.py
```diff
@@ -10,18 +10,21 @@ def ami_name(stackname):
...
-def create_ami(stackname):
+def create_ami(stackname, name=None):
     "creates an AMI from the running stack"
     with core.stack_conn(stackname, username=config.BOOTSTRAP_USER):
-        bootstrap.prep_ec2_instance()
+        bootstrap.clean_stack_for_ami()
     ec2 = core.find_ec2_instances(stackname)[0]
     kwargs = {
-        'Name': ami_name(stackname),
+        'Name': ami_name(stackname) if name is None else name,
```
lsh-0: `name or ami_name(stackname)` might be better

src/utils.py
```diff
@@ -94,8 +95,12 @@ def uin(param, default=0xDEADBEEF):
 
 
 def confirm(message):
-    print(message)
-    print('press Enter to confirm (ctrl-c to quit)')
+    if config.BUILDER_NON_INTERACTIVE:
+        errcho('Non-interactive mode, confirming automatically')
```
lsh-0: lacks context without the `message` being printed. I would suggest using the logger in non-interactive mode and have these as debug statements

giorgiosironi: a `logger` should be a standard way to use `stderr`

lsh-0: yeeeessss, I agree on principle. but to get the same functionality with a logger one has to create a special handler with a special formatter. it's not the first thought that comes to mind when you 'just need to print to stderr'.

giorgiosironi: Able to add a logger, but will use `info` as a level so that the output is included in the Jenkins logs

PR #359, opened 2018-07-02
Force failure on masterless instance denied access to builder-private

giorgiosironi:
See https://github.com/elifesciences/builder-base-formula/pull/150

This only shows the failure, but the linked build will then correctly fail because of the access denied. Once we know more about the reason for the failure the solution may change.


lsh-0:
ok, so a: `git revert b7d12b879bc38ebd440edcaaf401d2537e7eeacd` and then a `./bldr update:{masterless-inst}` fixes my masterless instance.

giorgiosironi:
#360 but doesn't address why the key we are uploading isn't working. I found the key from `master-server` gets there even without the additional code, but it has the wrong permissions:
```
root@2018-04-09-2--master-server:/home/elife# ls -l /root/.ssh/id_rsa
-rw------- 1 ubuntu ubuntu 1670 Apr  9 02:58 /root/.ssh/id_rsa
root@ip-10-0-2-185:/home/ubuntu# ls -l /root/.ssh/id_rsa             
-rw-rw-r-- 1 ubuntu ubuntu 1670 Jul  4 11:42 /root/.ssh/id_rsa
```
Going to fix that to see if it's the only problem.

giorgiosironi:
And the public key `id_rsa.pub` is a generated one not getting updated, hence it breaks the SSH connection silently.

giorgiosironi:
Works for me with `bldr masterless.launch:basebox,test-builder-private-clone4,standalone`

giorgiosironi:
`basebox` is very useful to debug this kind of processes, a lean stack that comes up quickly from launch to complete highstate.

giorgiosironi:
Suggestions for an automated test that is not an integration one?

giorgiosironi:
(feel free to merge if a build is waiting on this)

PR #360, opened 2018-07-03
Fix master key upload reversion

lsh-0:
Could probably do with an explanation of why it's needed ...

@giorgiosironi , if this fixes your problem feel free to merge.

### DIFF
--- a/src/buildercore/bootstrap.py
+++ b/src/buildercore/bootstrap.py
@@ -10,7 +10,7 @@ from collections import OrderedDict
 from datetime import datetime
 from . import utils, config, bvars, core, context_handler, project, cloudformation, terraform, sns as snsmod
 from .context_handler import only_if as updates
-from .core import stack_all_ec2_nodes, project_data_for_stackname, stack_conn
+from .core import stack_all_ec2_nodes, project_data_for_stackname, stack_conn, stack_pem
 from .utils import first, ensure, subdict, yaml_dumps, lmap, fab_get, fab_put, fab_put_data
 from .lifecycle import delete_dns
 from .config import BOOTSTRAP_USER
--- a/src/buildercore/bootstrap.py
+++ b/src/buildercore/bootstrap.py
@@ -478,6 +478,13 @@ def update_ec2_stack(stackname, context, concurrency=None, formula_revisions=Non
         master_builder_key = download_master_builder_key(stackname)
 
     def _update_ec2_node():
+        # upload private key if not present remotely
+        if not files.exists("/root/.ssh/id_rsa", use_sudo=True):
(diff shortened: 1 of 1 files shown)


lsh-0:
I'm beginning to think elife-reporting has never been included on masterless instances.

lsh-0:
success! finally! I can recreate the s3bash problem remotely.

giorgiosironi:
It seems to fix the heavybox masterless build, but makes no sense to me yet; this should be a generated key used only for this stack so how could it give access to `builder-private`?

giorgiosironi:
Just as a cross-check, there are no deploy keys defined for that repository so it must be associated to a user:
https://github.com/elifesciences/builder-private/settings/keys
In fact:
```
root@ip-10-0-2-204:/opt/builder-private# ssh git@github.com
Hi elife-master-builder! You've successfully authenticated, but GitHub does not provide shell access.
```
This is the key that ends up there:
```
master-server--2018-04-09-2
Fingerprint: 84:c0:88:3a:eb:f4:84:f7:19:ae:a6:fb:74:d1:99:e7 Added on 9 Apr 2018 
```
associated to the `elife-master-builder` Github user. So it must be downloaded from the master-server, which is correct, but not sure how it is passed to this code.

src/buildercore/bootstrap.py
```diff
@@ -478,6 +478,13 @@ def update_ec2_stack(stackname, context, concurrency=None, formula_revisions=Non
         master_builder_key = download_master_builder_key(stackname)
 
     def _update_ec2_node():
+        # upload private key if not present remotely
+        if not files.exists("/root/.ssh/id_rsa", use_sudo=True):
+            # if it also doesn't exist on the filesystem, die horribly.
+            # regular updates shouldn't have to deal with this.
+            pem = stack_pem(stackname, die_if_doesnt_exist=True)
```
giorgiosironi: Two things perplex me:
1. we download the right key at https://github.com/elifesciences/builder/pull/360/files#diff-bfa1d9cba2806d6d2d41031884d3e00eR478 so it should really use that rather than `stack_pem`?
2. It should only do this if `is_masterless` is `True`, otherwise we randomly upload some key to all stacks

giorgiosironi: And we are already doing both 1 and 2 in https://github.com/elifesciences/builder/pull/360/files#diff-bfa1d9cba2806d6d2d41031884d3e00eR504

lsh-0:
Feel free to do more investigation yourself. It is blocking formula PRs though. I'll have another one that will probably fail in a minute.

giorgiosironi:
Closing in favor of #359 which has found more problems on the upload of the master builder key.

PR #361, opened 2018-07-03
Make X-Forwarded-Host compatible with Fastly shielding

thewilkybarkid:
Untested, but I think enabling shielding would set `X-Forwarded-Host` to be the shield's host header (since the VCL is executed on the edge and again on the shield).

### DIFF
--- a/src/buildercore/fastly/vcl/original-host.vcl
+++ b/src/buildercore/fastly/vcl/original-host.vcl
@@ -1,1 +1,3 @@
-set req.http.X-Forwarded-Host = req.http.host;
+if (!req.http.Fastly-FF) {
+  set req.http.X-Forwarded-Host = req.http.host;
+}


src/buildercore/fastly/vcl/original-host.vcl
```diff
@@ -1 +1,3 @@
-set req.http.X-Forwarded-Host = req.http.host;
+if (!req.http.Fastly-FF) {
```
giorgiosironi: needs a comment explaining the reason for the check (which should end up in the generated VCL)

thewilkybarkid:
Will expand on this for other incompatibilities.

thewilkybarkid:
Can't see anything else.

thewilkybarkid:
Turned on shielding on `journal--continuumtest`, seems ok but will have a bit more of a play.

giorgiosironi:
Was there an uncommitted change to get it to work on `continuumtest`? Related to the `Host` header

thewilkybarkid:
Some other changes were applied last week.

src/buildercore/terraform.py
```diff
@@ -92,6 +92,11 @@ def render_fastly(context):
     if not context['fastly']:
         return {}
 
+    def _guess_shield(region):
+        return {
+            'us-east-1': 'iad-va-us',
```
thewilkybarkid: Obviously an incomplete list. Could be made a setting, but not sure why you wouldn't want it closest to the origin.

src/buildercore/terraform.py
```diff
@@ -123,15 +128,17 @@ def render_fastly(context):
             backends.append(_fastly_backend(
                 backend['hostname'],
                 name=name,
-                request_condition=backend_condition_name
+                request_condition=backend_condition_name,
+                shield=backend.get('shield')
```
thewilkybarkid: Doesn't respect `context['fastly']['shielding']`. Should it?

src/tests/fixtures/projects/dummy-project.yaml
```diff
@@ -378,7 +380,18 @@ project-with-fastly-gcs:
...
+
+project-with-fastly-shielding:
+    repo: ssh://git@github.com/elifesciences/dummy3
+    subdomain: www
+    aws:
+        fastly:
+            subdomains:
+                - "{instance}--cdn-of-www"
+            subdomains-without-dns:
+                - "{instance}--www"
```
thewilkybarkid: Currently we'll have to set this (as the `Host` header is changed). Could make it automatic?

src/buildercore/terraform.py
giorgiosironi: Can be expanded if needed, clear enough

src/buildercore/terraform.py
giorgiosironi: Can see a use case where the `shield` configuration is fixed for all environments or projects, but `shielding` is turned on only for a few; so yes, should be stripped if `shielding` is `False`

giorgiosironi: (the two cases "explicit-backends" and "default `full_hostname`" will also be consistent with each other then)

src/tests/fixtures/projects/dummy-project.yaml
giorgiosironi: Yes but as a follow-up

src/tests/fixtures/projects/dummy-project.yaml
```diff
@@ -336,7 +337,8 @@ project-with-fastly-complex:
                     hostname: default.example.org
                 articles:
                     hostname: "{instance}-special.example.org"
-                    condition: req.url ~ "^/articles" 
+                    condition: req.url ~ "^/articles"
+                    shield: iad-va-us
```
giorgiosironi: the meaning of `shielding` vs `shield` doesn't transpire. A way this is dealt with consistently for other options is to make the positive value of `shielding` a dictionary containing an optional key `pop`:

```
shielding:
    pop: iad-va-us
# or, if you want the default
shielding: True
# (as an empty dictionary would still be Falsy)
```

It works much better when there are already other values, but it leave space for them to be inserted if needed. Any other setting comes to mind, that for example would customize a VCL snippet or influence something else?

It does take away support for defaults shielding configuration on projects with `shielding` turned off, but those can be left as:
```
#shielding:
#    pop: iad-va-us
```

thewilkybarkid: Could rename `shielding` `shield` and trigger a failure if both `backends` and `shield` are set (as they're incompatible)?

Don't know if we would ever need any other values, but don't mind making it a dictionary.

src/tests/fixtures/projects/dummy-project.yaml
```diff
@@ -331,12 +332,23 @@ project-with-fastly-complex:
...
                 articles:
                     hostname: "{instance}-special.example.org"
-                    condition: req.url ~ "^/articles" 
+                    condition: req.url ~ "^/articles"
+                    shield:
+                        pop: amsterdam-nl
+                articles2:
+                    hostname: "{instance}-special2.example.org"
+                    condition: req.url ~ "^/articles2"
+                    shield: true
```
thewilkybarkid: A little unclear given the higher-level `shield` is also `true`.

src/tests/fixtures/projects/dummy-project.yaml
```diff
@@ -331,12 +332,23 @@ project-with-fastly-complex:
         ports:
             - 80
         fastly:
+            shield: true
```
thewilkybarkid: Might actually be better as `default-shield` now?

src/buildercore/cfngen.py
```diff
@@ -263,9 +263,23 @@ def build_context_fastly(context, parameterize):
...
-        b['hostname'] = parameterize(b['hostname'])
-        return b
+    def _build_shield(shield):
+        if shield is False:
+            return {}
+
+        if shield is True:
+            pop = {
+                'us-east-1': 'iad-va-us',
+            }.get(context['project']['aws']['region']) or 'us-east-1'
```
giorgiosironi: shouldn't it be `}.get(context['project']['aws']['region'] or 'us-east-1')`?

src/buildercore/cfngen.py
```diff
@@ -277,9 +291,10 @@ def _parameterize_gcslogging(gcslogging):
     if context['project']['aws'].get('fastly'):
         backends = context['project']['aws']['fastly'].get('backends', OrderedDict({}))
         context['fastly'] = {
-            'backends': OrderedDict([(n, _parameterize_hostname(b)) for n, b in backends.items()]),
+            'backends': OrderedDict([(n, _build_backend(b)) for n, b in backends.items()]),
             'subdomains': [_build_subdomain(x) for x in context['project']['aws']['fastly']['subdomains']],
             'subdomains-without-dns': [_build_subdomain(x) for x in context['project']['aws']['fastly']['subdomains-without-dns']],
+            'shield': _build_shield(context['project']['aws']['fastly'].get('shield', False)),
```
giorgiosironi: probably needs to be applied to all existing CDN even if `false`, since the Terraform resource definition changes

src/buildercore/terraform.py
```diff
@@ -123,18 +125,22 @@ def render_fastly(context):
             backends.append(_fastly_backend(
                 backend['hostname'],
                 name=name,
-                request_condition=backend_condition_name
+                request_condition=backend_condition_name,
+                shield=backend['shield'].get('pop')
             ))
```
giorgiosironi: this branch, which is for multiple backends, doesn't seem to add to `all_allowed_subdomains`?

src/tests/fixtures/projects/dummy-project.yaml
giorgiosironi: That may be unclear if `backends` isn't used, so if the intention is to have a value here and for `backends.*` to override it, keeping the same name is consistent with other configurations

src/buildercore/terraform.py
giorgiosironi: Read a comment about shielding being incompatible with multiple backends, it's probably not the case anymore if `shield` is added here?

thewilkybarkid: Originally was going to make the default `shield` and `backends` incompatible (with the former referring solely to the automatic backend), but now you can disable/overwrite in each backend as needed.

src/buildercore/cfngen.py
```diff
@@ -25,6 +27,27 @@
...
+    'ap-northeast-2': 'osaka-jp', # Seoul: Osaka
+    'ap-northeast-3': 'osaka-jp', # Osaka-Local: Osaka
+    'ap-south-1': 'singapore-sg', # Mumbai: Singapore (change to Mumbai when available)
+    'ap-southeast-1': 'singapore-sg', # Singapore: Singapore
+    'ap-southeast-2': 'sydney-au', # Sydney : Sydney
+    'ca-central-1': 'yul-montreal-ca', # Canada (Central): Montreal
+    'cn-north-1': 'hongkong-hk', # Beijing: Hong Kong
+    'cn-northwest-1': 'hongkong-hk', # Ningxia: Hong Kong
+    'eu-central-1': 'frankfurt-de', # Frankfurt: Frankfurt
+    'eu-west-1': 'london_city-uk', # Ireland: London City
```
thewilkybarkid: London (Slough, I think) closer?

src/buildercore/cfngen.py
```diff
@@ -25,6 +27,27 @@
...
+    'eu-west-2': 'london_city-uk', # London: London City
+    'eu-west-3': 'cdg-par-fr', # Paris: Paris
+    'sa-east-1': 'gru-br-sa', # São Paulo: São Paulo
```
thewilkybarkid: Going to open a Fastly issue since there doesn't seem to be a documented list anywhere (even of POP IDs).

src/buildercore/cfngen.py
```diff
@@ -25,6 +27,27 @@
...
 
+FASTLY_AWS_REGION_SHIELDS = {
+    'us-east-1': 'iad-va-us', # N. Virginia: Washington
+    'us-east-2': 'iad-va-us', # Ohio: Washington
+    'us-west-1': 'sjc-ca-us', # N. California: San Jose
+    'us-west-2': 'sea-wa-us', # Oregon: Seattle
+    'ap-northeast-1': 'tokyo-jp2', # Tokyo: Tokyo
+    'ap-northeast-2': 'osaka-jp', # Seoul: Osaka
+    'ap-northeast-3': 'osaka-jp', # Osaka-Local: Osaka
+    'ap-south-1': 'singapore-sg', # Mumbai: Singapore (change to Mumbai when available)
```
thewilkybarkid: Mumbai is available, but doesn't appear as a config option on Fastly...

src/buildercore/cfngen.py
thewilkybarkid: Opened 88013.

src/tests/fixtures/projects/dummy-project.yaml
```diff
@@ -378,7 +390,30 @@ project-with-fastly-gcs:
...
+    aws:
+        fastly:
+            shield:
+                pop: london-uk
+
+project-with-fastly-shield-aws-region:
+    repo: ssh://git@github.com/elifesciences/dummy3
+    subdomain: www
+    aws:
+        region: eu-central-1
```
thewilkybarkid: Hmm, not supported changing the region...

thewilkybarkid: Tried to find a few ways round it (including using `dummy-project-eu.yaml`), but can't get anything to work.

giorgiosironi: There is more underneath that doesn't support multiple regions (e.g. the `vpc-id` and the various `subnet-*` value are specific to that VPC, which is in `us-east-1`) so it would be a separate feature for builder to support multiple regions for different projects or project instances. The `region` value is there because it's required for API calls to target the right URLs, but it isn't meant to be configurable at the moment.

src/buildercore/cfngen.py
thewilkybarkid: Updated based on their list.

src/buildercore/cfngen.py
```diff
@@ -132,11 +155,11 @@ def build_context(pname, **more_context): # pylint: disable=too-many-locals
     def _parameterize(string):
         return string.format(instance=context['instance_id'])
 
-    for topic_template_name in context['project']['aws']['sns']:
+    for topic_template_name in context['project']['aws'].get('sns', {}):
```
giorgiosironi: the default for this seems to be `[]`

src/buildercore/cfngen.py
```diff
@@ -263,9 +286,21 @@ def build_context_fastly(context, parameterize):
...
+        if shield is True:
+            pop = FASTLY_AWS_REGION_SHIELDS.get(context['project']['aws']['region'], 'us-east-1')
+
+            return {'pop': pop}
+
+        return shield
+
+    def _build_backend(backend):
+        backend['hostname'] = parameterize(backend['hostname'])
+        backend['shield'] = _build_shield(backend.get('shield', context['project']['aws']['fastly'].get('shield', False)))
```
giorgiosironi: this seems to be information that is independent of the system provisioning it (Terraform, API calls, whatever) so good fit for the `context`

thewilkybarkid:
Will need applying everywhere. Starting now.

PR #362, opened 2018-07-09
Fix surrogate keys

thewilkybarkid:


projects/elife.yaml
```diff
@@ -826,7 +826,7 @@ generic-cdn:
             surrogate-keys:
                 article-versioned-assets:
                     url: "^/articles/([0-9]+)/elife-([a-z-0-9]+)-v([0-9]+)\\.(.+)$"
-                    value: "articles/\\1 article/\\1 article/\\1v\\3"
+                    value: "article/\\1 article/\\1v\\3"
```
thewilkybarkid: I think `articles/` was here for BC, but since it's not available everywhere let's just remove it now.

gnott: Not as experienced with Fastly as Giorgio is, still I think you got the two places I'm aware of - this, where the surrogate key is added, and then the bot fastly provider that issues the purge for the surrogate key.

There's some samples in the lines below this that can probably also be cleaned ``expected: "articles/10627 ...`` etc.  https://github.com/elifesciences/builder/blob/03c3afbee090b42fb70cde92e95b6f01cd1bce32/projects/elife.yaml#L833

thewilkybarkid: Updated the samples (would be useful if they were test cases!).

PR #364, opened 2018-07-16
Set X-eLife-Google-Scholar-Metadata header for Journal

thewilkybarkid:
To match https://github.com/elifesciences/journal/pull/1030.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -496,6 +496,8 @@ journal:
                     - "original-host"
                     - "ping-status"
                     - "strip-non-journal-cookies"
+                    - "journal-google-scholar"
+                    - "journal-google-scholar-vary"
                 gcslogging:
                     bucket: "{instance}-elife-fastly"
                     path: "journal--{instance}/"
@@ -624,6 +626,8 @@ journal:
                     - "original-host"
                     - "ping-status"
                     - "strip-non-journal-cookies"
+                    - "journal-google-scholar"
+                    - "journal-google-scholar-vary"
                 gcslogging:
                     bucket: "{instance}-elife-fastly"
                     path: "journal--{instance}/"
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -551,6 +551,8 @@ journal:
                     - "original-host"
                     - "ping-status"
                     - "strip-non-journal-cookies"
+                    - "journal-google-scholar"
+                    - "journal-google-scholar-vary"
                 gcslogging:
                     bucket: "{instance}-elife-fastly"
                     path: "journal--{instance}/"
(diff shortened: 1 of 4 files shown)


giorgiosironi:
Rolls back the outstanding https://github.com/elifesciences/builder/pull/361 but that's an experiment so it has lower priority.

projects/elife.yaml
```diff
@@ -551,6 +551,8 @@ journal:
                     - "original-host"
                     - "ping-status"
                     - "strip-non-journal-cookies"
+                    - "journal-google-scholar"
+                    - "journal-google-scholar-vary"
```
giorgiosironi: the naming correctly reflects the specificity of the snipper for the `journal` project

src/buildercore/fastly/vcl/journal-google-scholar.vcl
```diff
@@ -0,0 +1,5 @@
+if (req.http.User-Agent ~ "Googlebot/") {
+    set req.http.X-eLife-Google-Scholar-Metadata = "1";
```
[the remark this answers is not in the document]
giorgiosironi: Good solution to leave the logic in a testable place

src/buildercore/fastly/vcl/journal-google-scholar-vary.vcl
```diff
@@ -0,0 +1,6 @@
+if (!req.http.Fastly-FF && !req.http.Fastly-Debug) {
+    set resp.http.Vary = regsub(resp.http.Vary, "X-eLife-Google-Scholar-Metadata, ", "");
+    if (resp.http.Vary ~ "^\s*$") {
+        unset resp.http.Vary;
+    }
+}
```
giorgiosironi: not sure what does this do, shouldn't be `Vary` be purely managed by the origin?

thewilkybarkid: Stops it being sent to the client.

giorgiosironi: Because it's related to an header that the client didn't send, that makes sense. This is still end-to-end testable by setting `User-Agent` manually and going through the CDN.

thewilkybarkid:
Now deployed on `continuumtest`.

src/buildercore/fastly/vcl/journal-google-scholar.vcl
```diff
@@ -0,0 +1,5 @@
...
+} else {
+    unset req.http.X-eLife-Google-Scholar-Metadata;
```
giorgiosironi: what's the reason for explicitly unset rather than not taking an action? Preventing the header setting from the outside? (but it would be security through obscurity as the user agent can be set)

thewilkybarkid: Yep. Limits it to the only triggering mechanism. (That header being an internal detail.)

It we really cared we could limit it to Google's IP range, but I wasn't going to bother so its easily testable. (And isn't a problem if someone else triggers it, just reduces the caching rate.)

PR #366, opened 2018-07-19
Remove (most of) Journal's CloudFronts

thewilkybarkid:
Main problem is end2end is still using CloudFront rather than Fastly (https://github.com/elifesciences/elife-spectrum/pull/120).

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -452,37 +452,9 @@ journal:
                 protocol:
                     - https
                     - http
-            cloudfront:
-                subdomains:
-                    - "{instance}--cdn-journal"
-                headers:
-                    - X-Requested-With
-                    - Host
-                cookies:
-                    - journal
-                errors: 
-                    # TODO: this is not being deployed to, it's empty
-                    domain: end2end-elife-error-pages.s3-website-us-east-1.amazonaws.com
-                    pattern: "???.html"
-                    codes:
-                        # ELB with no active instances
-                        503: "/5xx.html"
-                    protocol: http
-                origins:
-                    CloudFrontCDNOrigin:
-                        hostname: end2end--journal.elifesciences.org
-                    LogInOrigin:
-                        hostname: end2end--journal.elifesciences.org
-                        pattern: /log-in
-                        headers:
-                            - X-Requested-With
-                            - Host
-                            - Referer
-                        cookies:
-                            - journal
             fastly:
                 subdomains:
-                    - "{instance}--fastly-journal"
+                    - "{instance}--cdn-journal"
                 healthcheck:
                     path: /ping-fastly
                     check-interval: 10000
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -597,20 +535,9 @@ journal:
...
-                            - Referer
-                        cookies:
-                            - journal
-                logging:
-                    bucket: elife-cloudfront-logs
             fastly:
                 subdomains:
-                    - "{instance}--fastly-journal"
+                    - "{instance}--cdn-journal"
                     - "placeholder-prod-journal" # only here to avoid rewriting the following values
```
thewilkybarkid: Unrelated, but not sure what this comment refers to.

thewilkybarkid:
Tried to apply to `continuumtest` and failed with `RuntimeError: stack status is 'UPDATE_ROLLBACK_COMPLETE'`.

thewilkybarkid:
Hmm, `ERROR - buildercore.core - stack_status is 'UPDATE_ROLLBACK_COMPLETE', cannot move from that`.

thewilkybarkid:
Same in `end2end`.

Seems to be it trying to update and delete the `{instance}--cdn-journal` entry at the same time. Removing it, then adding to Fastly as a second-step. Currently waiting for CloudFormation to finish on `continuumtest`.

thewilkybarkid:
Applied to `continuumtest`, doing `end2end`.

giorgiosironi:
> Seems to be it trying to update and delete the {instance}--cdn-journal entry at the same time. Removing it, then adding to Fastly as a second-step.

Yes, CloudFormation is not able to figure out that the DNS entry should be first deleted and added, but instead executes two different changesets that conflicts. There is downtime on the DNS entry absence but the one that is really important (`elifesciences.org`) should not change.

projects/elife.yaml
giorgiosironi: The DNS entries are numbered in the CloudFormation template:
```
$ cat .cfn/stacks/journal--prod.json | jq '.Resources | keys' | grep DNS
  "CloudFrontCDNDNS1",
  "CloudFrontCDNDNS2",
  "CloudFrontCDNDNS3",
  "CloudFrontCDNDNS4",
  "ExtDNS",
  "FastlyDNS1",
  "FastlyDNS2",
  "FastlyDNS3",
```
So if `placeholder-prod-journal` were to be removed, the following DNS entry would have to be deleted and re-added, probably in two-steps, causing unnecessary trouble.

thewilkybarkid: Do we create that value or CloudFormation?

thewilkybarkid:
Applied everywhere apart from `prod`.

projects/elife.yaml
```diff
@@ -576,17 +522,9 @@ journal:
                     - http
             cloudfront:
```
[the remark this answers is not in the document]
giorgiosironi: Makes sense, maybe they should be move to `redirects--prod` but at another time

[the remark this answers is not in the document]
thewilkybarkid: Investigated options in https://github.com/elifesciences/jira-import/issues/4314.

thewilkybarkid:
Applying to `prod` before merging.

projects/elife.yaml
thewilkybarkid: Would using a hash rather than an auto-increment have avoided the problem? (I get we can't change it now easily.)

thewilkybarkid:
Doesn't seem to have worked correctly, https://elife.elifesciences.org/ is now accessible rather than redirecting.

projects/elife.yaml
```diff
@@ -576,41 +522,20 @@ journal:
...
-                    - "{instance}--cdn-journal"
+                    - "placeholder2-prod-journal" # only here to avoid rewriting the following values
                     - www
                     - elife
                     - prod
-                subdomains-without-dns:
-                    - ""
-                headers:
-                    - X-Requested-With
-                    - Host
```
thewilkybarkid: Think we needed to keep this.

projects/elife.yaml
thewilkybarkid: Is reordering a safe operation? If so, could it be reordered so the placeholder is last, then remove it as a follow up?

thewilkybarkid:
All deployed now, merging when green.

thewilkybarkid:
Disk space problem?

projects/elife.yaml
giorgiosironi: Re-ordering not safe as every resource is independent, so re-additions would likely clash with removals done at the same time - no clear solution at the moment, generating something better than an incremental number as a function of the value would make it more stable but complicated to port to that

PR #368, opened 2018-07-23
Digests project, stateless

giorgiosironi:
A microservice ingesting and exposing digests data through a REST API. Backed by Postgres and emitting events about updates on the bus.

/cc @lsh-0


giorgiosironi:
Updated the various `bus` instances.
Updated `master-server` so that it picks up the new formula.
Created `digests--end2end`, `digests--continuumtest`, `digests--prod`.

PR #369, opened 2018-07-24
Serve stale content on an error

thewilkybarkid:
Currently viewable on https://continuumtest--cdn-journal.elifesciences.org/ (I've tweaked the `Cache-Control` there to make it cacheable, though it'll be reset on the next highstate so you might need to change it to `public, max-age=30, stale-if-error=86400`).

This also strips the `Cookie` header when restarting so logged-in users might get a stale logged-out version, rather than an error page. This is the only controversial part. /cc @BlueReZZ

Testing as public:

1. View a page on the site a few times to populate the cache (make sure you're logged out).
2. Stop Nginx.
3. Wait 30 seconds.
4. Verify the equivalent page on https://continuumtest--journal.elifesciences.org/ doesn't work.
5. Verify the page through the CDN is continues to be displayed and the `Age` is beyond the `max-age`.
6. Verify that other pages through the CDN return a 503 (when there's no cache entry).

Testing as a logged-in user:

1. View a page on the site a few times to populate the cache (make sure you're logged out).
2. Log in to the site (uses ORCID sandbox).
2. Stop Nginx.
3. Wait 30 seconds.
4. Verify the equivalent page on https://continuumtest--journal.elifesciences.org/ doesn't work.
5. Verify the page through the CDN is continues is displayed, the `Age` is beyond the `max-age` and that you are logged out.
6. Verify that other pages through the CDN return a 503 (when there's no cache entry).
7. Start Nginx.
8. Verify the equivalent page on https://continuumtest--journal.elifesciences.org/ now works.
9. Verify you appear back as logged in on https://continuumtest--cdn-journal.elifesciences.org/.

### DIFF
--- a/src/buildercore/fastly/vcl/error-page.vcl.tpl
+++ b/src/buildercore/fastly/vcl/error-page.vcl.tpl
@@ -1,5 +1,10 @@
 if (obj.status == ${code}) {
-   set obj.http.Content-Type = "text/html; charset=us-ascii";
-      synthetic {"${synthetic_response}"};
-     return(deliver);
+  if (stale.exists) {
+    return(deliver_stale);
+  }
+
+  set obj.http.Content-Type = "text/html; charset=us-ascii";
+  synthetic {"${synthetic_response}"};
+
+  return(deliver);
 }
--- a/src/buildercore/fastly/vcl/error-page.vcl.tpl
+++ b/src/buildercore/fastly/vcl/error-page.vcl.tpl
@@ -1,10 +1,5 @@
 if (obj.status == ${code}) {
-  if (stale.exists) {
-    return(deliver_stale);
-  }
-
-  set obj.http.Content-Type = "text/html; charset=us-ascii";
-  synthetic {"${synthetic_response}"};
-
-  return(deliver);
+   set obj.http.Content-Type = "text/html; charset=us-ascii";
+      synthetic {"${synthetic_response}"};
+     return(deliver);
 }
--- a/src/buildercore/fastly/vcl/main.vcl
+++ b/src/buildercore/fastly/vcl/main.vcl
@@ -11,8 +11,18 @@ sub vcl_recv {
 sub vcl_fetch {
   #FASTLY fetch
 
-  if ((beresp.status == 500 || beresp.status == 503) && req.restarts < 1 && (req.request == "GET" || req.request == "HEAD")) {
-    restart;
+  if (beresp.status >= 500 && beresp.status < 600) {
+    if (stale.exists) {
+      return(deliver_stale);
+    }
+
+    if (req.restarts < 1 && (req.request == "GET" || req.request == "HEAD")) {
+      unset req.http.Cookie;
+
+      restart;
+    }
+
+    error 503;
   }
 
   if (req.restarts > 0) {
@@ -27,13 +37,6 @@ sub vcl_fetch {
   if (beresp.http.Cache-Control ~ "private") {
     set req.http.Fastly-Cachetype = "PRIVATE";
     return(pass);
-  }
-
-  if (beresp.status == 500 || beresp.status == 503) {
-    set req.http.Fastly-Cachetype = "ERROR";
-    set beresp.ttl = 1s;
-    set beresp.grace = 5s;
-    return(deliver);
   }
 
   if (beresp.http.Expires || beresp.http.Surrogate-Control ~ "max-age" || beresp.http.Cache-Control ~ "(s-maxage|max-age)") {
@@ -61,6 +64,12 @@ sub vcl_miss {
 }
 
 sub vcl_deliver {
+  if (resp.status >= 500 && resp.status < 600 && stale.exists) {
+    unset req.http.Cookie;
+
+    restart;
+  }
+
   #FASTLY deliver
   return(deliver);
 }
--- a/src/buildercore/fastly/vcl/main.vcl
+++ b/src/buildercore/fastly/vcl/main.vcl
@@ -1,5 +1,6 @@
 sub vcl_recv {
   if (req.restarts < 1) {
+    # Sanitise header
     unset req.http.X-eLife-Restart;
   }
 
@@ -22,7 +23,7 @@ sub vcl_fetch {
 
     if (req.restarts < 1 && (req.request == "GET" || req.request == "HEAD")) {
       set req.http.X-eLife-Restart = "fetch," beresp.status;
-      unset req.http.Cookie;
+      unset req.http.Cookie; # Temporarily log out the user
 
       restart;
     }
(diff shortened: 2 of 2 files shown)


giorgiosironi:
I can successfully test the "age greater than max-age" behavior. There is a lag of some tens of seconds where the origin is detected as working again, which is expected as some healthcheck is re-executed periodically.

src/buildercore/fastly/vcl/main.vcl
```diff
@@ -11,8 +11,18 @@ sub vcl_recv {
...
+      return(deliver_stale);
+    }
+
+    if (req.restarts < 1 && (req.request == "GET" || req.request == "HEAD")) {
+      unset req.http.Cookie;
+
+      restart;
+    }
+
+    error 503;
```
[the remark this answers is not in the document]
giorgiosironi: From the guide I understand all the VCL customizations are used before the origin is marked as unhealthy.

BlueReZZ:
I think the logged in use case is good - much better than not having access to the content which is the primary use case for the journal. I'll mention it to others at the planning meeting as an FYI but I see this as an enhancement.

src/buildercore/fastly/vcl/main.vcl
```diff
@@ -11,8 +11,18 @@ sub vcl_recv {
...
 
-  if ((beresp.status == 500 || beresp.status == 503) && req.restarts < 1 && (req.request == "GET" || req.request == "HEAD")) {
-    restart;
+  if (beresp.status >= 500 && beresp.status < 600) {
+    if (stale.exists) {
+      return(deliver_stale);
+    }
+
+    if (req.restarts < 1 && (req.request == "GET" || req.request == "HEAD")) {
+      unset req.http.Cookie;
```
thewilkybarkid: Assumes that it's the only thing in `Vary`.

src/buildercore/fastly/vcl/main.vcl
```diff
@@ -11,8 +11,18 @@ sub vcl_recv {
...
   #FASTLY fetch
 
-  if ((beresp.status == 500 || beresp.status == 503) && req.restarts < 1 && (req.request == "GET" || req.request == "HEAD")) {
-    restart;
+  if (beresp.status >= 500 && beresp.status < 600) {
+    if (stale.exists) {
+      return(deliver_stale);
+    }
+
+    if (req.restarts < 1 && (req.request == "GET" || req.request == "HEAD")) {
```
giorgiosironi: given the lack of subroutines, the various cases need comments to describe actions like `retry without being logged in`

giorgiosironi: (in a programming language it would be `retry_without_being_logged_in()` which obviates comments)

src/buildercore/fastly/vcl/error-page.vcl.tpl
```diff
@@ -1,5 +1,10 @@
 if (obj.status == ${code}) {
-   set obj.http.Content-Type = "text/html; charset=us-ascii";
-      synthetic {"${synthetic_response}"};
-     return(deliver);
```
giorgiosironi: random indentation :man_shrugging:

src/buildercore/fastly/vcl/main.vcl
giorgiosironi: not sure if this loses the original HTTP status code, turning all `50x` into `500`. Would fit with the proxy model, but hides the real error

src/buildercore/fastly/vcl/main.vcl
thewilkybarkid: Varnish allows for custom subroutines. Wonder if Fastly does too. 🤔

src/buildercore/fastly/vcl/main.vcl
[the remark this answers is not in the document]
giorgiosironi: Understood this, I agree with the removal

src/buildercore/fastly/vcl/main.vcl
```diff
@@ -29,13 +39,6 @@ sub vcl_fetch {
     return(pass);
   }
 
-  if (beresp.status == 500 || beresp.status == 503) {
-    set req.http.Fastly-Cachetype = "ERROR";
```
giorgiosironi: no idea what this header was, just placed there for logging?

src/buildercore/fastly/vcl/error-page.vcl.tpl
```diff
@@ -1,5 +1,10 @@
...
+  if (stale.exists) {
+    return(deliver_stale);
```
thewilkybarkid: Should probably move this to `main.vcl` and serve it on any `5xx`.

src/buildercore/fastly/vcl/main.vcl
```diff
@@ -61,6 +64,12 @@ sub vcl_miss {
 }
 
 sub vcl_deliver {
+  if (resp.status >= 500 && resp.status < 600 && stale.exists) {
+    unset req.http.Cookie;
```
giorgiosironi: `Cookie` is unset in two places, so what are the two different use cases?

src/buildercore/fastly/vcl/main.vcl
thewilkybarkid: Not sure. Is part of the boilerplate.

src/buildercore/fastly/vcl/main.vcl
thewilkybarkid: I think they include it in `deliver` as well as you don't necessarily go through `fetch`.

src/buildercore/fastly/vcl/main.vcl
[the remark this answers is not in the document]
thewilkybarkid: Just tested removing it, ELB sends an empty 503 response when the backend are unhealthy. So we'll need to keep using the synthetic if there's no body.

src/buildercore/fastly/vcl/main.vcl
thewilkybarkid: Thinking about it, probably don't need it here (assume `stale.exists` would always be `false` if you have a cookie).

Not sure how to test this behaviour at the minute.

giorgiosironi: Can we add something like `req.http.Some-Standard-Elife-Debug-Header` and get it logged so that we see if this section is ever evaluated with live traffic?

src/buildercore/fastly/vcl/main.vcl
```diff
@@ -11,8 +16,21 @@ sub vcl_recv {
...
+    }
+
+    if (req.restarts < 1 && (req.request == "GET" || req.request == "HEAD")) {
+      set req.http.X-eLife-Restart = "fetch," beresp.status;
+      unset req.http.Cookie; # Temporarily log out the user
+
+      restart;
+    }
+
+    if (!beresp.http.Content-Length || beresp.http.Content-Length == "0") {
```
giorgiosironi: would also warrant a comment linking it to the ELB behavior

thewilkybarkid:
Will deploy everywhere then merge in the morning.

PR #370, opened 2018-07-25
Pass formula_revisions where needed

giorgiosironi:
Fixes what was seen in https://github.com/elifesciences/lax-formula/pull/30 and moves some logic around to try to avoid omitted arguments. Makes formula builds useless until fixed.


giorgiosironi:
First attempts to make this more robust show `cfn.launch` accepts arguments that are used only in the case of `masterless` and ignored otherwise. Going deeper to fix this problem, which should allow the arguments to be more explicit.

src/buildercore/bootstrap.py
```diff
@@ -447,7 +449,7 @@ def upload_master_configuration(master_stack, master_configuration_data):
 
 @updates('ec2')
 @core.requires_active_stack
-def update_ec2_stack(stackname, context, concurrency=None, formula_revisions=None, **kwargs):
+def update_ec2_stack(stackname, context, concurrency=None, formula_revisions=None):
```
lsh-0: I'm becoming a bit sceptical of funcs accepting a `**splat` these days. Explicit is better than implicit and all that.

src/buildercore/checks.py
```diff
@@ -2,14 +2,30 @@
...
 these should compliment not replicate any project configuration validation."""
 
-from . import project
+from . import core, project
 from .project import repo
 
+class AccessProblem(RuntimeError):
+    pass
+
+class StackAlreadyExistingProblem(RuntimeError):
```
lsh-0: `StackAlreadyExistsProblem` or `ExtantStackProblem` or perhaps just `StackExistsError` might be better

src/cfn.py
```diff
@@ -142,41 +142,25 @@ def helpfn(altkey):
 
 @task
 @requires_project
-def launch(pname, instance_id=None, alt_config=None, **kwargs):
-    try:
-        stackname = generate_stack_from_input(pname, instance_id, alt_config)
-        pdata = core.project_data_for_stackname(stackname)
-
-        print('attempting to create stack:')
```
lsh-0: these (I think) were literally just temporary debug statements that I forgot to remove years and years ago. I feel like they could be expanded upon, but not here, obviously.

```
attempting to create stack 'project--instance' in 'us-east-1'. stack has:
- ec2 (t2.large)
- rds (t2.db.small, "sadfasdfsad.rds.aws.amazon.com")
- dns (instance--project.elifesciences.org)
```
etc

PR #371, opened 2018-07-25
Expand IIIF surrogate keys

giorgiosironi:
For https://github.com/elifesciences/jira-import/issues/4349#issuecomment-407778723

Cater for:

- `/` service-to-file separator rather than Loris's `:`, for future use. But is it a
slash or an encoded slash?
- `%2F` in addition to the unencoded slash as a folder separator.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1593,8 +1593,8 @@ iiif:
                 timeout: 2000
             surrogate-keys:
                 article-id-versioned:
-                    url: "^/lax:([0-9]+)/elife-([a-z0-9-]+)-v([0-9]+)\\.(.+)$"
-                    value: "article/\\1 article/\\1v\\3"
+                    url: "^/lax[:/]([0-9]+)(/|%2F)elife-([a-z0-9-]+)-v([0-9]+)\\.(.+)$"
+                    value: "article/\\1 article/\\1v\\4"
                     samples:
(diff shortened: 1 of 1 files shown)


thewilkybarkid:
> But is it a slash or an encoded slash?

Plain slash. (This is the cause of the Loris problem: slashes are meaningful.)

projects/elife.yaml
```diff
@@ -1593,8 +1593,8 @@ iiif:
                 timeout: 2000
             surrogate-keys:
                 article-id-versioned:
-                    url: "^/lax:([0-9]+)/elife-([a-z0-9-]+)-v([0-9]+)\\.(.+)$"
-                    value: "article/\\1 article/\\1v\\3"
+                    url: "^/lax[:/]([0-9]+)(/|%2F)elife-([a-z0-9-]+)-v([0-9]+)\\.(.+)$"
+                    value: "article/\\1 article/\\1v\\4"
```
thewilkybarkid: Could not capture the new group instead.

giorgiosironi:
> Assume the samples are used in the tests?

They are run as cross-checks during `update_infrastructure`

giorgiosironi:
Applying now everywhere `iiif--*`.

giorgiosironi:
Purging all 3 of them now.

PR #373, opened 2018-08-01
crm, adds 10GB external volume

lsh-0:
@giorgiosironi , you've done this a bit now, anything I need to watch out for? Will there be downtime?


giorgiosironi:
No, there shouldn't be downtime as it is a new volume that can be attached to the existing instance. It should only be creating new CloudFormation resources, not modifying the existing `EC2Instance1`.

You will need to add the `elife.external-volume` Salt state which will format it and mount it for you.

lsh-0:
the other CRM issues have been resolved now, merging this in

PR #374, opened 2018-08-03
Capture buckets for data-pipeline

giorgiosironi:
/cc @de-code

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -2018,3 +2018,13 @@ fastly-logs:
             "{instance}-elife-fastly":
                 project: elife-fastly
     aws-alt: {}
+
+data-pipeline:
+    description: Data consolidation project
+    domain: false
+    intdomain: false
+    aws:
+        ec2: false
+        s3:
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Created `data-pipeline--demo`:

```
$ aws s3 ls | grep demo
2018-08-03 11:58:27 demo-elife-data-pipeline
2018-08-03 11:58:26 demo-elife-data-pipeline-archive
```

projects/elife.yaml
```diff
@@ -2018,3 +2018,13 @@ fastly-logs:
...
     aws-alt: {}
+
+data-pipeline:
+    description: Data consolidation project
+    domain: false
+    intdomain: false
+    aws:
+        ec2: false
+        s3:
+            "{instance}-elife-data-pipeline":
```
de-code: do we need a `deletion-policy`?

giorgiosironi: `delete` (throw away with the project instance) or `retain` (preserve after deletion)?

giorgiosironi: added

PR #376, opened 2018-08-03
RDS for digests

giorgiosironi:
For https://github.com/elifesciences/jira-import/issues/4338

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -72,6 +72,8 @@ defaults:
             version: '9.4'
             type: db.t2.small
             storage: 5 # GB
+            # Standard for backward compatibility
+            storage-type: Standard # standard|gp2|io1 https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html
             backup-retention: 28 # days
             # if rds.params are specified, a custom db parameter group is created
             params: []
@@ -1823,7 +1825,9 @@ digests:
         end2end:
             rds:
                 engine: postgres
+                type: db.t2.micro
                 storage: 5 # GB
+                storage-type: gp2
                 backup-retention: 2 # days
         continuumtest: {}
             #rds:
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1829,16 +1829,20 @@ digests:
                 storage: 5 # GB
                 storage-type: gp2
                 backup-retention: 2 # days
-        continuumtest: {}
-            #rds:
-            #    engine: postgres
-            #    storage: 5 # GB
-            #    backup-retention: 2 # days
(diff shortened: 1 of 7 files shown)


projects/elife.yaml
```diff
@@ -72,6 +72,8 @@ defaults:
             version: '9.4'
             type: db.t2.small
             storage: 5 # GB
+            # Standard for backward compatibility
+            storage-type: Standard # standard|gp2|io1 https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html
```
lsh-0: :+1:

projects/elife.yaml
```diff
@@ -1820,21 +1822,27 @@ digests:
...
-            #rds:
-            #    engine: postgres
-            #    storage: 5 # GB
-            #    backup-retention: 2 # days
-        prod: {}
-            #rds:
-            #    engine: postgres
-            #    storage: 5 # GB
-            #    multi-az: true
+        end2end:
```
lsh-0: eh, this repetition is annoying but I'm guessing you're having to do it for a good reason

src/buildercore/cfngen.py
```diff
@@ -217,7 +217,10 @@ def build_context_rds(context, existing_context):
 
     # pull password from existing context, if it exists
     generated_password = utils.random_alphanumeric(length=32)
-    rds_password = existing_context.get('rds_password', generated_password)
+    rds_password = existing_context.get('rds_password')
+    if not rds_password:
+        # may be present but None
```
lsh-0: `existing_context.get('rds_password') or generated_password`

lsh-0: I wonder if there is a familiarity vs readability argument that could be made here ...

projects/elife.yaml
giorgiosironi: Mainly the base stack doesn't have RDS (would be used for ad-hoc stacks or for testing formulas) as it takes half the time to provision; Then it means various environments have to repeat the RDS configuration, in this case `engine`, `type`, `storage` and `storage-type` (`backup-retention` differs). A way to remove this could be defaults that are applied to certain environments e.g. a top level `default-for-environments:` key with `end2end`, `prod`, etc. subsection?

src/buildercore/cfngen.py
giorgiosironi: Was a quick hack to make it work, I think I'll go through this module in a follow-up

projects/elife.yaml
lsh-0: I think the project definition logic is already too clever. I'm working on a replacement in my own time. For now, a little repetition is better than any more cleverness

PR #377, opened 2018-08-06
[WIP] Enable Fastly shielding

thewilkybarkid:

PR #378, opened 2018-08-06
Let Fastly set the host header

thewilkybarkid:
#377 is currently blocked as Fastly don't allow multiple domains. This unsets the `Host` header when making a request to the backend, Fastly will then set the value.

### DIFF
--- a/src/tests/test_buildercore_terraform.py
+++ b/src/tests/test_buildercore_terraform.py
@@ -87,13 +87,6 @@ class TestBuildercoreTerraform(base.BaseCase):
                                 'ssl_check_cert': True,
                             }],
                             'default_ttl': 3600,
-                            'request_setting': [{
-                                'name': 'default',
-                                'default_host': 'prod--www.example.org',
-                                'force_ssl': True,
-                                'timer_support': True,
-                                'xff': 'leave',
-                            }],
                             'gzip': {
                                 'name': 'default',
                                 'content_types': ['application/javascript', 'application/json',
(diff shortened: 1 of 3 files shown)


thewilkybarkid:
Currently enabled with shielding on `generic-cdn--continuumtest`, and without shielding on `api-gateway--continuumtest`.

src/tests/test_buildercore_terraform.py
```diff
@@ -87,13 +87,6 @@ def test_fastly_template_minimal(self):
                                 'ssl_check_cert': True,
                             }],
                             'default_ttl': 3600,
-                            'request_setting': [{
-                                'name': 'default',
-                                'default_host': 'prod--www.example.org',
-                                'force_ssl': True,
```
giorgiosironi: the other settings like `force_ssl` are guaranteed anyway?

thewilkybarkid: Backend is port 443, so should always be HTTPS. Plus: https://github.com/elifesciences/builder/blob/9641e7d6109f6ad460377aac5c1717f4c7dec0a5/src/buildercore/terraform.py#L323-L327

thewilkybarkid: Oh hang on, forces the client's request to be HTTPS. Looks to no longer work.

thewilkybarkid: Restored (and improved) in 7e865f34b369fe1cad01c9e8ab46224512fffa8e.

giorgiosironi:
Understood, the default request setting being applied to all backends :+1:

thewilkybarkid:
Will apply everywhere and merge.

PR #379, opened 2018-08-07
Skip parts of update_infrastructure

giorgiosironi:
update_infrastructure runs updates also on EC2, SQS, S3 resources after CloudFormation and Terraform. This is in general necessary e.g. extending the filesystem of a volume after its size has been changed, or subscribing a queue after it has been created. If you really know what you're doing, you can skip some steps for faster feedback

### DIFF
--- a/src/cfn.py
+++ b/src/cfn.py
@@ -79,10 +79,7 @@ def update_infrastructure(stackname, skip=None):
 
     Allows to skip EC2, SQS, S3 updates by passing `skip=ec2\\,sqs\\,s3`"""
 
-    if skip:
-        skip = skip.split(",")
-    else:
-        skip = []
+    skip = skip.split(",") if skip else []
 
     (pname, _) = core.parse_stackname(stackname)
     more_context = {}
--- a/src/cfn.py
+++ b/src/cfn.py
@@ -63,7 +63,7 @@ def update(stackname, autostart="0", concurrency='serial'):
 
 @task
 @timeit
-def update_infrastructure(stackname):
+def update_infrastructure(stackname, skip=None):
     """Limited update of the Cloudformation template and/or Terraform template.
 
     Resources can be added, but most of the existing ones are immutable.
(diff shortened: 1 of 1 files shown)


src/cfn.py
```diff
@@ -75,7 +75,14 @@ def update_infrastructure(stackname):
     but without any software being on it)
 
     Moreover, EC2 instances must be running while this is executed or their
-    resources like PublicIP will be inaccessible"""
+    resources like PublicIP will be inaccessible.
+
+    Allows to skip EC2, SQS, S3 updates by passing `skip=ec2\\,sqs\\,s3`"""
```
giorgiosironi: this weird notation is needed because `,` separates different arguments to `update_infrastructure` (which has 2: `stackname` and `skip`). Can replace with another separator if we prefer one

src/cfn.py
```diff
@@ -63,7 +63,7 @@ def update(stackname, autostart="0", concurrency='serial'):
 
 @task
 @timeit
-def update_infrastructure(stackname):
+def update_infrastructure(stackname, skip=None):
```
thewilkybarkid: Could make it a whitelist?

giorgiosironi: If it had a default of `ec2,sqs,s3` (to preserve safety of running `update_infrastructure` without arguments), but will make it slightly more complex to run `bldr update_infrastructure:journal--prod,components=[]`

src/cfn.py
lsh-0: can we do `./bldr update-infrastructure:project--instance,skip="ec2,sqs,s3"` ?

src/cfn.py
```diff
@@ -75,7 +75,14 @@ def update_infrastructure(stackname):
     but without any software being on it)
 
     Moreover, EC2 instances must be running while this is executed or their
-    resources like PublicIP will be inaccessible"""
+    resources like PublicIP will be inaccessible.
+
+    Allows to skip EC2, SQS, S3 updates by passing `skip=ec2\\,sqs\\,s3`"""
+
+    if skip:
```
lsh-0: `skip = skip.split(",") if skip else []`

src/cfn.py
lsh-0: approved, but lets fix up that escaped notation if possible

lsh-0: this works

```python
@task
def foo(args):
    print(args.split(" "))
```
and 
```bash
$ ./bldr foo:args="1 2 3"
['1', '2', '3']

done
```

PR #380, opened 2018-08-07
[wip] Fastly X-Forwarded-Host header debugging

giorgiosironi:
Do not merge. Applying to continuumtest to debug shielding issues.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -504,10 +504,11 @@ journal:
                     - "strip-non-journal-cookies"
                     - "journal-google-scholar"
                     - "journal-google-scholar-vary"
+                    - "log-outgoing-hosts"
                 gcslogging:
                     bucket: "{instance}-elife-fastly"
                     path: "journal--{instance}/"
-                    period: 600
+                    period: 60
         prod:
             description: prod environment for journal. ELB on top of multiple EC2 instances
             ec2:
--- a/src/buildercore/fastly.py
+++ b/src/buildercore/fastly.py
@@ -92,6 +92,12 @@ VCL_SNIPPETS = {
         type='recv',
         hook='before'
     ),
+    'log-outgoing-hosts': FastlyVCLSnippet(
+        name='original-host',
+        content=_read_vcl_file('log-outgoing-hosts.vcl'),
+        type='miss',
+        hook='before'
+    ),
     'gzip-by-content-type-suffix': FastlyVCLSnippet(
         name='gzip-by-content-type-suffix',
         content=_read_vcl_file('gzip-by-content-type-suffix.vcl'),
--- a/src/buildercore/fastly/vcl/log-outgoing-hosts.vcl
+++ b/src/buildercore/fastly/vcl/log-outgoing-hosts.vcl
@@ -1,0 +1,1 @@
+set req.http.X-Fastly-Outgoing-Original-Host = bereq.http.X-Forwarded-Host;
--- a/src/buildercore/fastly/vcl/original-host.vcl
+++ b/src/buildercore/fastly/vcl/original-host.vcl
@@ -1,3 +1,4 @@
+set req.http.X-Fastly-Set-Original-Host = req.http.X-Forwarded-Host;
 # Set X-Forwarded-Host if not a shield request
 if (!req.http.Fastly-FF) {
   set req.http.X-Forwarded-Host = req.http.host;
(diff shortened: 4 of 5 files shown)


giorgiosironi:
We can log the Fastly-set `X-Forwarded-Host`, which appears to be wrongly set on the shield node (but not at the edge node):
```
{
  "timestamp": "2018-08-07T15:22:43",
  "time_elapsed": 484905,
  "object_hits": 0,
  "object_lastuse": "1533655363.664",
  "is_tls": true,
  "client_ip": "212.44.25.140",
  "forwarded_for": "212.44.25.140, 157.52.92.37",
  "geo_city": "san francisco",
  "geo_country_code": "US",
  "pop_datacenter": "DCA",
  "pop_region": "US-East",
  "shield": "(null)",
  "request": "GET",
  "original_host": "continuumtest--cdn-journal.elifesciences.org, continuumtest--cdn-journal.elifesciences.org",
  "fastly_set_original_host": "continuumtest--cdn-journal.elifesciences.org, continuumtest--cdn-journal.elifesciences.org",
  "fastly_outgoing_original_host": "(null)",
  "host": "continuumtest--cdn-journal.elifesciences.org",
  "url": "/",
  "request_referer": "",
  "request_user_agent": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:61.0) Gecko/20100101 Firefox/61.0",
  "request_accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
  "request_accept_language": "en-GB,en;q=0.5",
  "request_accept_charset": "",
  "response_status": "200",
  "cache_status": "PASS"
}
```
The outgoing original host is probably being logged incorrectly, because it's always `null`.

giorgiosironi:
In terms of standards, [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) doesn't talk about any comma but the [Apache implementation does](https://httpd.apache.org/docs/2.2/mod/mod_proxy.html#x-headers).

giorgiosironi:
Could we simplify https://github.com/elifesciences/journal-formula/blob/master/salt/journal/config/etc-nginx-sites-available-journal.conf#L45-L50 to only use the `Host` now that Fastly passes that unchanged?

giorgiosironi:
Our nginx configuration appears to be magically able to deal with multiple values in `X-Forwarded-Host` so it may be a red herring:
```
curl -v localhost/ -H "Host: elifesciences.org" -H "X-Forwarded-Host: elifesciences.org, elifesciences.org" | grep og:url
<meta property="og:url" content="http://elifesciences.org/">
```
The first value is matched and used:
```
curl -v localhost/ -H "Host: elifesciences.org" -H "X-Forwarded-Host: a.elifesciences.org, b.elifesciences.org" | grep og:url
<meta property="og:url" content="http://a.elifesciences.org/">
```

The `Vary` never contains `Host` but [the RFC says it's included by default](https://tools.ietf.org/html/rfc7231#section-7.1.4).

giorgiosironi:
Let's try https://github.com/elifesciences/builder-private/commit/683ba9d444904486e4ed4f5ea5bb188348012d68

giorgiosironi:
Btw, we need to set `Cookie: journal=...` to get to see a full cache miss, any other cookie won't do.

giorgiosironi:
Just to exclude suspects, ELB does not set `X-Forwarded-Host`. Would add a log on Journal of what it is receiving.

giorgiosironi:
Getting too magic, the base host is preferred for some reason:
```
[17:50:09][giorgio@Newton:~/code/builder/builder-private]$ curl -H "Host: elifesciences.org" -H "X-Forwarded-Host: a.elifesciences.org, b.elifesciences.org" https://prod--journal.elifesciences.org | grep og:url
<meta property="og:url" content="https://a.elifesciences.org/">
[17:51:24][giorgio@Newton:~/code/builder/builder-private]$ curl -H "Host: elifesciences.org" -H "X-Forwarded-Host: a.elifesciences.org, elifesciences.org" https://prod--journal.elifesciences.org | grep og:url
<meta property="og:url" content="https://elifesciences.org/">
```

giorgiosironi:
Long story short: so far cannot reproduce on `continuumtest`, but also cannot reproduce by simulating shield-like requests on `prod`.

giorgiosironi:
The problem was identified in the nginx configuration not supporting the double `X-Forwarded-Host` value, so I'm going to close this, and eventually reapply to `continuumtest` the standard configuration but not while debugging is being done there.

giorgiosironi:
(it was actually reproducible by using `Host: prod--journal.elifesciences.org` which is what Fastly is passing)

PR #381, opened 2018-08-07
Turn off shielding in journal--prod

thewilkybarkid:
Currently has problems due to #380.


giorgiosironi:
Has been applied, so merging to reflect the current state.

PR #386, opened 2018-08-22
Feature: nifi demo

lsh-0:
this is a project entry for a small demo of Apache NiFi I've put together.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -211,6 +211,9 @@ defaults:
         ram: 1024
         cpus: 2
         cpucap: 100 # percent (vagrant default)
+        #ports:
+        #    1239: 80
+        #    1240: 8001
 
 basebox:
     formula-repo: https://github.com/elifesciences/basebox-formula
@@ -280,6 +283,14 @@ master-server:
             subdomains:
                 - master-server
     vagrant: {}
+
+nifi-demo:
+    formula-repo: https://github.com/elifesciences/nifi-demo-formula
+    vagrant:
+        cpus: 4
+        ram: 4098
+        ports:
+            1222: 8080
 
 lax:
     description: article-json store
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -286,11 +286,12 @@ master-server:
 
 nifi-demo:
     formula-repo: https://github.com/elifesciences/nifi-demo-formula
+    subdomain: nifidemo
     vagrant:
         cpus: 4
         ram: 4098
         ports:
-            1222: 8080
+            1222: 80
 
 lax:
     description: article-json store
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -284,15 +284,6 @@ master-server:
                 - master-server
     vagrant: {}
 
-nifi-demo:
-    formula-repo: https://github.com/elifesciences/nifi-demo-formula
-    subdomain: nifidemo
-    vagrant:
-        cpus: 4
-        ram: 4098
-        ports:
-            1222: 80
-
 lax:
     description: article-json store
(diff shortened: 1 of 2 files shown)


projects/elife.yaml
```diff
@@ -2040,3 +2043,17 @@ data-pipeline:
                  deletion-policy: delete
             "{instance}-elife-data-pipeline-archive":
                  deletion-policy: retain
+
+nifi-demo:
+    formula-repo: https://github.com/elifesciences/nifi-demo-formula
+    subdomain: nifidemo
+    aws:
+        ec2:
+            ami: ami-1d4e7a66
```
giorgiosironi: non-default `ami` -> add a comment on what it is

lsh-0:
@de-code , @giorgiosironi , this is me merging the nifi-demo and data-pipeline projects together.

I'll remove the nifi-demo project once I've transplated the configuration in the currently running instance.

I'll create a new formula 'data-pipeline-formula' with the nifi-demo work in it's own statefile `nifi.sls`

projects/elife.yaml
```diff
@@ -2047,12 +2050,40 @@ fastly-logs:
...
+    intdomain: pipeline
+    formula-repo: https://github.com/elifesciences/data-pipeline-formula
     aws:
-        ec2: false
+        ec2:
+            ami: ami-1d4e7a66 # 16.04
+        ports:
+            - 22
+            - 443
         s3:
```
giorgiosironi: this environment exists:
https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stack/detail?stackId=arn:aws:cloudformation:us-east-1:512686554592:stack%2Fdata-pipeline--demo%2F238635c0-970c-11e8-9a6a-50d5cd26c2d2
You can either remove it and create new ones, or run `update_infrastructure` to add the new resources.

projects/elife.yaml
```diff
@@ -2047,12 +2050,40 @@ fastly-logs:
...
+    domain: pipeline
+    intdomain: pipeline
+    formula-repo: https://github.com/elifesciences/data-pipeline-formula
     aws:
-        ec2: false
+        ec2:
+            ami: ami-1d4e7a66 # 16.04
+        ports:
+            - 22
+            - 443
```
giorgiosironi: HTTPS-only... ok

lsh-0: just an oversight. I've stuck nifi behind an nginx instance like most everything we use. NiFi is very strict about security, unsurprising given it's origins, and it took a bit of wrangling to ensure the headers were being proxied correctly. I think my browser did the upgrade to https without ever attempting http so it never came across my radar.

projects/elife.yaml
lsh-0: or we could leave it be or destroy it. @de-code , what is using the `demo` instance?

de-code: It was only there for the initial prototype. One was the _inbox_, from which it was moved to the _archive_. It kind of makes the backfill straightforward (just copy it back into the _inbox_). Personally I'd prefer read-only access to the input.

projects/elife.yaml
giorgiosironi: It is good for me as this `ports` settings is only used on AWS and `anything.elifesciences.org` is pinned to HTTPS anyway. Things port 80 is useful for:

- within Vagrant
- `anything.elife.internal` hostname use
- redirecting to HTTPS

projects/elife.yaml
giorgiosironi: :+1: for destroying and recreating? The buckets will be left around by the current deletion strategy I think, need to be manually cleaned up.

PR #388, opened 2018-08-30
Clarify instruction for giving access to a new master server

giorgiosironi:
Deploy keys do not work here, as they cannot be used for more than one repository; both `builder-private` and `crm-formula` are private.

### DIFF
--- a/scripts/init-master.sh
+++ b/scripts/init-master.sh
@@ -52,10 +52,10 @@ if [ ! -d /opt/builder-private ]; then
         echo "
 ----------
 
-could not clone your 'builder-private' repository:
+could not clone the 'builder-private' repository:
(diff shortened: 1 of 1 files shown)


scripts/init-master.sh
```diff
@@ -67,7 +67,7 @@ after the setup of the key, complete this process by running builder's 'update'
 
     ./bldr update:$stackname
 
-Note this key must be added also tom any private formula you want the master-server to use.
+This key will also be used to access to any private formula you want the master-server to use. The elife-master-builder Github user should already have access to these formulas.
```
lsh-0: typo here, and do a line break

giorgiosironi:
Ready and green now.

PR #389, opened 2018-08-30
Add builder-configuration on master servers

giorgiosironi:
Part of https://github.com/elifesciences/issues/issues/3691

The repository is still mostly empty, but it needs to be deployed before it can be referenced further.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -357,9 +357,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         # configure the instance as if it were a master server
         if IS_MASTER
             pillar_repo = "https://github.com/elifesciences/builder-private-example"
+            configuration_repo = "https://github.com/elifesciences/builder-configuration"
             all_formulas = project_cmd("--formula")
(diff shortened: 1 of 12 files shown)


giorgiosironi:
After tests pass, going to try to apply to the real `master-server`.

giorgiosironi:
(probably will take more just to get the tests together)

lsh-0:
I would err on the side of caution. I won't be available until Mon

lsh-0:
I'll merge this in first thing tomorrow so I can handle any problems that might come up

lsh-0:
instance required a `./bldr update_infrastructure` to refresh the instance data ('context') otherwise it would throw a keyerror in `buildercore/bootstrap.py` and we don't seem to have a task for simply refreshing the context. The test instance was only good up to a point, but crashes with an invalid master private key when it tries to run highstate.
The prod master-server applies the changes without issue but I'll keep an eye out for any weirdness

giorgiosironi:
> instance required a ./bldr update_infrastructure to refresh the instance data ('context') 

Makes sense, wouldn't do that for each deployment.

PR #390, opened 2018-09-04
issue 4411, remove layering violation from cfngen.py/trop.py

lsh-0:
https://github.com/elifesciences/issues/issues/4411

* cfngen, all context wrangling shifted into build_context_* funcs
* isolate each wrangler, lots of assumptions going on in each one
* removed `context.project` access during context creation time
* updated/replaced `context.project` access outside of context creation time with a pared down version of the project data

### DIFF
--- a/src/buildercore/bootstrap.py
+++ b/src/buildercore/bootstrap.py
@@ -206,7 +206,7 @@ def sub_sqs(stackname, context_sqs, region):
 @updates('sqs')
 @core.requires_active_stack
 def update_sqs_stack(stackname, context, **kwargs):
-    region = context['project']['aws']['region']
+    region = context['aws']['region']
     unsub_sqs(stackname, context['sqs'], region)
     sub_sqs(stackname, context['sqs'], region)
 
@@ -220,10 +220,7 @@ def update_s3_stack(stackname, context, **kwargs):
     This function adds also a Statement to the Policy of the involved queue so that this bucket can send messages to it.
     """
     context_s3 = context['s3']
-
-    # pdata = project_data_for_stackname(stackname) # suspect.
-    pdata = context['project'] # more likely
-    region = pdata['aws']['region'] # region is now pulled from context['project']['aws']['region'] like sqs. but is this correct?
+    region = context['aws']['region']
 
     ensure(isinstance(context_s3, dict), "Not a dictionary of bucket names pointing to their configurations: %s" % context_s3)
 
@@ -460,23 +457,23 @@ def update_ec2_stack(stackname, context, concurrency=None, formula_revisions=Non
     script that can be downloaded from the web and then very conveniently
     installs it's own dependencies. Once Salt is installed we give it an ID
     (the given `stackname`), the address of the master server """
+
+    # no longer a snapshot of the project data at creation time, now it's
+    # mostly just top-level project attributes as we need them.
     pdata = context['project']
+
     # backward compatibility: old instances may not have 'ec2' key
     # consider it true if missing, as newer stacks e.g. bus--prod
     # would have it explicitly set to False
     default_ec2 = {'masterless': False, 'cluster-size': 1}
-    ec2 = pdata['aws'].get('ec2', default_ec2)
-
-    # TODO: check no longer necessary
-    if not ec2:
-        return
+    ec2 = context.get('ec2', default_ec2)
 
     # TODO: check if any active stacks still have ec2: True in their context
     # and refresh their context then remove this check
     if ec2 is True:
         ec2 = default_ec2
 
-    region = pdata['aws']['region']
+    region = context['aws']['region']
     is_master = core.is_master_server_stack(stackname)
     is_masterless = ec2.get('masterless', False)
 
--- a/src/buildercore/bootstrap.py
+++ b/src/buildercore/bootstrap.py
@@ -477,7 +477,7 @@ def update_ec2_stack(stackname, context, concurrency=None, formula_revisions=Non
     if is_masterless:
         master_builder_key = download_master_builder_key(stackname)
 
-    fkeys = ['formula-repo', 'formula-depedencies', 'private-repo', 'configuration-repo']
+    fkeys = ['formula-repo', 'formula-dependencies', 'private-repo', 'configuration-repo']
     fdata = subdict(context['project'], fkeys)
 
     def _update_ec2_node():
--- a/src/buildercore/bootstrap.py
+++ b/src/buildercore/bootstrap.py
@@ -356,7 +356,7 @@ def template_info(stackname):
     # replaces the standand list-of-dicts 'outputs' with a simpler dict
     data['outputs'] = reduce(utils.conj, map(lambda o: {o['OutputKey']: o['OutputValue']}, data['outputs']))
 
-    return utils.subdict(data, keepers.values())
+    return subdict(data, keepers.values())
 
 def write_environment_info(stackname, overwrite=False):
     """Looks for /etc/cfn-info.json and writes one if not found.
@@ -458,10 +458,6 @@ def update_ec2_stack(stackname, context, concurrency=None, formula_revisions=Non
     installs it's own dependencies. Once Salt is installed we give it an ID
     (the given `stackname`), the address of the master server """
 
-    # no longer a snapshot of the project data at creation time, now it's
-    # mostly just top-level project attributes as we need them.
-    pdata = context['project']
-
     # backward compatibility: old instances may not have 'ec2' key
     # consider it true if missing, as newer stacks e.g. bus--prod
     # would have it explicitly set to False
@@ -481,11 +477,14 @@ def update_ec2_stack(stackname, context, concurrency=None, formula_revisions=Non
     if is_masterless:
         master_builder_key = download_master_builder_key(stackname)
 
+    fkeys = ['formula-repo', 'formula-depedencies', 'private-repo', 'configuration-repo']
+    fdata = subdict(context['project'], fkeys)
+
     def _update_ec2_node():
         # write out environment config (/etc/cfn-info.json) so Salt can read CFN outputs
         write_environment_info(stackname, overwrite=True)
 
-        salt_version = pdata['salt']
+        salt_version = context['project']['salt']
         install_master_flag = str(is_master or is_masterless).lower() # ll: 'true'
 
         build_vars = bvars.read_from_current_host()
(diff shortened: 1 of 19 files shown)


lsh-0:
... it actually wasn't that hard, there were just *a lot* of accesses using this pattern. Hopefully the distinction between reading the project data and reading the context data is now clearer.

src/project.py
```diff
@@ -17,16 +17,30 @@ def lst():
...
         'json': core_utils.json_dumps,
         'yaml': core_utils.yaml_dumps,
         # None: core_utils.remove_ordereddict
         None: lambda v: v
     }
+    ensure(output_format in formatters.keys(), "unknown output format %r" % output_format)
     formatter = formatters.get(output_format)
     return formatter(project.project_data(pname))
 
+@task
```
giorgiosironi: very useful to introspect the shape of `context`

src/masterless.py
```diff
@@ -114,8 +114,11 @@ def launch(pname, instance_id=None, alt_config='standalone', *repolist):
 @requires_masterless
 def set_versions(stackname, *repolist):
     "call with formula name and a revision, like: builder-private@ab87af78asdf2321431f31"
-    ctx = context_handler.load_context(stackname)
-    repolist = parse_validate_repolist(ctx['project'], *repolist)
+    #ctx = context_handler.load_context(stackname)
+    #repolist = parse_validate_repolist(ctx['project'], *repolist)
+
+    pdata = core.project_data_from_stackname(stackname)
```
giorgiosironi: uses `elife.yaml` data rather than an existing context for an older instance, but for `masterless` usage this is fine

src/masterless.py
```diff
@@ -25,7 +25,7 @@ def requires_masterless(fn):
     @wraps(fn)
     def wrapper(stackname=None, *args, **kwargs):
         ctx = context_handler.load_context(stackname)
-        ensure(stackname and ctx['project']['aws']['ec2']['masterless'], "this command requires a masterless instance.")
+        ensure(stackname and ctx['aws']['ec2']['masterless'], "this command requires a masterless instance.")
```
giorgiosironi: shouldn't this be `ctx['ec2']['masterless']`?

src/master.py
```diff
@@ -116,13 +117,13 @@ def remaster(stackname, new_master_stackname):
...
 
-    LOG.info("setting new master address")
-    context = cfngen.set_master_address(context, master_ip) # mutator
-
     LOG.info("upgrading salt client")
     pdata = core.project_data_for_stackname(stackname)
     context['project']['salt'] = pdata['salt']
 
+    LOG.info("setting new master address")
+    cfngen.set_master_address(pdata, context, master_ip) # mutator
```
giorgiosironi: not clear in general what it mutates (`context` as opposed to `pdata`), could be the first argument

src/master.py
```diff
@@ -97,15 +97,16 @@ def remaster(stackname, new_master_stackname):
...
-        # TODO: duplicates bad ec2 data wrangling in cfngen.build_context
-        # ec2 == True for some reason, which is completely useless
-        LOG.warn("bad context for stack: %s", stackname)
-        context['ec2'] = {}
-        context['project']['aws']['ec2'] = {}
-
-    if context.get('ec2') == False:
-        LOG.info("ec2 == False, skipping %s", stackname)
+
+    # remove if no longer an issue
```
giorgiosironi: no idea on whether this will happen again, good call on commenting it out

giorgiosironi: (rather than removing)

src/project.py
giorgiosironi: should probably support an environment too, e.g. `lax,end2end` to check for stuff that is related only to particular environments like RDS

src/buildercore/trop.py
```diff
@@ -205,18 +200,14 @@ def ec2instance(context, node):
...
     project_ec2 = {
-        "ImageId": lu('project.aws.ec2.ami'),
-        "InstanceType": context['ec2']['type'], # t2.small, m1.medium, etc
+        "ImageId": lu('ec2.ami'),
+        "InstanceType": lu('ec2.type'), # "t2.small", "m1.medium", etc
         "KeyName": Ref(KEYPAIR),
         "SecurityGroupIds": [Ref(SECURITY_GROUP_TITLE)],
-        "SubnetId": subnet_id, # ll: "subnet-1d4eb46a"
+        "SubnetId": subnet_id, # "subnet-1d4eb46a"
         "Tags": instance_tags(context, node),
```
giorgiosironi: the `trop` module in general seems to have been simplified

src/buildercore/cfngen.py
```diff
@@ -109,108 +120,169 @@ def build_context(pname, **more_context): # pylint: disable=too-many-locals
...
-    context.update(core.hostname_struct(stackname))
+    # provides: 'domain', 'int_domain', 'subdomain',
+    #           'hostname', 'project_hostname', 'int_project_hostname',
+    #           'full_hostname', 'int_full_hostname'
+    context.update(core.hostname_struct(context['stackname']))
+
+    # project data
+    # preseve some of the project data. all of it is too much
+    keepers = [
+        'salt',
```
giorgiosironi: later on `salt` should probably be moved to `ec2` as it's only used for that. `formula-*` fields can in theory be used with K8S too so it's the right place.

src/buildercore/cfngen.py
```diff
@@ -109,108 +120,169 @@ def build_context(pname, **more_context): # pylint: disable=too-many-locals
...
+
+    return context
+
+def build_context_aws(pdata, context):
+    if 'aws' not in pdata:
+        return context
+    keepers = [
+        'region',
+        'vpc-id',
+        'subnet-id',
```
giorgiosironi: maybe the `subnet-*` and `redundant-subnet-*` fields could be moved to `context['ec2']` (not part of this PR)
RDS refers to them only as "allows EC2 instances to connect to me"; not sure if ElastiCache is the same.

giorgiosironi: `context['elasticache']` has its own `subnets` field so yes, we'll move them.
`region` and `vpc-id` are so general for AWS that they are already in the right place.

giorgiosironi: `context['elb']` will need to get them from `aws.elb.subnets`, as it doesn't at the moment

src/buildercore/cfngen.py
```diff
@@ -109,108 +120,169 @@ def build_context(pname, **more_context): # pylint: disable=too-many-locals
...
+        build_context_vault,
+    ]
+
+    # exceptions to the rule ...
+    wrangler_list = [project_wrangler] + wrangler_list
+
+    for wrangler in wrangler_list:
+        # so functions can't modify the context in-place or
+        # reference a bit of project_data and then change it
+        context = wrangler(deepcopy(project_data), deepcopy(context))
```
giorgiosironi: that's good, true immutability. Ideally they wouldn't need `context` as an input, right? But it would be difficult to merge them all.

giorgiosironi: `project_name` and `stackname` are the only fields that are read from the context in this module, so I'd try to simplify that to remove `deepcopy(context)` as an argument. Would be a follow-up when this has been integrated.

src/buildercore/bootstrap.py
```diff
@@ -460,23 +457,23 @@ def update_ec2_stack(stackname, context, concurrency=None, formula_revisions=Non
     script that can be downloaded from the web and then very conveniently
     installs it's own dependencies. Once Salt is installed we give it an ID
     (the given `stackname`), the address of the master server """
+
+    # no longer a snapshot of the project data at creation time, now it's
+    # mostly just top-level project attributes as we need them.
     pdata = context['project']
```
giorgiosironi: should be renamed from `pdata` to something like `project_repositories` to avoid confusion with the "snapshot of the project data"

src/master.py
lsh-0: I'll elaborate

src/buildercore/trop.py
lsh-0: a little nip and tuck

src/buildercore/cfngen.py
```diff
@@ -222,125 +294,130 @@ def build_context_rds(context, existing_context):
...
 
-def build_context_gcp(context, parameterize):
-    if 'gcs' in context['project']['aws']:
+def build_context_gcp(pdata, context):
+    context['gcs'] = False
+    if 'gcs' in pdata['aws']:
         context['gcs'] = OrderedDict()
-        for bucket_template_name, options in context['project']['aws']['gcs'].items():
-            bucket_name = parameterize(bucket_template_name)
+        for bucket_template_name, options in pdata['aws']['gcs'].items():
```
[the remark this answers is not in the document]
lsh-0: yes, we have another ticket (I think?) to move `gcs` out from under `aws`.

src/buildercore/cfngen.py
lsh-0: this snippet was copied from trop.py and trop.py pares down even further

src/buildercore/cfngen.py
lsh-0: :+1:

src/buildercore/cfngen.py
lsh-0: I can see pros and cons, but even though I commented it out, I think shuffling the order of the wranglers is a good idea to prevent ordering dependencies. Removing the context as a parameter would obviate the shuffling.

lsh-0:
@giorgiosironi , I think I addressed your questions. if you're happy please merge.

src/masterless.py
lsh-0: hm, missed this one. you might be right.

lsh-0: yes, it happens here: https://github.com/elifesciences/builder/blob/3886100f47ecf87c0c39019c1b001cb345bb92c6/src/buildercore/cfngen.py#L265

I have it all checked out, I'll fix it up

giorgiosironi:
I am finding some problems in the `masterless` module, will do some full testing on it.

lsh-0:
no worries - let me know if I can help

giorgiosironi:
We need to expand `integration_tests` to cover also some basic `masterless` tasks; I am afraid of the test suite slowdown however.

lsh-0:
well done, I wonder how that typo made it past testing

lsh-0:
can we run those integration tests in parallel?

giorgiosironi:
> can we run those integration tests in parallel?

Probably, but there is some [outstanding problem to solve I'll get on today](https://github.com/elifesciences/builder/pull/290)

PR #393, opened 2018-09-24
Add RDS to elife-xpub--prod

giorgiosironi:
For https://github.com/elifesciences/elife-xpub/issues/623

Sets up RDS only for `elife-xpub--prod`; splitted out:

- encryption at rest, will be covered later by re-creating the instance.
- [`elife-xpub--end2end`, will be created later](https://github.com/elifesciences/elife-xpub/issues/648)

To be precise, we standardize on RDS Postgres, not Aurora.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1972,6 +1972,15 @@ elife-xpub:
             s3:
                 "{instance}-elife-xpub":
                      deletion-policy: retain
+            rds:
+                multi-az: true
+                engine: postgres # or 'MySQL'
+                version: '10.5'
+                type: db.t2.small
+                storage: 5 # GB
+                storage-type: gp2
+                backup-retention: 28 # days
+                deletion-policy: Snapshot
     vagrant:
         ram: 2048
         ports:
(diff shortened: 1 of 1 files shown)


damnhipster:
Happy with this :)

Do we have anything setup to alert us for the database? e.g. disk space, connections, etc.

projects/elife.yaml
```diff
@@ -1972,6 +1972,15 @@ elife-xpub:
...
                      deletion-policy: retain
+            rds:
+                multi-az: true
+                engine: postgres # or 'MySQL'
+                version: '10.4'
+                type: db.t2.small
+                storage: 5 # GB
+                storage-type: gp2
+                backup-retention: 28 # days
+                deletion-policy: Snapshot
```
damnhipster: What does the deletion policy do?

giorgiosironi: In case the database is deleted, it takes a snapshot before doing so. Resources in `prod` are stopped from being deleted anyway by some IAM policies we have.

giorgiosironi:
> anything setup to alert us for the database

CloudWatch is a pain to configure, I tried New Relic in the past (giving it access to our account essentially) but wasn't very effective for general AWS services. Have no off-the-shelf solution at the moment.

PR #394, opened 2018-09-24
Support RDS encryption at rest

giorgiosironi:
Eventually for https://github.com/elifesciences/elife-xpub/issues/623

A key is managed by AWS, still to be seen which permissions that key should have for RDS to pick it up.

Key at https://console.aws.amazon.com/iam/home?region=us-east-1#/encryptionKeys/us-east-1/b626157a-1e5b-4bf6-8799-211505efe072 if you can see it.

### DIFF
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -224,7 +224,6 @@ def project_wrangler(pdata, context):
     context['project'] = subdict(pdata, keepers)
 
     # limited to just master/masterless servers
-    print(pdata)
     is_masterless = pdata['aws'].get('ec2') and pdata['aws']['ec2']['masterless']
     is_master = core.is_master_server_stack(context['stackname'])
     if is_master or is_masterless:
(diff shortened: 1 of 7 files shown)


giorgiosironi:
Is there a usecase? Not for encryption in general, but is there a project that will be using it? 

`elife-xpub`, see link in description

> And will it be strict about what it logs (because EBS isn't encrypted-at-rest and Loggly isn't either)?

Application concern, which I'll share (can't do much from the infrastructure side)

> And will it affect restoring backups at all? (will the backup itself be encrypted?)

As far as I know RDS encryption is equivalent to encrypting the disk of your laptop, so [backups made through a snapshot (targets directly the disk) are also encrypted](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.Encryption.html#Overview.Encryption.Overview). A backup made with UBR wouldn't be, by default.

lsh-0:
thanks for the link - I'll do some more reading. I was under the impression that AWS already went to some lengths to ensure volumes were encrypted at rest, but this is probably something else

giorgiosironi:
Note: I didn't give any specific permissions to AWS over the KMS key, so I think it was picked up by CloudFormation as it was my IAM user `GiorgioSironi` adding the database.

PR #395, opened 2018-09-25
Deploy RDS with encryption in elife-xpub--prod

giorgiosironi:
For https://github.com/elifesciences/elife-xpub/issues/623

Implemented in https://github.com/elifesciences/builder/pull/394

This needed deletion (don't try this at home) and recreation of the RDS instance.


giorgiosironi:
Already applied, so should not be kept open.

PR #396, opened 2018-09-25
Error on generic Cloud Formation bad request

giorgiosironi:
We whitelist an error related to idempotence: if there are no updates we should continue without errors. All the other errors should be raised. Difficult to trigger, by the way.

### DIFF
--- a/src/buildercore/cloudformation.py
+++ b/src/buildercore/cloudformation.py
@@ -215,7 +215,12 @@ def destroy(stackname, context):
         meta = ex.response['ResponseMetadata']
         err = ex.response['Error']
         # ll: [400: ValidationError] No updates are to be performed (request-id: dc28fd8f-4456-11e8-8851-d9346a742012)
+        if "No updates are to be performed" in err['Message']:
+            return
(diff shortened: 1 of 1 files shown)


lsh-0:
> Can this test be made more robust?

If this is what we think it is then we could make more of an effort to ensure initial state is clean and isn't dirtying the test. I haven't visited this code in a bit but it seems reasonable.

giorgiosironi:
Merging this later after I've finished trying to optimize the build.

giorgiosironi:
Not that I had time to optimize the build...

giorgiosironi:
Interesting failure in `master`:
```
Exceeded quota of account 512686554592 (Service: AmazonElasticLoadBalancing; Status Code: 400; Error Code: TooManyLoadBalancers; Request ID: 220919e6-c65d-11e8-a445-37b5a6cc30f8)
```

giorgiosironi:
Mostly they derive from a bunch of CloudFormation stacks left over from today's failed builds:
https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks?filter=active&stackId=arn:aws:cloudformation:us-east-1:512686554592:stack%2Fdummy1--jenkins-20181002140051-761523%2F92bd81a0-c64b-11e8-91b1-500c3d4416c5
Deleting them, but there may be more.

PR #399, opened 2018-09-27
[WIP] Make the Journal CDN fail faster if there's problems

thewilkybarkid:


projects/elife.yaml
```diff
@@ -474,8 +474,11 @@ journal:
                     - "{instance}--cdn-journal"
                 healthcheck:
                     path: /ping-fastly
-                    check-interval: 10000
-                    timeout: 2000
+                    check-interval: 5000
+                    timeout: 1000
+                between-bytes-timeout: 5000
+                error-threshold: 2
```
thewilkybarkid: Not sure what a sensible figure would be here, but probably not this low.

projects/elife.yaml
```diff
@@ -387,8 +387,8 @@ api-gateway:
                 - "{instance}--cdn-gateway"
             healthcheck:
                 path: /ping-fastly
-                check-interval: 10000
-                timeout: 2000
+                check-interval: 5000
+                timeout: 1000
```
giorgiosironi: https://github.com/elifesciences/builder/pull/310 set this to 2 seconds because not all POPs in the world can reach the US in 1 second

giorgiosironi: Even with shielding on they would still have to reach a us-east POP :thinking:

thewilkybarkid: Over faster connections I think. But they still do have the healthcheck to the origin.

src/buildercore/cfngen.py
```diff
@@ -400,6 +400,9 @@ def _parameterize_gcslogging(gcslogging):
             'dns': pdata['aws']['fastly']['dns'],
             'default-ttl': pdata['aws']['fastly']['default-ttl'],
             'healthcheck': pdata['aws']['fastly']['healthcheck'],
+            'between-bytes-timeout': pdata['aws']['fastly'].get('between-bytes-timeout'),
```
giorgiosironi: these are all options to be applied to all backends (if multiple ones are defined); so the level of nesting seems ok. The naming should be kept as Fastly defines these, with no translation

PR #400, opened 2018-09-27
Basic support for BigQuery

giorgiosironi:
For https://github.com/elifesciences/issues/issues/4461

Adds a `gcp` top-level configuration akin to `aws` (with no support for `gcp-alt` yet). This key allows to configure one or more BigQuery datasets, each with zero or more tables. Each table has a schema to be stored inside a folder here with a codename.

Has been used to create the [`iiif` table](https://console.cloud.google.com/bigquery?authuser=1&folder=&organizationId=567577722038&project=elife-fastly&p=elife-fastly&d=test&t=iiif&page=table) in the `test` dataset in the `elife-fastly` project.

Various TODOs are left throughout the code to be addressed before merging.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -2070,12 +2070,11 @@ fastly-logs:
                 project: elife-fastly
     gcp:
         bigquery:
-            datasets:
-                "{instance}":
-                    project: elife-fastly
-                    tables:
-                        iiif:
-                            schema: fastly-logs-201809
+            "{instance}":
+                project: elife-fastly
+                tables:
+                    iiif:
+                        schema: fastly-logs-201809
     aws-alt: {}
 
 data-pipeline:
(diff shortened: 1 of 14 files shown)


de-code:
FYI the current BigQuery tables for the eJP data are uploaded with an automatically detected schema and it's replacing the existing tables. What does definitely need to be created in advance is the dataset (and the project of course).

giorgiosironi:
> What does definitely need to be created in advance is the dataset (and the project of course).

We support this use case: empty dataset associated to a project. The one I'm targeting however is Fastly logs, which have an explicit schema.

lsh-0:
I wouldn't have coupled the fastly schema to the builder project.

giorgiosironi:
> the fastly schema

The BigQuery schema? It could go into some sort of formula, how could it then be deployed?

giorgiosironi:
(it has to remain in sync with https://github.com/elifesciences/builder/blob/master/src/buildercore/terraform.py#L39-L63)

PR #401, opened 2018-10-01
Introduce elife-xpub--staging

giorgiosironi:
Identical to prod, but:

- [x] no `multi-az`, halves the cost
- [x] no long backup retention
- [ ] possibly could use a smaller instance? No, `db.t2.micro` doesn't support encryption at rest.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1975,6 +1975,17 @@ elife-xpub:
         demo:
             s3:
                 "{instance}-elife-xpub":
+        staging:
+            s3:
+                "{instance}-elife-xpub":
+            rds:
+                engine: postgres # or 'MySQL'
+                version: '10.4'
+                type: db.t2.small
+                storage: 5 # GB
+                storage-type: gp2
+                backup-retention: 2 # days
+                encryption: arn:aws:kms:us-east-1:512686554592:key/b626157a-1e5b-4bf6-8799-211505efe072 
         prod:
             subdomains:
                 - xpub


projects/elife.yaml
```diff
@@ -1975,6 +1975,17 @@ elife-xpub:
         demo:
             s3:
                 "{instance}-elife-xpub":
+        staging:
+            s3:
+                "{instance}-elife-xpub":
+            rds:
+                engine: postgres # or 'MySQL'
+                version: '10.4'
+                type: db.t2.small
```
giorgiosironi: could use `db.t2.micro` (1 GB of RAM instead of 2 GB)?

projects/elife.yaml
```diff
@@ -1975,6 +1975,17 @@ elife-xpub:
...
+                storage: 5 # GB
+                storage-type: gp2
```
giorgiosironi: 5GB of this storage cost ~$0.50/month, so I'm not going to bother lowering it

projects/elife.yaml
```diff
@@ -1975,6 +1975,17 @@ elife-xpub:
...
+                backup-retention: 2 # days
+                encryption: arn:aws:kms:us-east-1:512686554592:key/b626157a-1e5b-4bf6-8799-211505efe072
```
giorgiosironi: I assume it could be feasible to import a dump of production here sooner or later, so enabling encryption too

projects/elife.yaml
damnhipster: As long as we document that staging is not prepared for load tests.

giorgiosironi: Definitely, a separate temporary environment perfectly replicating `prod` should be used for that.

giorgiosironi:
```
DB Instance class db.t2.micro does not support encryption at rest (Service: AmazonRDS; Status Code: 400; Error Code: InvalidParameterCombination; Request ID: d53eab6a-0
f87-4ab7-bc90-aaa4255a74b1)'
```

diversemix:
@giorgiosironi  - so do we need a larger db instance? Is this blocked now?

damnhipster:
We can use `db.t2.small` instead if the cost isn't significant.

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.Encryption.html#Overview.Encryption.Enabling

giorgiosironi:
Yes, `db.t2.small` was the original choice so I just reverted my attempt. 
https://staging--xpub.elifesciences.org/ is now available.

diversemix:
@giorgiosironi Great Work! ... looking in the footer it appears to be an old build

PR #402, opened 2018-10-04
support for specifying non-local path to bigquery schema files

lsh-0:
Mostly some very boring filename wrangling. 

The `schema` attribute will now download any schemas to the project's state directory `./cfn/bigquery-schema/stackname/whatever.json` if the path is prefixed with `https`. If the path is prefixed with `./`, it will look for the schema relative to the project root. This preserves fastly and testing schemas until we have nicer homes for them. Finally, you could just have 'schemaname.json' in the project definition, but nothing is writing schema files there during project init, unless we wanted to inline their content which is ... a meh idea.

I think returning paths-to-files is better than returning open file handles.


lsh-0:
@giorgiosironi , you have a problem with your terraform code:

```
************* Module buildercore.terraform
E:208, 8: Instance of 'OrderedDict' has no 'extend' member (no-member)
E:222, 8: Instance of 'OrderedDict' has no 'append' member (no-member)
```

You've declared it as an ordereddict but you're using it as a list

src/buildercore/bigquery.py
```diff
@@ -1,4 +1,34 @@
...
-def schema(schema_name):
-    return open('%s/%s.json' % (BIGQUERY_SCHEMAS_FOLDER, schema_name))
+LOG = logging.getLogger(__name__)
+
+def local_schema_path(stackname, schema_name):
+    "returns a local path to the given `schema_name`"
+    # "/.../.cfn/bigquery-schema/elife-data-pipeline--prod/fooschema.json"
+    return join(config.BIGQUERY_SCHEMA_DIR, stackname, schema_name)
+
+def fetch_remote_schema(stackname, file_url):
```
giorgiosironi: should use [the Terraform HTTP data source](https://www.terraform.io/docs/providers/http/data_source.html) to avoid hand-rolling the download, we use it already [for HTML error pages](https://github.com/elifesciences/builder/blob/master/src/buildercore/terraform.py#L289-L291)

lsh-0: good call

projects/elife.yaml
```diff
@@ -2099,3 +2099,17 @@ data-pipeline:
...
             "{instance}-elife-data-pipeline-archive":
                  deletion-policy: retain
+    gcp:
+        bigquery:
+            "{instance}": # dataset
+                # does 'project' belong beneath dataset? do we want an instance affecting multiple projects?
+                project: elife-data-pipeline
+                tables:
+                    # bit of repetition here
+                    380_datascience_early_career_researchers:
```
giorgiosironi: storing the schema here is supported, but if the project needs to change it often (as in every deploy), it should be able to not specify one; not sure how a migration process works for BigQuery?

giorgiosironi:
> ************* Module buildercore.terraform
> E:208, 8: Instance of 'OrderedDict' has no 'extend' member (no-member)
> E:222, 8: Instance of 'OrderedDict' has no 'append' member (no-member)

This field is initialized as a list:
```
         tf_file['resource'][RESOURCE_TYPE_FASTLY][RESOURCE_NAME_FASTLY]['vcl'] = [
             {
                 'name': snippet_name,
                 'content': _generate_vcl_file(context['stackname'], fastly.VCL_SNIPPETS[snippet_name].content, snippet_name),
             } for snippet_name in vcl_constant_snippets
]
```

Then `extend` and `append` are used to build it. Where does the `OrderedDict` come from, some other place in the code?

lsh-0:
I think it may be this line: https://github.com/elifesciences/builder/blob/master/src/buildercore/terraform.py#L161

giorgiosironi:
Yes, that's the wrong one; even if it gets overwritten, pylint correctly sees the difference in type.

projects/elife.yaml
lsh-0: I was thinking the flow would ideally be like this: 

1. remote schema changes
2. ./bldr update_infrastructure:stackname
3. terraform is given a remote path to the schema, pulls it down, and tells you if changes are to be made
4. terraform applies changes

we're not managing the schema of other databases through builder, we leave it to the application typically. I wonder if we should at the dataset level - just ensure the project+dataset exists and let the app deal with table creation

lsh-0: or maybe we just create and delete tables and that's it, no updates

lsh-0:
@giorgiosironi , the code has changed significantly from when I first opened it. Could you cast your eye over it again now that it's passing tests?

.ci/pylint
```diff
@@ -9,11 +9,9 @@ find src/ -name '*.py[c|~]' -delete
 find src/ -regex "\(.*__pycache__.*\|*.py[co]\)" -delete
 
 # E1129=doesn't like Fabric's context manager
-# E1101=no-member, temporary while python2 is still supported
```
giorgiosironi: python2 is still supported, but if this can be removed, all the better

projects/elife.yaml
giorgiosironi: We can create the dataset only if the project requires it, it's already supported. If the schema changes are so involved though then having them in `builder` is not a big limitation, as they need coordination during deployments anyway.

projects/elife.yaml
```diff
@@ -2099,3 +2099,17 @@ data-pipeline:
...
+                        schema: https://github.com/elifesciences/data-pipeline/schemas/380_datascience_early_career_researchers.json
+                    489_datascience_editor_keywords:
+                        schema: https://github.com/elifesciences/data-pipeline/schemas/489_datascience_editor_keywords.json
+                    455_datascience_editors:
+                        schema: https://github.com/elifesciences/data-pipeline/schemas/455_datascience_editors.json
```
giorgiosironi: gives me `404`, where is this file?

projects/elife.yaml
```diff
@@ -2099,3 +2099,17 @@ data-pipeline:
                  deletion-policy: delete
             "{instance}-elife-data-pipeline-archive":
                  deletion-policy: retain
+    gcp:
+        bigquery:
+            "{instance}": # dataset
+                # does 'project' belong beneath dataset? do we want an instance affecting multiple projects?
```
giorgiosironi: I put datasets at the top level as there doesn't seem to be a resource to create that is spanning multiple datasets. The alternative is in fact to move them into a `datasets` key and have `project` and possibly other defaults right inside `bigquery`. We can probably change this later when we know more, it only propagates up to the `cfngen` module and then `context` hides this anyway.

src/buildercore/bigquery.py
```diff
@@ -1,4 +0,0 @@
-BIGQUERY_SCHEMAS_FOLDER = 'src/buildercore/bigquery/schemas'
```
giorgiosironi: like `buildercore.fastly`, this module is meant to store BigQuery-specific code that is not Terraform-specific (and hence would go into `buildercore.terraform`). Let's see what emerges

src/buildercore/terraform.py
```diff
@@ -362,7 +359,7 @@ def _generate_vcl_file(stackname, content, key, extension='vcl'):
     """
     with _open(stackname, key, extension=extension, mode='w') as fp:
         fp.write(str(content))
-        return '${file("%s")}' % basename(fp.name)
+        return '${file("%s")}' % os.path.basename(fp.name)
```
giorgiosironi: `from os.path import join` is inconsistent with `os.path.basename()`, which is the chosen style?

src/buildercore/terraform.py
```diff
@@ -381,7 +378,7 @@ def render_gcs(context):
         },
     }
 
-def render_bigquery(context):
+def render_bigquery(context, tf_file):
```
giorgiosironi: I agree this parameter is too thorny, as it introduces a coupling from a previous `render_*()` (which has to create the `data` key) to this one. I'd rather either do a not-too-clever merge:
```
tf_file = {'resources': {}, 'data': {}}
renders = [render_fastly, render_bigquery, ...]
partials = [r(c) for r in renders]
for p in partials:
    if p['resources']:
        tf_file['resources'] = tf_file['resources'].merge(p['resources'])
    if p['data']:
        tf_file['data'] = tf_file['data'].merge(p['data'])
// remove 'data' and 'resources' if empty
```
More long-term, we could even put each partial in its own `.tf` file as Terraform should be able to resolve them all into a single set of resources, data, etc. That's too involved for this PR however.

src/buildercore/terraform.py
```diff
@@ -401,27 +398,38 @@ def render_bigquery(context):
         } for dataset_id, options in context['bigquery'].items()
     }
 
-    if tables:
+    def add_table(table_id, table_options):
+        schema = table_options['schema']
+        stackname = context['stackname']
+        fqrn = "%s_%s" % (table_options['dataset_id'], table_id) # 'fully qualified resource name'
```
giorgiosironi: it was not any standard terminogy, could just be called `unique_name` while inside this scope

src/buildercore/terraform.py
```diff
@@ -401,27 +398,38 @@ def render_bigquery(context):
...
+
+        if schema.startswith('https://'):
+            # remote schema, add a 'http' provider and have terraform pull it down for us
+            # https://www.terraform.io/docs/providers/http/data_source.html
+            tf_file['data'][DATA_TYPE_HTTP][fqrn] = {'url': schema}
+            schema_ref = '${data.http.%s.body}' % fqrn
+
+        else:
+            # local schema. the `schema` is relative to `PROJECT_PATH`
+            schema_path = join(PROJECT_PATH, schema)
```
giorgiosironi: given `PROJECT_PATH` is the current working directory, this `join` should not be necessary

src/buildercore/terraform.py
```diff
@@ -401,27 +398,38 @@ def render_bigquery(context):
...
+            schema_file = os.path.basename(schema)
+            terraform_schema_dir = join(TERRAFORM_DIR, stackname)
+            mkdir_p(terraform_schema_dir)
+            shutil.copyfile(schema_path, join(terraform_schema_dir, schema_file))
```
giorgiosironi: this should be easier than reading/writing the file

src/buildercore/terraform.py
```diff
@@ -550,14 +558,17 @@ def destroy(stackname, context):
     terraform_directory = join(TERRAFORM_DIR, stackname)
     shutil.rmtree(terraform_directory)
 
+# TODO: not a great function name. 'stack_tform_path' ? 'tform_stackfile_path' ?
 def _file_path(stackname, name, extension='tf.json'):
```
giorgiosironi: Terraform has no concept of stacks, but each of these is a `configuration file`, so `_configuration_file_path`.

With hungarian notation it could be `_tf_configuration_file_path` would be clear for some cases, but it is really responsible for any file that ends up there, even if it is a `.vcl` (Varnish Configuration Language). `tform` is not in use anywhere.

giorgiosironi: (`_single_leading_underscore` is from [PEP 8](https://www.python.org/dev/peps/pep-0008/))

src/buildercore/terraform.py
```diff
@@ -550,14 +558,17 @@ def destroy(stackname, context):
...
     return join(TERRAFORM_DIR, stackname, '%s.%s' % (name, extension))
 
 def _open(stackname, name, extension='tf.json', mode='r'):
+    "`open`s a file in the conf.TERRAFORM_DIR belonging to given `stackname` (./cfn/terraform/$stackname/)"
```
giorgiosironi: the directory is `.cfn`

src/buildercore/terraform.py
giorgiosironi: personally I use more `os.path.*(...)` calls as they declare their dependency

projects/elife.yaml
lsh-0: doesn't exist yet

src/buildercore/terraform.py
lsh-0: implicit vs explicit. joining with `PROJECT_PATH` makes an absolute path

src/buildercore/terraform.py
lsh-0: right you are

src/buildercore/terraform.py
lsh-0: > from os.path import join is inconsistent with os.path.basename(), which is the chosen style?

yeah, it's an exception to the rule. It *is* consistent with all my other code though. And I've not seen `from os.path import basename` anywhere else (yet) at elife. I don't really care though, I change code as I read through it and this was just one case I couldn't be bothered reverting before commit. I'll leave it be from now on if I see it happening elsewhere.

src/buildercore/bigquery.py
lsh-0: If there was any significant logic in there it would have been preserved. Should the BigQuery wrangling get large we can always re-create it

src/buildercore/terraform.py
lsh-0: We should look to cfngen.py and trop.py for how we're handling this pattern already. 

https://github.com/elifesciences/builder/blob/master/src/buildercore/cfngen.py#L133-L156
https://github.com/elifesciences/builder/blob/master/src/buildercore/trop.py#L928-L946

wow - trop.py hasn't gotten *big*. we don't have too many individual files at 1k+ loc

.ci/pylint
lsh-0: I use python2 locally and all was fine. I pushed the pylint disable comment into the file itself, so it's still there, just not affecting the *entire* project

src/buildercore/terraform.py
de-code: In general I would prefer `from ... import ...` but _join_(`from os.path import join` ) is too generic. `pathlib` can help with that and we should use it more often I think.

src/buildercore/bigquery.py
giorgiosironi: Given the delegation to Terraform of most of the file-loading stuff, this module should not be needed for now

src/buildercore/terraform.py
giorgiosironi: `cfngen` doesn't match to a good example in my view, as the `context` it creates its our own data structure, rather than an external one we have to conform to; in fact, it is designed with no overlapping keys e.g. `s3`, `ec2`, ... and the rendering functions are all independent.

`trop` instead passes in a `template` object that the `render_*()` function can write to, while returning `None`. The `tf_file` dictionary however runs the risk that a `render_2()` either:

- overwrites the tf_file['data']['http'] of another `render_1()`
- does lots of null checks `if 'http' in tf_file['data']:` to add to or create the subdictionaries

as there are many overlaps of a single rendering function providing some `resource` and some `data` keys.

`trop` solves this problem by providing a [`Template` class](https://github.com/cloudtools/troposphere/blob/master/troposphere/__init__.py#L552) that can receive `add_resource` and `add_output` calls and be dumped to a dictionary when fully populated. This is a good solution for this case.

src/buildercore/terraform.py
lsh-0: "too generic" - I disagree. There is no builtin called "join" and I've never written a function called "join" nor kicked myself for shadowing the imported "join" with another's "join" function.

I think this is getting out of hand. Beyond the stuff controlled by `autopep8` can we agree that different authors of code have different styles? and that coercing others to conform to your own personal style so it's easier for *you* to read, as if you were some sort of authority on universal code readability, is a jerky thing to do? I try not to pick at other's style beyond the obvious bad patterns, but this time I trod on @giorgiosironi toes.

src/buildercore/terraform.py
lsh-0: you know - I almost apologised in my last comment but I take that almost-apology back ;)

objects don't horrify me, they're just almost always the wrong answer to any problem. 

the examples I gave were not meant to be exact fits, but examples of building up a data struct (or object) by passing through it through a simple list of functions. I'll take a look at this chunk of code today.

projects/elife.yaml
lsh-0: PR here: https://github.com/elifesciences/data-pipeline/pull/2

src/buildercore/terraform.py
lsh-0: ... the emphatic 'you' wasn't any specific individual.

src/buildercore/terraform.py
lsh-0: > as there are many overlaps of a single rendering function providing some resource and some data keys.

yes, I was just looking at the code and your `render_fastly` might have it's `resource`s nuked by `render_gcs` that comes next. I mean right now it's probably not a problem as a fastly project doesn't have any buckets (or vice versa). A nice testable version would be the fastly, gcs, bigquery stuff all generating the `data` and `resources` struct that they need and then merging the result afterwards.

src/buildercore/terraform.py
giorgiosironi: That's why we use `autopep8` and other coding standard tools, so that there is an arbitrary choice made by it and people don't have to discuss it :) so if it's not enforced there, there is no need to edit war - I see this mostly as knowledge sharing as I have become aware of `pathlib` for example which I didn't know before

src/buildercore/terraform.py
giorgiosironi: The long-term version should not need to merge them as they can be stored in different `.tf` files, I'm looking for a safe way to integrate this to avoid divergence in the `buildercore.terraform` module.

giorgiosironi: > A nice testable version would be the fastly, gcs, bigquery stuff all generating the data and resources

I have no horse in this game, so if you implement this procedural solution with some logic in this module, I'll approve it as it's much better than the coupling on `tf_file`

lsh-0: I didn't know about indivudlal .tf files, definitely a better solution in the end than more-code. For now though I'll just get something safe and workable happening

lsh-0:
I'm a little uncertain about the merging strategy I've used (same as that used by the project definitions), but it is passing tests ... I'll resolve the conflict and maybe add a test before merging.

lsh-0:
alright, there is a test now and it revealed a dumb bug I've fixed. will merge once green

src/tests/test_buildercore_terraform.py
```diff
@@ -470,6 +470,46 @@ def test_bigquery_full_template(self):
...
         })
 
+    def test_bigquery_remote_paths(self):
+        "remote paths require terraform to fetch and load the files, which requires another entry in the 'data' list"
+        pname = "project-with-bigquery-remote-schemas"
+        iid = pname + "--prod"
+        context = cfngen.build_context(pname, stackname=iid)
+        terraform_template = json.loads(terraform.render(context))
+
+        expecting = json.loads('''{
```
giorgiosironi: this is less effective than the standard of this file to use dictionaries; syntax errors in the JSON produce `ValueError`s with the JSON line numbering rather than the `.py` file lines, cannot put comments into it, `"` needs double escaping, etc.

Won't hold this up but needs to be uniformed to the rest

giorgiosironi:
I'll merge and deal with the expected JSON in my PR, to avoid code conflicts over the `terraform` module and its tests.

PR #403, opened 2018-10-09
Feature: data pipeline ext space

lsh-0:


projects/elife.yaml
```diff
@@ -2109,6 +2109,10 @@ data-pipeline:
                  deletion-policy: delete
             "{instance}-elife-data-pipeline-archive":
                  deletion-policy: retain
+    aws-alt:
+        prod:
+            ext:
+                size: 100 # GB
```
giorgiosironi: consider also `type: ssd` if it's not the default, if there are performance risks. If it's a staging area for files no worries, magnetic would be enough

lsh-0: I'm a bit fuzzy about what will pass through there to be honest. Nifi has backpressure and throughput controls and once it's configured it just hums along. Perhaps once there are more flows happening and competing for resources we can look at speeding up disk access, or bumping the instance type.

giorgiosironi:
Can this be updated and merged?

lsh-0:
absolutely, looks like it got forgotten about

PR #404, opened 2018-10-09
S3 encryption at rest

giorgiosironi:
For https://github.com/elifesciences/elife-xpub/issues/704

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1976,12 +1976,10 @@ elife-xpub:
             device: /dev/sdh
             type: gp2
     aws-alt:
-        demo:
-            s3:
-                "{instance}-elife-xpub":
         staging:
             s3:
                 "{instance}-elife-xpub":
+                    encryption: arn:aws:kms:us-east-1:512686554592:key/b6f44e77-46d8-41bb-a2e5-75c0588a2b20 
             rds:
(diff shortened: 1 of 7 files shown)


diversemix:
Are you happy to store the key alongside others? If we add this to the docs .... where shall we say it is stored?

projects/elife.yaml
```diff
@@ -1975,12 +1976,10 @@ elife-xpub:
             device: /dev/sdh
             type: gp2
     aws-alt:
-        demo:
-            s3:
-                "{instance}-elife-xpub":
         staging:
```
diversemix: Is `prod` getting added in another PR?

giorgiosironi:
> Are you happy to store the key alongside others?

It is stored in the KMS console which a few IT admins can access; the idea of KMS being it stores it for you rather than having to set up yet another Git repo and its permissions.

projects/elife.yaml
giorgiosironi: This is the deletion of `demo` which is configured but not used; both `staging` and `prod` have been configured to use `encryption`, look at the next additions in this file.

giorgiosironi:
Previous of [KMS in RDS encryption at rest](https://github.com/elifesciences/builder/pull/394#issuecomment-424545203)

PR #405, opened 2018-10-10
Feature: more convenient uploads

lsh-0:
* adds node parameter to 'upload_file' task
* adds a default destination if one not specified

### DIFF
--- a/src/cfn.py
+++ b/src/cfn.py
@@ -303,6 +303,7 @@ def download_file(stackname, path, destination='.', node=None, allow_missing="Fa
 @requires_aws_stack
 def upload_file(stackname, local_path, remote_path=None, overwrite=False, node=1):
     remote_path = remote_path or os.path.join("/tmp", os.path.basename(local_path))
+    overwrite, node = str(overwrite).lower() == "true", int(node)
     with stack_conn(stackname, node=node):
         print('stack:', stackname, 'node', node)
         print('local:', local_path)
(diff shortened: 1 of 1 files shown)


src/cfn.py
```diff
@@ -300,9 +301,10 @@ def _download(path, destination):
 
 @task
 @requires_aws_stack
-def upload_file(stackname, local_path, remote_path, overwrite=False):
-    with stack_conn(stackname):
-        print('stack:', stackname)
+def upload_file(stackname, local_path, remote_path=None, overwrite=False, node=1):
+    remote_path = remote_path or os.path.join("/tmp", os.path.basename(local_path))
+    with stack_conn(stackname, node=node):
```
giorgiosironi: didn't remember this supported `node`

It seems it is forced to `int` inside the function (`node` here will be a string)

src/cfn.py
```diff
@@ -300,9 +301,10 @@ def _download(path, destination):
...
+        print('stack:', stackname, 'node', node)
```
giorgiosironi: maybe we should start using a logger here

src/cfn.py
lsh-0: > didn't remember this supported node

`stack_conn` supporting a `node` parameter is something I introduced relatively recently - this year some time. I think it was so I could ssh into a specific node in one command

good catch with the typing. I'll fix that up

src/cfn.py
lsh-0: it's not a debug statement, it's just for the user. if anything it should be on stderr. if we have a logger that doesn't emit any prefix that might be ok, but the benefits of that would be quite tiny.

PR #406, opened 2018-10-10
Reduce default Fastly TTL

thewilkybarkid:
If a response doesn't have a `Cache-Control` header it should be treated as misconfiguration. But in the meantime, a 1-hour TTL can create problems (especially with shielding turn on), so this reduces it to 5 minutes.

Checklist:

- [x] `api-gateway--end2end`
- [x] `api-gateway--continuumtest`
- [x] `api-gateway--prod`
- [x] `generic-cdn--end2end`
- [x] `generic-cdn--continuumtest`
- [x] `generic-cdn--prod`
- [x] `iiif--end2end`
- [x] `iiif--continuumtest`
- [x] `iiif--prod`
- [x] `journal--end2end`
- [x] `journal--continuumtest`
- [x] `journal--prod`

### DIFF
--- a/src/buildercore/fastly/vcl/main.vcl
+++ b/src/buildercore/fastly/vcl/main.vcl
@@ -64,7 +64,7 @@ sub vcl_fetch {
     # keep the ttl here
   } else {
     # apply the default ttl
-    set beresp.ttl = 3600s;
+    set beresp.ttl = 300s;
   }
 
   return(deliver);


giorgiosironi:
Going to roll out and merge this today, while we are not modifying other Fastly configurations.

giorgiosironi:
Incidentally, `journal` is also updating error pages in the generated VCL.

giorgiosironi:
Applied everywhere, checklist in description.

giorgiosironi:
Scenario encountered today: when the backend is S3 there is (usually?) no `Cache-Control` header, hence this default applies.

PR #407, opened 2018-10-10
Always treat aws-alt keys as strings

giorgiosironi:
Otherwise `alt_config` arguments passed to `launch` are not matched, being strings vs. YAML integers

### DIFF
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -578,6 +578,9 @@ def find_region(stackname=None):
     if more than one region available, it will raise an MultipleRegionsError.
     until we have some means of supporting multiple regions, this is the best solution"""
     if stackname:
+        # TODO: should use context, not project data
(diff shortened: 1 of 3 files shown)


giorgiosironi:
Created https://alfred.elifesciences.org/job/process/job/process-basebox-1804/ which will need this to run I think. Once it runs, it may reproduce the failure of creating a fresh `basebox--1804` to then generate an AMI.

src/buildercore/core.py
```diff
@@ -578,6 +578,9 @@ def find_region(stackname=None):
     if more than one region available, it will raise an MultipleRegionsError.
     until we have some means of supporting multiple regions, this is the best solution"""
     if stackname:
+        # TODO: should use context, not project data
```
lsh-0: :+1:

PR #408, opened 2018-10-10
Allow destroying someone else's stack

giorgiosironi:
We cannot rely on a local `.cfn/stacks/*.json` file, as it may have been
launched from another node. Better to use as the source of truth the
CloudFormation API, which we were calling anyway.

Other operations don't have to rely on a template at all as they can regenerate it from `context` which is synced through S3. For destroy this would be overkill, especially if the stack is in a broken state.


giorgiosironi:
Like https://github.com/elifesciences/builder/pull/407 it blocks https://alfred.elifesciences.org/job/process/job/process-basebox-1804

giorgiosironi:
As a test I tried destroying `peerscout--reloadtest3` which was a test from August, seems to work without me having related files in `.cfn/stacks/`.

lsh-0:
ah! I was making similar changes on Weds to destroy a stack made on my laptop that wasn't present on my desktop. I had to revert them during git wrangling though. Will merge

PR #409, opened 2018-10-11
Configure elife-xpub--end2end

giorgiosironi:
For https://github.com/elifesciences/elife-xpub/issues/648

Needs configuration before it can be created.


giorgiosironi:
Has now been created.

PR #410, opened 2018-10-11
Subscribe journal-cms to bus for digests

giorgiosironi:
For https://github.com/elifesciences/issues/issues/4294

Will need to manually apply this, but probably a deployment must be made first to accept the new kind of data that is coming on the queue.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -78,6 +78,7 @@ defaults:
             backup-retention: 28 # days
             # if rds.params are specified, a custom db parameter group is created
             params: []
+            encryption: false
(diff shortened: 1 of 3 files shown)


giorgiosironi:
Taking a bit longer to update as it has to apply the new default `'StorageEncrypted': 'false'` to all RDS instances, should be no actual change.

giorgiosironi:
Failed on `end2end` because of the RDS "requiring replacement", which should not happen. Will remove the field altogether when not encrypted as in this case :fire:

giorgiosironi:
Has been applied everywhere.

PR #411, opened 2018-10-17
Remove various possibly dead projects

giorgiosironi:
No instances of them around, neither soon to be created.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1978,26 +1978,6 @@ elife-xpub:
         ports:
             1201: 80
 
-elife-ink:
-    subdomain: ink
-    repo: git@github.com:elifesciences/elife-ink
-    formula-repo: git@github.com:elifesciences/elife-ink-formula
-    aws:
-        type: t2.medium
-        ports:
-            - 22
-            - 443 # ink
-            - 3000 # ink
-            - 4567 # slanger
-            - 4569 # slanger nginx https
-            - 8080 # slanger api
-            - 8082 # slanger api nginx https
-    vagrant:
-        ram: 4096
-        ports:
-            1223: 80
-
-
 task-adept:
     description: production data dumps from TaskAdept
     default-branch: null
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1988,38 +1988,6 @@ task-adept:
         s3:
             "{instance}-elife-task-adept":
 
-exeter-adapter:
-    description: Exeter history API integration
-    domain: false
-    intdomain: False
-    aws:
-        ec2: false
-        sqs:
-            exeter-adapter--{instance}: {}
-    aws-alt:
-        fresh:
-            ec2: false
-        1604:
-            ec2: false
-        1404:
-            ec2: false
-        standalone1604:
-            ec2: false
-        standalone1404:
-            ec2: false
-        standalone:
-            ec2: false
-        continuumtest:
-            sqs:
-                exeter-adapter--{instance}:
-                    subscriptions:
-                        - elife-bot-event-property--continuumtest
-        prod:
-            sqs:
-                exeter-adapter--{instance}:
-                    subscriptions:
-                        - elife-bot-event-property--prod
-
 fastly-logs:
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -1016,14 +999,6 @@ elife-api:
             1233: 80
 
 
-elife-reporting:
```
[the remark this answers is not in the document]
lsh-0: that should be fine, there is no dependency on the project definition itself. And soon, hopefully, we might be able to detach the ejp scraping from lax (cc @de-code). The ejp scraping has been used to pre-populate lax with article 'stubs' and some things like submission and acceptance dates, that are in turn queried by Paul Kelly and exported to his dashboard. I think there are also RSS feeds that a pre-production service is using from lax that is using ejp data. I might make a 'containerise lax' epic-scoped task and fill it with these issues.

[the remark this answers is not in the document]
lsh-0: actually, I'm reverting elife-reporting. I thought you were removing it because it had aws configuration. elife-reporting is a working environment of what production/James use to generate reports. It can be removed when James' reporting tools are no longer being used

lsh-0:
we have a "jats4r--2016-08-22" instance I can't access. I think the last time we looked it wasn't even being used. we're certainly not managing it in any reasonable way.

giorgiosironi:
> jats4r--2016-08-22

AFAIK it's an EC2 we have given over to the person in the comment to do what he wants - not managed by a formula

projects/elife.yaml
```diff
@@ -1015,7 +998,6 @@ elife-api:
         ports:
             1233: 80
 
-
```
giorgiosironi: good feedback, this is definitely used then

lsh-0: heh. looks like that was the day github was broken. it was a full 24 hours before it's status page was back to green.

giorgiosironi:
Added `elife-ink` which won't be used by Xpub anymore.

lsh-0:
that spreadsheet you've put up with projects and lifecycles is good, well done.

lsh-0:
Melissa has gotten back to us regarding jats4r. That inaccessible instance can be destroyed but future work is expected on the validator. When it does, I want it integrated into the elife CI like any other service, no exceptions. I think @nlisgo will be taking over jats4r validator duties.

The project def in this instance can be removed and we'll add it again when/as needed.

projects/elife.yaml
```diff
@@ -1192,16 +1174,6 @@ api-dummy:
         ports:
             1242: 80
 
-anonymous:
```
giorgiosironi: destroyed `anonymous--continuum`

projects/elife.yaml
```diff
@@ -1215,32 +1187,6 @@ jats4r:
             - 80
             - 443
 
-# distinct from the 'jats4r' project, this runs an instance of the validator only
-jats4r-validator:
```
giorgiosironi: no instances around

projects/elife.yaml
```diff
@@ -2071,26 +1978,6 @@ elife-xpub:
         ports:
             1201: 80
 
-elife-ink:
```
giorgiosironi: no instances around

projects/elife.yaml
```diff
@@ -2101,38 +1988,6 @@ task-adept:
         s3:
             "{instance}-elife-task-adept":
 
-exeter-adapter:
```
giorgiosironi: destroyed instances yesterday

projects/elife.yaml
```diff
@@ -2195,21 +2050,3 @@ data-pipeline:
         ram: 2048
         ports:
             1222: 80
-
-
-# deprecated, remove once all nifi-demo instances are destroyed
-nifi-demo:
-    formula-repo: https://github.com/elifesciences/nifi-demo-formula
```
giorgiosironi: no instances around

giorgiosironi:
Just cross-checked all the projects to ensure we have no leftovers. Good to merge.

lsh-0:
nice

PR #413, opened 2018-10-22
Idempotent, non-interactive launch

giorgiosironi:
```
[11:22:20][giorgio@Newton:~/code/builder]$ bldr launch:journal,pr-test
...
INFO - cfn - stack journal--pr-test already exists

Done.
```

### DIFF
--- a/src/buildercore/checks.py
+++ b/src/buildercore/checks.py
@@ -9,7 +9,9 @@ class AccessProblem(RuntimeError):
     pass
 
 class StackAlreadyExistsProblem(RuntimeError):
-    pass
+    def __init__(self, message, stackname):
+        RuntimeError.__init__(self, message)
+        self.stackname = stackname
 
 def can_access_builder_private(pname):
     "True if current user can access the private-repo for given project"
@@ -24,7 +26,7 @@ def ensure_can_access_builder_private(pname):
 
 def ensure_stack_does_not_exist(stackname):
     if core.stack_is_active(stackname):
-        raise StackAlreadyExistsProblem("%s is an active stack" % stackname)
+        raise StackAlreadyExistsProblem("%s is an active stack" % stackname, stackname)
 
 def requires_certificate(stackname):
     "True if the given stack has a subdomain but is not using a properly configured certificate"
--- a/src/tests/test_buildercore_s3.py
+++ b/src/tests/test_buildercore_s3.py
@@ -3,37 +3,39 @@ import os
 from . import base
 from buildercore import s3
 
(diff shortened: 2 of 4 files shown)


giorgiosironi:
Most of the PR is making the tests run fine with the recent parallelization.

src/buildercore/checks.py
```diff
@@ -9,7 +9,9 @@ class AccessProblem(RuntimeError):
     pass
 
 class StackAlreadyExistsProblem(RuntimeError):
-    pass
+    def __init__(self, message, stackname):
+        RuntimeError.__init__(self, message)
```
lsh-0: ah, another python3-ism: https://www.journaldev.com/15911/python-super

src/tests/test_buildercore_s3.py
```diff
@@ -3,56 +3,60 @@
...
+class TestReadWrite(base.BaseCase):
     def setUp(self):
-        s3.delete_contents("test/")
-        os.system("mkdir -p /tmp/builder/")
         self.envname = base.generate_environment_name()
 
+        self.prefix = "test/%s" % self.envname
+        s3.delete_contents("%s/" % self.prefix)
+
+        self.local_path = "/tmp/builder-%s/" % self.envname
```
lsh-0: we should replace this with `buildercore.utils.tempdir`

lsh-0:
replace that tempdir wrangling and I'm happy for this to be merged

src/buildercore/checks.py
giorgiosironi: can `super` be used in both Python versions?

lsh-0: looks like it. Python3:

```python3
Python 3.7.0 (default, Sep 15 2018, 19:13:07) 
[GCC 8.2.1 20180831] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> class Foo:
...     pass
... 
>>> class Bar(Foo):
...     def __init__(self):
...         super(Bar, self).__init__()
... 
>>> baz = Bar()
```

python2:
```python
Python 2.7.15 (default, Jun 27 2018, 13:05:28) 
[GCC 8.1.1 20180531] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> class Foo(object):
...     pass
... 
>>> class Bar(Foo):
...     def __init__(self):
...         super(Bar, self).__init__()
... 
>>> baz = Bar()
```

lsh-0: with python2 you have to use 'new style' classes though, like the above

PR #419, opened 2018-10-26
Update AMI for 1804 to ami-053d78b58ace2e62c

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/process/job/process-ami-update/4/display/redirect which resulted in this pull request.


giorgiosironi:
(another test, only the `.yaml` file would actually be changed)

PR #420, opened 2018-10-29
Avoid Terraform conflicts on existing/missing keys

giorgiosironi:
I initially thought it would be enough to separate each provider into its own `*.tf` file. I was wrong on this as even inside the same resource (`fastly_service_v1`) it is possible to get errors on:

- the various branches of the tree being missing (trying to configure `data.http.something` before `data.http` exists
- a branch existing when it shouldn't e.g. `data.http` or even `data` being rendered as an empty list which [causes a Terraform error](https://github.com/elifesciences/builder/pull/412/files#diff-c266af1366f88d308f6fcb68d77552bcR480)

This approach borrows from `troposphere` and the way it allows multiple rendering processes to populate a single template, even when different types are being added; in CloudFormation, `resources` vs. `outputs` while here `resources` vs. `data`.

### DIFF
--- a/src/buildercore/terraform.py
+++ b/src/buildercore/terraform.py
@@ -159,30 +159,6 @@ def render_fastly(context):
             shield=shield
         ))
 
-    tf_file = {
-        'resource': {
-            RESOURCE_TYPE_FASTLY: {
-                # must be unique but only in a certain context like this, use some constants
-                RESOURCE_NAME_FASTLY: {
-                    'name': context['stackname'],
-                    'domain': [
-                        {'name': subdomain} for subdomain in all_allowed_subdomains
-                    ],
-                    'backend': backends,
-                    'default_ttl': context['fastly']['default-ttl'],
-                    'gzip': {
-                        'name': 'default',
-                        # shouldn't need to replicate the defaults
-                        # https://github.com/terraform-providers/terraform-provider-fastly/issues/66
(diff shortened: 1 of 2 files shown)


giorgiosironi:
> there were ordering problems with the single data structure irregardless of whether it was eventually split across multiple .tf files

Not ordering, but dictionary key conflicts even inside the single `fastly` section. Here's an example:

```
    data = {}
    data[DATA_TYPE_VAULT_GENERIC_SECRET] = {}
    # render lots of stuff...
    if not data[DATA_TYPE_VAULT_GENERIC_SECRET]:
        del data[DATA_TYPE_VAULT_GENERIC_SECRET]
```
In some conditions (when `gcslogging` and/or `bigquerylogging` are used), this dictionary is populated by different parts of the code (implementing one or the other). If at least one is used, it works; but if the dictionary is empty it must be removed or `terraform apply` will fail.

On the other hand, if we don't initialize `data[DATA_TYPE_VAULT_GENERIC_SECRET]`, then every time a key needs to be added it would have to check:
```
if DATA_TYPE_VAULT_GENERIC_SECRET not in data:
    data[DATA_TYPE_VAULT_GENERIC_SECRET] = OrderedDict()
# now you can put a key into it...
```

This problem is repeated for `resource` at various levels. We centralize this check, so that it's very easy for `render_*()` functions to add new data or resources with a single call.

PR #426, opened 2018-11-08
Catch generic FabricException on boot of node

giorgiosironi:
For https://github.com/elifesciences/issues/issues/4573

### DIFF
--- a/src/buildercore/lifecycle.py
+++ b/src/buildercore/lifecycle.py
@@ -118,8 +118,9 @@ def start(stackname):
         call_while(
             lambda: _some_node_is_not_ready(stackname),
             interval=2,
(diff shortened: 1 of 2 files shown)


giorgiosironi:
```
>       return open(_file_path_for_generation(stackname, name, extension), mode)
E       IOError: [Errno 2] No such file or directory: '.cfn/terraform/dummy1--test/generated.tf.json'
```
Long-standing race condition to solve, we should use a generated environment name rather than `test` to prevent clashes.

giorgiosironi:
Started with one problem and had to fix another one; not the best way to parameterize all names but given Python 2 it's probably the best that can be done.

PR #427, opened 2018-11-12
Make context from S3 authoritative

giorgiosironi:
Current situation is you may have a old context around on your local disk. This applies to `elife-alfred--prod` as well. Making the context always downloaded from S3 avoids this stale copy causing weird bugs like https://alfred.elifesciences.org/job/process/job/process-master-server/7, but it may be slower.


giorgiosironi:
(This is what Terraform does btw: all state remotely stored on an S3 bucket)

PR #429, opened 2018-11-14
Simplify double retry loop

giorgiosironi:
`_some_node_is_not_ready` is already called via `call_while` so it's in a
retry loop for 120 seconds. `_wait_daemons` (renamed to `_daemons_ready`) inside it does not need to implement its own retry loop but just to fail quickly.


giorgiosironi:
Has been causing some apparently infinite (or just very long loop) e.g. https://alfred.elifesciences.org/job/test-elife-xpub/478/console so I'll merge to get simpler code to reason on.

PR #435, opened 2018-12-05
Fix Fastly percent encoding

thewilkybarkid:
Seems that the recent Unicode change in Fastly means that some of our regexs are broken (see https://github.com/elifesciences/issues/issues/4639).

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1610,7 +1610,7 @@ iiif:
                 timeout: 2000
             surrogate-keys:
                 article-id-versioned:
-                    url: "^/lax[:/]([0-9]+)(?:/|%2F)elife-(?:[a-z0-9-]+)-v([0-9]+)\\.(?:.+)$"
+                    url: "^/lax[:/]([0-9]+)(?:/|%252F)elife-(?:[a-z0-9-]+)-v([0-9]+)\\.(?:.+)$"
                     value: "article/\\1 article/\\1v\\2"
                     samples:
                         figure:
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -1623,10 +1623,10 @@ iiif:
                             path: /lax:18215/elife-18215-app2-fig1-figsupp2-v2.tif
                             expected: "article/18215 article/18215v2"
                         encoded-urls:
-                            path: /lax/18215%2Felife-18215-app2-fig1-figsupp2-v2.tif
+                            path: /lax/18215%252Felife-18215-app2-fig1-figsupp2-v2.tif
```
thewilkybarkid: Other option is to always percent-encoding `%` when generating VCL/Fastly config.

projects/elife.yaml
```diff
@@ -1639,25 +1639,20 @@ iiif:
...
+                            path: /lax/26866%252Felife-26866-fig3-video1.jpg
                             expected: "article/26866 article/26866/videos"
                 digest-id:
-                    # does work:
-                    url: "^/digests/([0-9]+).*$"
-                    # doesn't work:
-                    #url: "^/digests/([0-9]+)(?:/|%2F)(?:.+)$"
-                    # also doesn't work:
-                    #url: "^/digests/([0-9]+)(?:/|%2F)"
+                    url: "^/digests/([0-9]+)(?:/|%252F)(?:.+)$"
```
thewilkybarkid: Turns out we had encountered this problem before, just hadn't worked out what was happening.

projects/elife.yaml
giorgiosironi: It would look more readable and less error-prone if we can do it reliably

giorgiosironi: These are the place where to encode it:
https://github.com/elifesciences/builder/blob/master/src/buildercore/terraform.py#L307
https://github.com/elifesciences/builder/blob/master/src/buildercore/terraform.py#L314

giorgiosironi: For example, variables of the template too:
https://github.com/elifesciences/builder/blob/master/src/buildercore/terraform.py#L363
We'd need a `VCLExpression` object for that:
```
VCLExpression('req.url ~ "%s"').format(surrogate['url'])
```

thewilkybarkid:
Going to deploy as-is to fix the problems in production.

PR #438, opened 2018-12-18
[wip] Add support for private schemas for BigQuery tables

giorgiosironi:
For https://github.com/elifesciences/issues/issues/4467

### DIFF
--- a/src/buildercore/terraform.py
+++ b/src/buildercore/terraform.py
@@ -21,6 +21,7 @@ DATA_TYPE_TEMPLATE = 'template_file'
 DATA_NAME_VAULT_GCS_LOGGING = 'fastly-gcs-logging'
 DATA_NAME_VAULT_GCP_LOGGING = 'fastly-gcp-logging'
 DATA_NAME_VAULT_FASTLY_API_KEY = 'fastly'
+DATA_NAME_VAULT_GITHUB = 'github'
 
 # keys to lookup in Vault
 # cannot modify these without putting new values inside Vault:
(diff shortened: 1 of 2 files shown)


src/buildercore/terraform.py
```diff
@@ -555,6 +562,13 @@ def add_table(table_id, table_options):
             'schema': schema_ref,
         }
 
+        if github_token:
```
giorgiosironi: boolean flags are a mess, needs refactoring but this managed to create tables in the `staging` dataset

src/buildercore/terraform.py
```diff
@@ -530,13 +532,18 @@ def add_table(table_id, table_options):
...
         fqrn = "%s_%s" % (table_options['dataset_id'], table_id) # 'fully qualified resource name'
+        github_token = False
 
         if schema.startswith('https://'):
             # remote schema, add a 'http' provider and have terraform pull it down for us
             # https://www.terraform.io/docs/providers/http/data_source.html
             tf_file['data'][DATA_TYPE_HTTP][fqrn] = {'url': schema}
             schema_ref = '${data.http.%s.body}' % fqrn
-
+            if re.match('^https://raw\.githubusercontent\.com/', schema):
```
de-code: (might as well use `startswith`?)

de-code: Might also be missing the `r` prefix if this is to use a regular expression. Otherwise '\.' would ask it to (un)escape '.' as a string, rather than a regular expression.

giorgiosironi: It kinda works because there is no meaning for `\.` like there is for `\n`, but yes I'll make it robust

src/buildercore/terraform.py
```diff
@@ -532,18 +532,18 @@ def add_table(table_id, table_options):
         schema = table_options['schema']
         stackname = context['stackname']
         fqrn = "%s_%s" % (table_options['dataset_id'], table_id) # 'fully qualified resource name'
-        github_token = False
+        needs_github_token = False
```
de-code: much better name

PR #441, opened 2019-01-14
Add end2end--xpub--1.elifesciences.org

giorgiosironi:
A primary node is defined as node 1 of a cluster of N identical, which are usually behind a load balancer. Sometimes this node is special as it runs processes like a testing database or a unique process that should not be running on other nodes.

This adds support for `dns-external-primary`, which adds a public DNS for node 1 of a cluster. This makes the primary node accessible to the outside world, for example to `elife-spectrum` tests.

- Create with the `ExtDNS1` DNS resource in CloudFormation
- Maintain after reboots in `update_dns`

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -29,6 +29,9 @@ defaults:
         ec2:
             # how many EC2 instance per stack instance
             cluster-size: 1
+            # whether the first EC2 node should get an external DNS entry such as env--project--1.elifesciences.org
+            # only makes sense if cluster-size > 1
+            dns-external-primary: True
             # whether the EC2 nodes should get a per-node internal DNS entry such as env--project--1.elife.internal
             # only makes sense if cluster-size > 1
             dns-internal: False
@@ -1920,6 +1923,14 @@ elife-xpub:
             ec2:
                 cluster-size: 2
                 dns-internal: True
+            ports:
+                - 22
+                - 80
+                - 443
+                - 2222:
+                    cidr-ip: 10.0.0.0/16 # access via VPC ip range only
+                - 8020 # nginx-public-folders via HTTP
+                - 8021 # nginx-public-folders via HTTPS
             elb:
                 protocol:
                     - 443
@@ -1937,6 +1948,7 @@ elife-xpub:
                 encryption: arn:aws:kms:us-east-1:512686554592:key/b626157a-1e5b-4bf6-8799-211505efe072 
             ec2:
                 cluster-size: 2
+                dns-external-primary: True
                 dns-internal: True
             elb:
                 protocol:
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1312,7 +1312,7 @@ search:
                     cidr-ip: 10.0.0.0/16 # follower nodes
             ec2:
                 cluster-size: 2
-                dns-internal: True
+                dns-internal: true
             elb: 
                 protocol: 
                     - http
@@ -1327,7 +1327,7 @@ search:
                     cidr-ip: 10.0.0.0/16 # follower nodes
             ec2:
                 cluster-size: 2
-                dns-internal: True
+                dns-internal: true
             elb: 
                 protocol: 
                     - http
@@ -1922,7 +1922,7 @@ elife-xpub:
                 encryption: arn:aws:kms:us-east-1:512686554592:key/b626157a-1e5b-4bf6-8799-211505efe072 
             ec2:
                 cluster-size: 2
-                dns-internal: True
+                dns-internal: true
             ports:
                 - 22
                 - 80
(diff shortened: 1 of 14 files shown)


giorgiosironi:
Also thrown in @GiancarloFusiello in the deep end, but will provide explanations.

src/tests/fixtures/projects/dummy-project.yaml
```diff
@@ -454,6 +455,7 @@ project-with-cluster:
             - 80
         ec2:
             cluster-size: 2
+            dns-external-primary: True
```
GiancarloFusiello: I think `True` should be `true`

src/tests/fixtures/projects/dummy-project.yaml
```diff
@@ -454,6 +455,7 @@ project-with-cluster:
...
             dns-internal: True
```
GiancarloFusiello: `true`

src/tests/fixtures/projects/dummy-project2.yaml
```diff
@@ -22,6 +22,7 @@ defaults:
     aws:
         ec2:
             cluster-size: 1
+            dns-external-primary: false
             dns-internal: False
```
GiancarloFusiello: `false`

src/tests/fixtures/projects/dummy-project.yaml
lsh-0: I don't think it matters

src/buildercore/core.py
```diff
@@ -668,8 +668,8 @@ def hostname_struct(stackname):
         return struct
 
     # removes any non-alphanumeric or hyphen characters
-    subsubdomain = re.sub(r'[^\w\-]', '', instance_id)
-    hostname = subsubdomain + "--" + subdomain
+    instance_subdomain_fragment = re.sub(r'[^\w\-]', '', instance_id)
```
lsh-0: ha `subsubdomain` - I can tell that was me, once upon a time

src/buildercore/core.py
```diff
@@ -678,6 +678,7 @@ def hostname_struct(stackname):
     if domain:
         updates['project_hostname'] = subdomain + "." + domain
         updates['full_hostname'] = hostname + "." + domain
+        updates['ext_node_hostname'] = hostname + "--%s." + domain
```
lsh-0: we should have an example I think: `... # "prod--journal--1.elifesciences.org"`

src/buildercore/core.py
lsh-0: also, it's no longer a sub-sub-domain anyway, we got rid of that ages ago. huh. code archaeology

src/buildercore/lifecycle.py
```diff
@@ -278,6 +284,7 @@ def _wait_for_running_nodes(stackname):
     # We don't strictly need to do this, as the private ip address
     # inside a VPC should stay the same. For consistency we update all DNS
     # entries as the operation is idempotent
+    # TODO: possibly drop
```
lsh-0: so drop it. we need less noise in the logging

src/tests/fixtures/projects/dummy-project.yaml
giorgiosironi: Afaik YAML is very forgiving and will accept all sorts of casing for booleans, but it would be useful to have a single all-lowercase convention so that `grep`, find/replace and other tools work consistently.

giorgiosironi:
`lifecycle` is a module specialized in starting/stopping servers, hence it's not a primary source of documentation. I'll find a place in `docs/`.

src/buildercore/core.py
giorgiosironi: Have documented in `docs/dns.md`, I'd leave this as internal unless we document the `context` format

projects/elife.yaml
```diff
@@ -1926,6 +1929,8 @@ elife-xpub:
                 - 443
                 - 2222:
                     cidr-ip: 10.0.0.0/16 # access via VPC ip range only
+                - 8020 # nginx-public-folders via HTTP
+                - 8021 # nginx-public-folders via HTTPS
```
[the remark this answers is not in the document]
giorgiosironi: A way to expose some files from the `end2end--xpub--1.elifesciences.org` node via an nginx web server:
https://github.com/elifesciences/builder-base-formula/blob/d98dd72cd86fe2e334ac7c3844470bda13d7362c/elife/config/etc-nginx-sites-enabled-public-folders.conf

src/tests/fixtures/projects/dummy-project.yaml
lsh-0: > grep, find/replace and other tools work consistently.

that's kind of a weak reason ;) 

"I don't think it matters" == "I don't really care" , it's the sort of minutia teams get fixated on when there aren't more pressing problems. Perhaps what we actually need is an elife 3.0?

PR #445, opened 2019-01-29
Uniform TerraformTemplate usage

giorgiosironi:
Follow-up https://github.com/elifesciences/builder/pull/420 which removed the need for continuous null checks, but wasn't used in all `render_*()` functions. Follows the `troposphere` model of a `template` output parameter, which is stateful but makes it very easy to grow the `.tf` file without accidentally overwriting conflicting keys coming from different providers like `data.*` ones.

### DIFF
--- a/src/buildercore/terraform.py
+++ b/src/buildercore/terraform.py
@@ -490,22 +490,22 @@ def render_gcs(context):
     if not context['gcs']:
         return {}
 
-    return {
-        'resource': {
-            'google_storage_bucket': {
-                bucket_name: {
-                    'name': bucket_name,
-                    'location': 'us-east4',
-                    'storage_class': 'REGIONAL',
-                    'project': options['project'],
-                } for bucket_name, options in context['gcs'].items()
-            },
-        },
-    }
+    template = TerraformTemplate()
+    for bucket_name, options in context['gcs'].items():
+        template.populate_resource('google_storage_bucket', bucket_name, block={
+            'name': bucket_name,
+            'location': 'us-east4',
+            'storage_class': 'REGIONAL',
+            'project': options['project'],
+        })
+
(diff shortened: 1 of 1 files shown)


src/buildercore/terraform.py
```diff
@@ -91,25 +90,22 @@
...
         render_fastly,
         render_gcs,
         render_bigquery
     ]
-    partial_tf_list = [fn(context) for fn in fn_list]
-
-    def merge(a, b):
-        deepmerge(a, b)
-        return a
+    [fn(context, template) for fn in fn_list]
```
de-code: (minor) I'd probably change that to a regular loop because we don't use the array (and that may be a bit confusing):

```python
for fn in fn_list:
    fn(context, template)
```

src/buildercore/terraform.py
```diff
@@ -486,24 +481,21 @@ def _generate_vcl_file(stackname, content, key, extension='vcl'):
...
-                    'name': bucket_name,
-                    'location': 'us-east4',
-                    'storage_class': 'REGIONAL',
-                    'project': options['project'],
-                } for bucket_name, options in context['gcs'].items()
-            },
-        },
-    }
+    for bucket_name, options in context['gcs'].items():
+        template.populate_resource('google_storage_bucket', bucket_name, block={
```
de-code: (very minor) maybe `block` is not the best name. But it was defined outside this PR.

giorgiosironi: it's terminology coming from Terraform itself, although not "official" https://www.terraform.io/docs/configuration/terraform.html#description

src/buildercore/terraform.py
giorgiosironi: Yes, no need for side-effects in list comprehensions

PR #447, opened 2019-01-31
Switch elife-xpub ELB healthcheck to dedicated path

giorgiosironi:
For https://github.com/elifesciences/elife-xpub/issues/1409

Still to be applied as it depends on https://github.com/elifesciences/elife-xpub-formula/pull/70

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1934,6 +1934,10 @@ elife-xpub:
             elb:
                 protocol:
                     - 443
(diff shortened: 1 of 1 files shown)


giorgiosironi:
No, it is only nginx responding without caring about the application. Which I don't like it very much because if a single node has problems it will start responding with nginx errors but not be detached from the load balancer. So perhaps we should shelve this.

PR #451, opened 2019-02-06
Vault integration for projects

giorgiosironi:
Part of https://github.com/elifesciences/issues/issues/4406

## Policies
- Documents a `master-server` Vault policy that is used to generate a non-root token for the Salt master
- Documents a `elife-xpub` Vault policy that is used by `elife-xpub` Salt minions
- Updates the policies with the syntax required by the KV store version 2, which has an additional `data` prefix
- Upgrade minimum requires Vault to avoid policy commands failing
- Adds support for grains and a `project` grain that can be used to choose the Vault policy for the Salt minion

## Masterless

- Generates a token with limited permissions for the master instance
- Connects the masterless instance to Vault during its bootstrap so that it can read secrets


giorgiosironi:
Can't reproduce the `shellcheck` problem locally, checking versions.

giorgiosironi:
0.3.7 in both cases.

giorgiosironi:
Ah, it does a whole script `.ci/shell-checking`, now can reproduce.

giorgiosironi:
The build checks that the scripts can run in [4 different shells](https://github.com/elifesciences/builder/blob/master/.shell-lint.sh#L18-L21) but the scripts themselves specify they should always be run with `bash`:
```
scripts/init-vagrant-formulas.sh:#!/bin/bash
scripts/customize.sh:#!/bin/bash
scripts/init-masterless-formulas.sh:#!/usr/bin/bash
scripts/mount-shared-folders.sh:#!/bin/bash
scripts/bootstrap.sh:#!/bin/bash
scripts/update-master.sh:#!/bin/bash
scripts/init-master.sh:#!/bin/bash
scripts/clean-stack-for-ami.sh:#!/bin/bash
scripts/highstate.sh:#!/bin/bash
scripts/test.sh:#!/bin/bash
scripts/update-masterless-formula.sh:#!/bin/bash
scripts/new-project.sh:#!/bin/bash
```
I'm not sure if someone symlinks `bash` to `ksh` or similar, but these scripts should not be run locally, they are meant for a Vagrant or EC2 environment where we control the base image?

docs/vault.md
```diff
@@ -2,9 +2,11 @@
 
 Vault is a tool for managing secrets, such as API keys and other credentials. In the context of builder, these credentials are mostly necessary to access or modify infrastructure.
 
-## User scenario
+## Command-line interface
 
-builder assumes a logged-in Vault client, probably running on the master server.
+### User scenario
+
+To perform `updte_infrastructure`, builder assumes a logged-in Vault client, probably running on the master server.
```
lsh-0: typo

lsh-0:
> I'm not sure if someone symlinks `bash` to `ksh` or similar, but these scripts should not be run locally, they are meant for a Vagrant or EC2 environment where we control the base image?

I know, think of shellcheck as an optional good practice check. It keeps us from using esoteric language features that we may not realize are such and does a decent job in linting.

giorgiosironi:
> shellcheck as an optional good practice check

I think it's very useful to statically check scripts that would take minutes to be run. I only reduced the breadth of shell support as the scripts specify to only use `bash` in the hashbang line.

giorgiosironi:
These changes have been tested and do not affect formulas, though most of them are documentation. I'm going to merge this to simplify testing of https://github.com/elifesciences/containers-formula/pull/5 which requires to create instances using this branch.

PR #452, opened 2019-02-07
Simplify project configuration options

giorgiosironi:
For https://github.com/elifesciences/issues/issues/4712

Directed to a branch because it builds on top of that, to avoid conflicts.

So far, the scope of this is removing various `settings.yml` dependencies, but still allows multiple project files. We can get to a situation where we only have one project file at a time too, but that's a larger change.


src/buildercore/keypair.py
```diff
@@ -17,22 +16,21 @@
...
 
-@if_enabled('write-keypairs-to-s3', silent=True)
 def write_keypair_to_s3(stackname):
     # this is the path to where .save() puts it
     # http://boto.readthedocs.io/en/latest/ref/ec2.html#boto.ec2.keypair.KeyPair
     path = stack_pem(stackname, die_if_doesnt_exist=True)
     key = s3_keypair_key(stackname)
-    s3.write(key, open(path, 'r'))
+    with open(path, 'r') as fp:
+        pem_contents = fp.read()
```
lsh-0: a simple `pem_contents = open(path, 'r').read()` is all that is needed. the file handle is closed immediately

src/integration_tests/test_validation.py
```diff
@@ -21,8 +21,6 @@ def test_validation(self, project_name):
 class TestValidationElife():
     @classmethod
     def setup_class(cls):
-        # HERE BE DRAGONS
-        # resets the testing config.SETTINGS_FILE we set in the base.BaseCase class
```
lsh-0: eh, isn't there something still dodgy here? or was that fixed?

src/buildercore/keypair.py
giorgiosironi: it doesn't seem closed after these two calls?
```
>>> fp = open('projects/elife.yaml')
>>> len(fp.read())
77047
>>> fp.seek(0)
>>> len(fp.read())
77047
```

src/integration_tests/test_validation.py
giorgiosironi: it is now changing `config.PROJECT_FILES`, but it seems to have become an established pattern, so lest of introducing dependency injection it's here to stay

src/buildercore/keypair.py
lsh-0: well you're keeping a reference to it with `fp` ;)

After some googling, what I recall is that the CPython garbage collector closes file handles with no references to it immediately. Given that explicit is better than implicit and all that and we shouldn't rely on compiler/interpreter/garbage collector behaviour, I don't really mind.

PR #453, opened 2019-02-08
Turn up user filter to cover 30% of traffic

giorgiosironi:
Also upgrade Terraform Vault provider to allow reading secrets from the versioned key value store (V2):

https://github.com/terraform-providers/terraform-provider-vault/issues/140


giorgiosironi:
Just applied to `prod`, the only affected environment.

giorgiosironi:
(upgrading Vault was necessary to get a bug fix needed to apply any change to `journal--prod`)

PR #455, opened 2019-02-12
Fix conflict on multiple tables each settings its own github token

giorgiosironi:


giorgiosironi:
Bugfix applied to `data-pipeline--*` stacks, test suite is green.

PR #459, opened 2019-02-18
added 795_datascience_person_merge_info_all schema

de-code:
that schema was missing

/cc @lsh-0

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -2058,6 +2058,8 @@ data-pipeline:
                         schema: https://raw.githubusercontent.com/elifesciences/data-pipeline-ejp-csv-deposit/master/schemas/767_datascience_person_roles.json
                     779_datascience_reviewer_info_all:
                         schema: https://raw.githubusercontent.com/elifesciences/data-pipeline-ejp-csv-deposit/master/schemas/779_datascience_reviewer_info_all.json
+                    795_datascience_person_merge_info_all:
+                        schema: https://raw.githubusercontent.com/elifesciences/data-pipeline-ejp-csv-deposit/master/schemas/795_datascience_person_merge_info_all.json
                     manuscript:
                         schema: https://raw.githubusercontent.com/elifesciences/data-pipeline-ejp-to-json-converter/develop/schemas/manuscript.bqschema.json
                     manuscript_full:
--- a/src/buildercore/terraform.py
+++ b/src/buildercore/terraform.py
@@ -703,7 +703,7 @@ def init(stackname, context):
                     'api_key': "${data.%s.%s.data[\"api_key\"]}" % (DATA_TYPE_VAULT_GENERIC_SECRET, DATA_NAME_VAULT_FASTLY_API_KEY),
                 },
                 'google': {
-                    'version': "= %s" % '1.13.0',
+                    'version': "= %s" % '1.20.0',
                     'region': 'us-east4',
                     # TODO: the system-wide authentication is being used
                     # to provision through this provider (see `gcloud auth list`)


de-code:
> It then fails applying in one of them:
> 
> ```
> 1 error(s) occurred:
> 
> * google_bigquery_table.ci_person_v2: 1 error(s) occurred:
> 
> * google_bigquery_table.ci_person_v2: googleapi: Error 400: Provided Schema does not match Table elife-data-pipeline:ci.person_v2. Field merged_into_person_ids has change$ type from INTEGER to STRING, invalid
> ```

Let's just delete `person_v2` (and let it re-create the file). No need to try to migrate the schema. Should I do that manually?

giorgiosironi:
Will do that through `builder`, since it's tracked in Terraform's state.

de-code:
In general the `manuscript*` and `person*` tables (coming from the XML) are being replaced rather than appended. Making it unnecessary to migrate a schema in general. It can be replaced. Apart from maybe `prod`, where it would be better to let the pipeline replace the table (rather than replacing it with an empty one)

For the other schemas, relating to appended tables, migration of the schema would make sense in `prod`. And as a test in `staging` (although currently without data).

giorgiosironi:
Removed and readded to table to `ci`; can do the same to `staging`; what should I do to `prod`?
If you migrate it manually then it can be readded to Terraform with the `import` command but it's not a sustainable process.

de-code:
> Removed and readded to table to `ci`

Thank you.

> can do the same to `staging`

Yes, please.

> what should I do to `prod`?

The corresponding code had already been deployed to `prod` and the table is already using the new schema. Do we need to do anything or will Terraform still complain?

giorgiosironi:
It will complain that the change is already there, but I'll post what I do to fix it.

giorgiosironi:
Got to `staging`, but like in `ci` after repeating the application Terraform still finds 4 tables to update. Shortest example:
```
~ google_bigquery_table.staging_manuscript
      schema: "[{\"mode\":\"NULLABLE\",\"name\":\"country\",\"type\":\"STRING\"},{\"mode\":\"NULLABLE\",\"name\":\"doi\",\"type\":\"STRING\"},{\"mode\":\"NULLABLE\",\"name
\":\"long_manuscript_identifier\",\"type\":\"STRING\"},{\"mode\":\"NULLABLE\",\"name\":\"manuscript_id\",\"type\":\"STRING\"},{\"mode\":\"NULLABLE\",\"name\":\"modified_ti
mestamp\",\"type\":\"TIMESTAMP\"},{\"fields\":[{\"mode\":\"NULLABLE\",\"name\":\"source_filename\",\"type\":\"STRING\"},{\"mode\":\"NULLABLE\",\"name\":\"imported_timestam
p\",\"type\":\"TIMESTAMP\"}],\"mode\":\"NULLABLE\",\"name\":\"provenance\",\"type\":\"RECORD\"}]" => "[{\"mode\":\"NULLABLE\",\"name\":\"country\",\"type\":\"STRING\"},{\"
mode\":\"NULLABLE\",\"name\":\"doi\",\"type\":\"STRING\"},{\"mode\":\"NULLABLE\",\"name\":\"long_manuscript_identifier\",\"type\":\"STRING\"},{\"mode\":\"NULLABLE\",\"name
\":\"manuscript_id\",\"type\":\"STRING\"},{\"mode\":\"NULLABLE\",\"name\":\"modified_timestamp\",\"type\":\"TIMESTAMP\"},{\"fields\":[{\"mode\":\"NULLABLE\",\"name\":\"imp
orted_timestamp\",\"type\":\"TIMESTAMP\"},{\"mode\":\"NULLABLE\",\"name\":\"source_filename\",\"type\":\"STRING\"}],\"mode\":\"NULLABLE\",\"name\":\"provenance\",\"type\":
\"RECORD\"}]"
```
There must be some cosmetic difference between the schema as stored in the repository and the schema that BigQuery shows from the API. How do we dissect this?

de-code:
> There must be some cosmetic difference between the schema as stored in the repository and the schema that BigQuery shows from the API. How do we dissect this?

Diff shows me that the order of `imported_timestamp` and `source_filename` has changed (within `provenance`). The order should be irrelevant.

giorgiosironi:
On `prod` I had to:
```
$ terraform import google_bigquery_table.prod_795_datascience_person_merge_info_all elife-data-
pipeline:prod.795_datascience_person_merge_info_all                          
```
but after that Terraform does not detect any more change to make (so it works correctly).

giorgiosironi:
Going to merge as-is to unblock the other PRs.

PR #465, opened 2019-02-27
Turn up the user dial to 100%

diversemix:
This will mean ~64% of total submissions will be coming via xPub

PR #466, opened 2019-02-27
downgrade pip and reduce docker image size for python 3

GiancarloFusiello:
This should resolve the issues relating to the errors [here](https://github.com/elifesciences/builder/pull/465)

### DIFF
--- a/Dockerfile.py3
+++ b/Dockerfile.py3
@@ -1,7 +1,23 @@
-FROM python:3.5-slim-jessie
+FROM python:3.5-alpine3.8
 
-RUN pip install virtualenv
-RUN mkdir /venv && virtualenv --python=python3 /venv
+RUN apk add --no-cache \
+    libressl-dev \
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Context: the Docker images are an attempt to simplify running `builder` for various Mac users but they are not actively being published or run. There is also a Python 2 Docker image.

GiancarloFusiello:
> Context: the Docker images are an attempt to simplify running `builder` for various Mac users but they are not actively being published or run. There is also a Python 2 Docker image.

@giorgiosironi thanks for the context. I'll go ahead and push the big green button.

giorgiosironi:
A separate problem is that we mostly use Python 2 because on Python 3 Fabric/Paramiko has performance problems with large SSH outputs. All this to say changes to the Python 3 Docker image are low risk as there isn't much depending on it apart from this build.

PR #468, opened 2019-03-04
Upgrade elife-libraries to 16.04

giorgiosironi:
For https://github.com/elifesciences/issues/issues/4780

- Support suppression of EC2 without deleting its volume
- Port elife-libraries--powerful* to 16.04
- Add guide to upgrading a node without destroying the stack


giorgiosironi:
https://alfred.elifesciences.org/job/libraries/job/library-bot-lax-adaptor/455/console is the first build running on one of the new stacks (`elife-libraries--powerful3`).

src/buildercore/trop.py
```diff
@@ -305,27 +305,34 @@ def render_rds(context, template):
...
     lmap(template.add_output, outputs)
 
-def render_ext_volume(context, context_ext, template, node=1):
+def render_ext_volume(context, context_ext, template, actual_ec2_instances, node=1):
     vtype = context_ext.get('type', 'standard')
+    # TODO: remove if not in use
     if vtype == 'ssd':
-        LOG.warn("deprecated value of 'ssd' used in aws.ext.type: %s", context['stackname'])
-        vtype = 'gp2'
+        raise RuntimeError("deprecated value of 'ssd' used in aws.ext.type: %s", context['stackname'])
```
lsh-0: if you're now failing on it, then it's obsoleted rather than deprecated.

a better message might be "Obsolete value of 'ssd' used in 'aws.ext.type' on stack '%s'. This should be a value of 'gp2'."

giorgiosironi:
Recreating `ci` and `load`.

src/buildercore/trop.py
giorgiosironi: I found out this is not in use anymore in any file, so time to remove the check

giorgiosironi:
Recreating `spectrum`.

src/buildercore/trop.py
lsh-0: good call

PR #469, opened 2019-03-06
Feature: shifts default ec2 ami and vagrant box to Ubuntu 18.04

lsh-0:
ticket: https://github.com/elifesciences/issues/issues/4793

* changed default ec2 ami to 18.04 elife basebox AMI
* changed default 'fresh' aws-alt to 18.04 public AMI
* changed default 'standalone' aws-alt to inherit 18.04 basebox AMI
* changed default vagrant box to 18.04
* added '1804' aws-alt with inherited 18.04 basebox AMI
* added 's1804' (standalone1804) aws-alt with inherited 18.04 basebox AMI
* updates all projects fixing their version of ubuntu

minor:
* removed box-url in vagrantfile (ancient setting, now unused)
* changed 'unreasonable' vagrant ram threshold to 4096
* changed default vagrant ram to 2048
* fixed description of 'fresh' aws-alt
* added 'deprecated' flags against mentions of 14.04 and 16.04 usage


lsh-0:
doing some `update_infrastructure` tests of affected projects now

giorgiosironi:
https://github.com/elifesciences/builder/blob/master/Jenkinsfile.update-ami is supposed to open pull requests when we want to regenerate the 18.04 `basebox` AMI. Needs to be checked to see how it will continue to work.

lsh-0:
I just found this: https://github.com/elifesciences/builder/blob/master/update-ami.sh

I didn't realise you were relying on the precise structure of those comments.

It should also probably live in ./scripts

I'll try it out today

giorgiosironi:
It relies on the single line structure so moving things up and down should be safe

> live in ./scripts

I haven't mixed it with the scripts there that run on servers, because its purpose is to update this repository instead; but it could use its own subfolder.

projects/elife.yaml
```diff
@@ -179,67 +177,76 @@ defaults:
...
-            description: isolated from the master-server and uses Ubuntu 18.04 (Bionic)
-            ec2:
-                ami: ami-00aedeb287fe8c9d6 # GENERATED created from basebox--1804
-                                           # us-east-1, build date 20180814, hvm:ebs-ssd, AMI built on 20180828
-                masterless: true
-        s1804:
-            description: alias for standalone1604 short enough to fit into hostnames
+            description: isolated from the master-server and uses same AMI as default configuration
             ec2:
-                ami: ami-00aedeb287fe8c9d6
```
giorgiosironi: went through all of the `ami` fields at https://console.aws.amazon.com/ec2/v2/home?region=us-east-1#Images:visibility=owned-by-me;search=92002785;sort=imageId and everything checks out

projects/elife.yaml
```diff
@@ -341,13 +370,6 @@ lax:
             bot-lax-{instance}-inc: {}
             bot-lax-{instance}-out: {}
     aws-alt:
-        standalone:
```
giorgiosironi: if you remove this, all `lax-formula` builds will start to build on 18.04?

giorgiosironi: Same goes for the other `standalone` configurations: they should be pinned to 14.04 now (with the AMI) and that should be removed for each of them when the formula is ready for 18.04 and the instances are recreated.

projects/elife.yaml
```diff
@@ -1179,6 +1159,9 @@ elife-alfred-nodes:
     domain: false
     intdomain: false
     aws:
+        # 16.04, deprecated
+        ami: ami-14a1d06e # GENERATED created from basebox--1604
```
giorgiosironi: of course, projects that are already on 16.04 should stay pinned to that rather than 14.04, waiting for their 18.04 upgrade

projects/elife.yaml
lsh-0: no, you're mistaken. The default `standalone` configuration doesn't specify an ami, just that it should be masterless. The ami is pulled from the project configuration, which is overridden (pinned) to use a deprecated 14.04/16.04. 

you can test this yourself with:

```
source ./venv/bin/activate
./.project.py lax
```

giorgiosironi: Good to go then

lsh-0:
I'll merge this in after I've updated the machines to use the new salt version.

projects/elife.yaml
giorgiosironi: Some pins were badly specified and `elife-alfred-formula` started building on 18.04: https://github.com/elifesciences/builder/pull/488

giorgiosironi:
`standalone1804` got removed but there are builds like `builder-base-formula` depending on it (rather than `s1804`).

projects/elife.yaml
lsh-0: nuts, I see what you're saying, thanks for the fix.

PR #470, opened 2019-03-06
Feature: bumps the patch version of salt from 2017.7.5 to 2017.7.8

lsh-0:
* upgraded salt provides support for 18.04 LTS
* master-server will need to be updated first

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1,6 +1,6 @@
 defaults:
     description: defaults for all projects in this file
-    salt: '2017.7.5' # the version of salt these project use
+    salt: '2017.7.8' # the version of salt these project use
     # use false with a subdomain to assign internal addresses only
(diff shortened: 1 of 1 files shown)


lsh-0:
taking a look at the patch notes while these 18.04 vagrant instances build ...

giorgiosironi:
> master-server will need to be updated first

Even for a patch version change?

lsh-0:
> Even for a patch version change?

I'm not sure. The docs (https://docs.saltstack.com/en/2015.8/topics/installation/index.html#upgrading-salt) say the master should be updated first but doesn't distinguish between major/minor/patch versions.

lsh-0:
I'll do this Monday (your Sunday). I don't recall any catastrophic failures with Salt ever but Friday isn't a day to discover them.

PR #471, opened 2019-03-06
Feature: upgrades pattern-library to use Ubuntu 18.04

lsh-0:

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -662,12 +662,11 @@ journal:
             1240: 80
 
 pattern-library:
+    salt: '2017.7.8' 
     subdomain: ui-patterns
     repo: https://github.com/elifesciences/pattern-library
     formula-repo: https://github.com/elifesciences/pattern-library-formula
     aws:
-        ec2:
-            ami: ami-92002785 # Ubuntu 14.04, deprecated
         ports:
             - 22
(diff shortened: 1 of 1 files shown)


lsh-0:
hrm - `update_infrastructure` for `--ci` and `--prod` say nothing needs updating. Do I need to do something special here to get the ec2 instance replaced? or is it easier to just delete and recreate it?

giorgiosironi:
https://github.com/elifesciences/builder/blob/master/docs/playbooks/upgrade-os-instances.md is a procedure that allows to destroy and recreate the EC2 instance without destroying a stack. In the case of `pattern-library` it doesn't matter much because it's stateless, but it needs to be well rehearsed to be used on `journal-cms`. Destroying the whole stack there would also delete the RDS database and the EBS volume with the uploaded files, prolonging the downtime as they are restored from backups.

giorgiosironi:
`update_infrastructure` by default will ignore changes in the AMI (and a few other EC2 fields) because that will destroy all EC2 instances in our account whenever we update the AMI in `defaults`.

lsh-0:
>  because that will destroy all EC2 instances in our account whenever we update the AMI in defaults.

ah, that makes sense, I felt there had to be a good reason

PR #472, opened 2019-03-06
Feature: update journal-cms configuration to use ubuntu 18.04

lsh-0:
* shifts config to 16.04
* adds a default /ext volume


lsh-0:
(poor builder pipeline, I've sent a lot of changes through you today. I wonder if we should separate project configuration from builder code)

lsh-0:
Just tried to restore a backup but the file was too large to both exist and unpack at the same time.

projects/elife.yaml
```diff
@@ -926,13 +926,17 @@ journal-cms:
...
         ec2:
-            ami: ami-92002785 # Ubuntu 14.04, deprecated
+            ami: ami-14a1d06e # GENERATED created from basebox--1604
+                              # us-east-1, build date 20170811, hvm:ebs-ssd, AMI built on 20171215
         ports:
             - 22
             - 443
             - 80:
                 cidr-ip: 10.0.0.0/16 # access via VPC ip range only
+        ext:
```
giorgiosironi: this will also affect formula builds, but it's probably for the best as they may run out of space otherwise. The additional times it takes to provision shouldn't be a big deal, it's RDS that takes tens of minutes.

giorgiosironi:
> Just tried to restore a backup

Just an heads up that there are two sources of state: the database and the `sites/` folder containing uploaded files. To replace the servers, RDS should cover the first one but not the second. Not sure which backups you are restoring now.

projects/elife.yaml
lsh-0: I agree. I'll revert those `ext` changes. `journal-cms` is just so fat every case is a special case. I will add a special alt-config intended for adhoc instances though with rds and the large ext disk.

giorgiosironi:
Title is wrong, correcting.

PR #473, opened 2019-03-11
Feature: upgrade elife metrics config to 18.04

lsh-0:
elife-metrics now works on 16.04 and 18.04

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -692,8 +692,6 @@ elife-metrics:
     repo: https://github.com/elifesciences/elife-metrics
     formula-repo: https://github.com/elifesciences/elife-metrics-formula
     aws:
-        ec2:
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Should we actually replace the existing servers to get this in? New servers will be identical to existing servers in that way.

lsh-0:
weren't we going to use your new ec2 replacement logic in builder?

I need to merge in the upgraded Salt version and upgraded 18.04 defaults first.

giorgiosironi:
> weren't we going to use your new ec2 replacement logic in builder?

Yes, I am suggesting (when prerequisites PRs are merged) to apply this to existing stacks with the replacement procedure in `docs/`, and only after that has gone through all the environments merge. I routinely do this with `update_infrastructure` changes (apply before merge) as I can't be surprised with `master` having a change that breaks one of the stacks.

PR #480, opened 2019-03-18
Feature: shift observer to 1804

lsh-0:


giorgiosironi:
Blocked on https://github.com/elifesciences/builder/pull/469 which contains these commits?

giorgiosironi:
Should just go in that branch?

giorgiosironi:
But we want to merge that branch first without having to replace various instances of various projects to reflect the new OS, so hold. Same for similar PRs like https://github.com/elifesciences/builder/pull/479/files which should be merged at the same time as the servers they are replacing.

Everything seems to be blocked on https://github.com/elifesciences/issues/issues/4826

PR #484, opened 2019-03-19
Use Vault to authenticate to GCP

giorgiosironi:
For https://github.com/elifesciences/issues/issues/4833

Run successfully:
```
bldr update_infrastructure:fastly-logs--end2end,skip=ec2
bldr update_infrastructure:data-pipeline--ci,skip=ec2
```
after having revoked my local `gcloud` authentication.

The providers look like:
```
$ cat .cfn/terraform/fastly-logs--end2end/providers.tf.json | jq .provider
{
  "fastly": {
    "api_key": "${data.vault_generic_secret.fastly.data[\"api_key\"]}",
    "version": "= 0.4.0"
  },
  "google": {
    "credentials": "${data.vault_generic_secret.gcp.data[\"credentials\"]}",
    "region": "us-east4",
    "version": "= 1.20.0"
  },
  "vault": {
    "version": "= 1.3",
    "address": "https://master-server.elifesciences.org:8200"
  }
}
```
so this is now using the same pattern as Fastly.

### DIFF
--- a/docs/vault.md
+++ b/docs/vault.md
@@ -72,3 +72,16 @@ Non-`master-server` stacks can pull secrets from Vault through the Salt master, 
 | Vagrant         | Masterless?        | No             | -             |
 | EC2             | Masterful          | Only for tests | Itself        |
 | EC2             | Masterless         | Only for tests | Itself        |
+
+## Builder secrets dependencies
+
+```
+VAULT_ADDR=https://master-server.elifesciences.org:8200 vault kv list secret/builder/apikey
+```
+gives the following:
+
+- `fastly`: API key for Fastly `it-admin@elifesciences.org` user, with permission to modify the CDN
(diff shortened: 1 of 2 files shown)


lsh-0:
the bigquery changes in datapipeline--ci don't appear to be idempotent. If you updated them, and I just updated them and it's prompting me to update them again now ... also, `skip=bigquery` didn't seem to work either. I don't think these are blockers to this PR

giorgiosironi:
> the bigquery changes in datapipeline--ci don't appear to be idempotent.

Yes, there is an unresolved ordering problem where you apply a schema and the API returns the fields in a different order, Terraform not detecting they are equivalent.

giorgiosironi:
> `skip=bigquery` didn't seem to work

`skip` is implemented for post-CloudFormation and post-Terraform actions (mostly `ec2` but also `sqs` and `s3`); doesn't cover anything else as the two tools need to be applied wholesale.

PR #488, opened 2019-03-20
Fix pinning to 16.04 of several stacks

giorgiosironi:


giorgiosironi:
Re-running because of an outdated Terraform on Alfred.

PR #489, opened 2019-03-22
Feature: improves Salt documentation and upgrade tasks

lsh-0:
* removed some dead code from `master.py`
* split the salt-update code from the remaster-minion code in `master.py`
* updated the salt docs

### DIFF
--- a/docs/salt.md
+++ b/docs/salt.md
@@ -8,7 +8,8 @@ The process for installing Salt and configuring it differs slightly depending on
 
 A project called [salt-bootstrap](https://github.com/saltstack/salt-bootstrap) is used to install Salt on both Vagrant and AWS EC2 minions. The builder script that does this is found here: `./scripts/bootstrap.sh`.
 
-For AWS EC2, the builder commands `update` and `launch` upload and execute this script. If Salt is not detected or an older version is detected, `bootstrap.sh` will call the `salt-bootstrap` script with the correct parameters.
+For AWS EC2, the builder commands `update` and `update_infrastructure` will upload and execute this script. 
+If Salt is not detected then `bootstrap.sh` will call the `salt-bootstrap` script with the correct parameters. If an older version of Salt is detected, then only calls using `update_infrasture` (not `update`) will upgrade the version of Salt.
 
 For Vagrant, the `vagrant up` and `vagrant provision` commands will run this script.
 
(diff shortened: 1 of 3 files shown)


docs/salt.md
```diff
@@ -8,29 +8,37 @@ The process for installing Salt and configuring it differs slightly depending on
 
 A project called [salt-bootstrap](https://github.com/saltstack/salt-bootstrap) is used to install Salt on both Vagrant and AWS EC2 minions. The builder script that does this is found here: `./scripts/bootstrap.sh`.
 
-For AWS EC2, the builder commands `update` and `launch` upload and execute this script. If Salt is not detected or an older version is detected, `bootstrap.sh` will call the `salt-bootstrap` script with the correct parameters.
+For AWS EC2, the builder commands `update` and `update_infrastructure` will upload and execute this script. 
+If Salt is not detected then `bootstrap.sh` will call the `salt-bootstrap` script with the correct parameters. If an older version of Salt is detected, then only calls using `update_infrasture` (not `update`) will upgrade the version of Salt.
```
giorgiosironi: `update_infrasture` is a typo

docs/upgrading-salt.md
```diff
@@ -0,0 +1,75 @@
...
+
+    $ ./bldr master.update_salt:master-server--2018-04-09-2
+
+## 7. update minions
+
+Minions are updated the same way as the master server, except you probably have a ~100 cases to handle:
+
+    $ ./bldr master.update_salt:project--instance
+
+At the time of writing there is no automated solution to updating the installed version of Salt.
```
giorgiosironi: as in "automated for all 100+ servers in one shot"?

src/master.py
```diff
@@ -141,12 +103,109 @@ def work():
...
+
+    return True
+
+# this is untested and probably unsafe/buggy
+#@debugtask
+def update_salt_everywhere():
+
+    # lsh, 2019-03-22: below code is copied from remaster_all, which *has* been used before
+
+    assert False, "this is meant for our future selves as an example in bulk updating the salt version on all minions"
```
giorgiosironi: I'll be ignoring anything after this line due to this comment, perhaps comment out if not used at the moment?

src/master.py
```diff
@@ -141,12 +103,109 @@ def work():
...
 
-# TODO: extract just the salt update part from `remaster`
 @debugtask
 @requires_aws_stack
 def update_salt(stackname):
-    current_master_stack = core.find_master_for_stack(stackname)
-    return remaster(stackname, current_master_stack)
+    "updates the installed version of salt to the version specified in the project file. see `docs/salt.md`"
+    # TODO: turn this into a decorator
+    import cfn
```
giorgiosironi: `cfn` is a module with `@task`s, it shouldn't be imported from other modules because you get duplicated tasks:
```
    masterless.launch
    masterless.set_versions               updates the cloned formulas on a masterl...
    masterless.cfn.aws_stack_list         returns a list of realized stacks. does ...
    masterless.cfn.cmd
    masterless.cfn.destroy                Delete a stack of resources.
    ...
```
`_check_want_to_be_running` should be moved to `buildercore.lifecycle` but there is a `utils._pick` calls to address. I'll make a separate PR to simplify it so that you can just `import buildercore.lifecycle` here.

src/master.py
```diff
@@ -141,12 +103,109 @@ def work():
...
+    # 2. update the stack's current context and write that back to s3
+    # 3. update the subset of the context on the ec2 instance itself (buildvars)
+    # 4. run highstate
+    LOG.info("updating salt version in context file")
+    pdata = core.project_data_for_stackname(stackname)
+    context['project']['salt'] = pdata['salt']
+    context_handler.write_context(stackname, context)
+
+    LOG.info("updating buildvars on ec2 instance")
+    buildvars.refresh(stackname, context)
```
giorgiosironi: `buildvars` are essentially a full copy of the `context`; if there is an update of infrastructure pending, we are propagating the new `context` to the instance without necessarily having created/updated/destroyed that infrastructure. Since [the formulas extensively refer to `buildvars`](https://github.com/elifesciences/builder-base-formula/search?q=elife.cfg&unscoped_q=elife.cfg) to decide what to do, I see this optimization as unsafe.

As an alternative we could remove the Salt version from the `context` to make it more dynamic, but that isn't right either because we wanted a controlled upgrade.

Last alternative is to use `update_infrastructure` and make sure we avoid drift. Should we calculate drifts in Alfred and publish some kind of warning when there are some?

docs/salt.md
lsh-0: sharp eye

docs/upgrading-salt.md
lsh-0: yup

src/master.py
lsh-0: this is good feedback but unfortunately part of this changeset made it into master. 

I've opened a [branch](https://github.com/elifesciences/builder/pull/665) to deal with this for the next upgrade.

PR #493, opened 2019-03-29
Install Helm on EKS clusters

giorgiosironi:


src/buildercore/terraform.py
```diff
@@ -816,6 +818,46 @@ def _render_eks_workers(context, template):
...
+            ],
+        })
+
+        # TODO: extract constants
+        template.populate_data(DATA_TYPE_HELM_REPOSITORY, DATA_NAME_HELM_INCUBATOR, block={
+            'name': 'incubator',
+            'url': 'https://kubernetes-charts-incubator.storage.googleapis.com',
+        })
+
+        # creating at least one release is necessary to trigger the Tiller installation
```
de-code: Where is that coming from? Is that just for a test?

src/buildercore/terraform.py
```diff
@@ -816,6 +818,46 @@ def _render_eks_workers(context, template):
         }
     })
 
+    if context['eks']['helm']:
```
de-code: This function is massive (required many clicks to get to the function signature). Maybe it would deserve being broken up?

src/buildercore/terraform.py
giorgiosironi: Otherwise Tiller wouldn't be installed, as it is done in a lazy way starting from the `helm_release` resource (at the moment).

src/buildercore/terraform.py
giorgiosironi: This part should definitely be extracted

src/buildercore/terraform.py
de-code: Is that of the way builder is working or where is the _lazyness_ coming from? I may be misreading it, but we probably don't want to install hello world into every cluster?
`helm init` seems to install `tiller` otherwise.

giorgiosironi: The `helm` Terraform provider is easy to authenticate with the Kubernetes cluster, whereas `helm init` isn't and is outside of Terraform management. This provider has the defect of requiring at least 1 chart to be installed, but it seems it's improving over time.  I'll rename the resource because it doesn't really execute an hello world, it's simply an empty chart that could be used to add any Kubernetes resource that we want to be standard in our clusters.

PR #494, opened 2019-04-01
Fix: fixes a bug where setting 'subdomain' to false causes a type error

lsh-0:


src/buildercore/cfngen.py
```diff
@@ -467,7 +467,8 @@ def complete_domain(host, default_main):
     return host + '.' + default_main # something + '.' + elifesciences.org
 
 def build_context_subdomains(pdata, context):
-    context['subdomains'] = [complete_domain(s, pdata['domain']) for s in pdata['aws'].get('subdomains', [])]
```
giorgiosironi: these are CNAMEs IIRC

src/buildercore/cfngen.py
```diff
@@ -467,7 +467,8 @@ def complete_domain(host, default_main):
...
+    # note! a distinction is being made between 'subdomain' and 'subdomains'
+    context['subdomains'] = [complete_domain(s, pdata['domain']) for s in pdata['aws'].get('subdomains', []) if pdata['domain']]
```
giorgiosironi: I think we resorted to comment out `subdomains` whenever `domain` was not available.

src/buildercore/cfngen.py
lsh-0: good to know, I wasn't sure what the distinction was

PR #496, opened 2019-04-09
Feature: skips updating tags on RDS

lsh-0:
regarding this: https://github.com/elifesciences/issues/issues/4860#issuecomment-481083463

another approach might be comparing sorted versions of maps..

### DIFF
--- a/src/buildercore/aws.py
+++ b/src/buildercore/aws.py
@@ -6,8 +6,8 @@ def generic_tags(context, name=True):
     tags = OrderedDict([
         ('Project', context['project_name']), # "journal"
         ('Environment', context['instance_id']), # "prod"
-        ('Cluster', context['stackname']), # "journal--prod"
         # the name AWS Console uses to label an instance
         ('Name', context['stackname']), # "journal--prod"
+        ('Cluster', context['stackname']), # "journal--prod"
     ])
     return tags
--- a/src/buildercore/aws.py
+++ b/src/buildercore/aws.py
@@ -7,7 +7,7 @@ def generic_tags(context, name=True):
         ('Project', context['project_name']), # "journal"
         ('Environment', context['instance_id']), # "prod"
         ('Cluster', context['stackname']), # "journal--prod"
+        # the name AWS Console uses to label an instance
+        ('Name', context['stackname']), # "journal--prod"
     ])
-    # the name AWS Console uses to label an instance
-    tags['Name'] = context['stackname'] # "journal--prod"
     return tags
--- a/src/tests/test_buildercore_terraform.py
+++ b/src/tests/test_buildercore_terraform.py
@@ -883,8 +883,8 @@ class TestBuildercoreTerraform(base.BaseCase):
                 'tags': {
                     'Project': 'project-with-eks',
                     'Environment': self.environment,
-                    'Cluster': 'project-with-eks--%s' % self.environment,
                     'Name': 'project-with-eks--%s' % self.environment,
+                    'Cluster': 'project-with-eks--%s' % self.environment,
                 }
             }
         )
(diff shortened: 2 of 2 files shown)


giorgiosironi:
Tags weren't ordered until recently, https://github.com/elifesciences/builder/commit/19e45bd7eaa48d64c19e110f7fa948c324d4dfd0 started to specify an `OrderedDict` that results in the "new" order. The delta to be applied is transitional to get to the new order, but no further change to the code should be necessary.

lsh-0:
> is transitional to get to the new order, but no further change to the code should be necessary.

well, something is necessary because a bad comparison is being done resulting in a spurious upgrade. If we don't want to affect the tags I'll update this PR to compare two ordered things. That should handle however long and however many project this transitional period affects

giorgiosironi:
The old order is unspecified but consistent:
```
$ for i in `grep -rl AttachedDB .cfn/stacks/ | grep 'end2end\|prod'`; do cat $i | jq .Resources.AttachedDB.Properties.Tag[].Key
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"                                                                                                                                                     
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
"Project"
"Environment"
"Name"
"Cluster"
```

lsh-0:
is that from the master-server? it's not that I don't believe you, I 100% do, it's just I have evidence to the contrary :)

more than happy to remove my snippet of code that ignores RDS tags if we have a better way of

lsh-0:
so - was there a reason that tag was left out?

giorgiosironi:
> is that from the master-server?

It's from `.cfn/stacks/` which is downloaded from `s3://elife-builder/stacks`. No tag was left out, the order seems to have changed when switching from a `{}` to `OrderedDict`.

Doing some tests with `end2end` stacks. Now the databases that have been already updated will have to go back to the original order (`elife-dashboard` for example), but the rest of the population should not be updated.

giorgiosironi:
This is also affecting `ElasticLoadBalancer` and `ExtraStorage` resources, but the fix is the same.

`lax--end2end`, `elife-metrics--end2end`, `elife-dashboard--end2end` all have an empty delta.
`elife-metrics--end2end` has an empty delta.

src/buildercore/cfngen.py
```diff
@@ -633,6 +634,11 @@ def _title_has_been_updated(title, section):
             if title_in_old['Type'] == 'AWS::EC2::Instance':
                 for property_name in EC2_NOT_UPDATABLE_PROPERTIES:
                     title_in_new['Properties'][property_name] = title_in_old['Properties'][property_name]
+
+        if 'Type' in title_in_old and title_in_old['Type'] == 'AWS::RDS::DBInstance':
+            for property_name in RDS_NOT_UPDATEABLE_PROPERTIES:
+                title_in_new['Properties'][property_name] = title_in_old['Properties'][property_name]
+
```
giorgiosironi: remove changes to this file if satisfied. Rebasing and removing the commit would be the best way since there's no squash in the project

lsh-0: change reverted. I don't mind git log noise.

giorgiosironi:
Made the reverted commit vanish.

PR #497, opened 2019-04-15
Feature: adds the large-repo-wrangler project

lsh-0:
for elife-publications git-lfs wrangling

Nothing terribly complex here or in the [related formula](https://github.com/elifesciences/large-repo-wrangler-formula), just a collection of tools to help deal with awkward third party repositories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ settings.yml
 cloned-projects/
 builder-private/
 builder-private-example/
-builder-configuration/
 custom-vagrant/
 
 .vagrant*
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ settings.yml
 cloned-projects/
 builder-private/
 builder-private-example/
+builder-configuration/
 custom-vagrant/
 
 .vagrant*
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -2118,3 +2118,17 @@ kubernetes-aws:
                 max-size: 3
                 desired-capacity: 3
             helm: true
+
+large-repo-wrangler:
+    description: large git repo wrangler
+    domain: False
+    intdomain: False
+    aws:
+        ec2: {}
+        ext:
+            type: gp2
+            size: 20 # GB
+            device: /dev/sdh
+        ports:
+            - 22
+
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -8,6 +8,7 @@ settings.yml
 cloned-projects/
 builder-private/
 builder-private-example/
+builder-configuration/
```
giorgiosironi: https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore would help since not everyone keeps this repositories in there

projects/elife.yaml
```diff
@@ -2118,3 +2118,19 @@ kubernetes-aws:
...
                 desired-capacity: 3
             helm: true
+
+large-repo-wrangler:
+    description: when elife-publications requires a large git repository to be wrangled. adhoc instances only.
+    domain: False
+    intdomain: False
+    formula-repo: https://github.com/elifesciences/large-repo-wrangler-formula
+    aws:
+        ec2: {}
```
giorgiosironi: wait, why do you need this? It could be omitted as it's inherited?

projects/elife.yaml
```diff
@@ -2118,3 +2118,19 @@ kubernetes-aws:
...
+        ext:
+            type: gp2
+            size: 20 # GB
```
giorgiosironi: space is cheap so raise if needed

projects/elife.yaml
lsh-0: that's a good point, let me test that ...

lsh-0: appears to work. I'll change it.

projects/elife.yaml
lsh-0: I think 9GB is the largest repo I've had to deal with so far, most are just large enough to cause editorial problems with the github import process

.gitignore
lsh-0: bet it helps those who do ;)

I agree in principal though. Are you worried about another `builder-configuration` directory being created that isn't the current `builder-configuration` project and it missing out on being committed?

The whole ignore file has legacy crap in there. I'll open a PR to clean it all out and take a second look at the global gitignore then

giorgiosironi: Thank you, not worried about clashes; it's plain old information hiding (other people don't need to see or know about these folders that sometimes exist but not for them.)

lsh-0: I disagree, but it's a piffling point. When the great AI revolution comes, we'll all be up against the wall.

:robot: :skull:

PR #499, opened 2019-04-26
Feature: linting

lsh-0:
* cleans up .gitignore

cc @giorgiosironi

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,25 +6,20 @@ settings.yml
 *~
 
 cloned-projects/
-builder-private/
-builder-private-example/
 custom-vagrant/
 
-.vagrant*
-payload.tar.gz
-stacks/*.json
+.vagrant/
+.vagrant-error
 *.orig
 *.log
 venv/
 *.swp
 .cfn/
 logs/
-public/
 private/
 downloads
 .coverage.*
 .coverage
-projects/*/
 scripts/customize.sh
 scripts/salt/*.minion
 /custom-vagrant/salt/


lsh-0:
I'm thinking this could be an opportunity to pull out the `custom/vagrant/` stuff, maybe the `scripts/customize.sh` script and whatever is going on with `scripts/salt/*.minion`

.gitignore
```diff
@@ -6,25 +6,20 @@ settings.yml
...
 .cfn/
 logs/
-public/
 private/
 downloads
 .coverage.*
 .coverage
-projects/*/
 scripts/customize.sh
 scripts/salt/*.minion
```
giorgiosironi: These are generated scripts that probably need to reside there to be visible inside `/vagrant`

lsh-0: I don't know about that. That directory used to contain project file overrides. You could change the default region for the 'elife' organisation to 'eu' with a tiny bit of yaml. I don't recall ever storing scripts in there.

I can't see any paths being built that reference `./projects/*/something` by grepping the source dirs. Can you point to where you think this may be happening?

lsh-0: were you thinking of `./scripts/salt/*.minion` ?

giorgiosironi: Yes:
```
$ ls scripts/salt/ | head
annotations--vagrant.minion
api-dummy--vagrant.minion
api-gateway--vagrant.minion
basebox--vagrant.minion
bastion--vagrant.minion
containers--vagrant.minion
crm--vagrant.minion
digests--vagrant.minion
elife-alfred--vagrant.minion
elife-api--vagrant.minion
```

PR #500, opened 2019-05-03
Feature: master-server, vagrant, opens vault port

lsh-0:
fyi @giorgiosironi

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -372,6 +372,8 @@ master-server:
     vagrant:
         # 16.04, deprecated
         box: bento/ubuntu-16.04
+        ports:
+            8200: 8200 # Vault GUI
 
 lax:
     description: article-json store


projects/elife.yaml
```diff
@@ -372,6 +372,8 @@ master-server:
     vagrant:
         # 16.04, deprecated
         box: bento/ubuntu-16.04
+        ports:
+            8200: 8200 # Vault GUI
```
giorgiosironi: yes, it's already open on AWS
Though it's not the UI port, it's just Vault's port hosting both a REST API and the `/ui`.

lsh-0: Don't quite follow - it's not the UI port but it hosts `/ui` ?

giorgiosironi: It hosts both `/api` and `/ui` (not sure if the first path is correct, but you get the point)

PR #501, opened 2019-05-10
Feature: shift search to 1604

lsh-0:


lsh-0:
derp - I branched off of journal-cms changes

PR #503, opened 2019-05-17
Fix: fail hard when bootstrap user's private key doesn't exist on the fs

lsh-0:
encountered while updating the journal-cms instances

I *think* it may have to do with never having performed any action on that instance on my machine while doing the update, hence no opportunity to download the bootstrap key. I have a hack elsewhere that gets around that case that I'll add to this PR once I integrate it better

### DIFF
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -246,12 +246,16 @@ def stack_pem(stackname, die_if_exists=False, die_if_doesnt_exist=False):
 
 def _ec2_connection_params(stackname, username, **kwargs):
     "returns a dictionary of settings to be used with Fabric's api.settings context manager"
+    # http://docs.fabfile.org/en/1.14/usage/env.html
     params = {'user': username}
     pem = stack_pem(stackname)
     # handles cases where we want to establish a connection to run a task
     # when machine has failed to provision correctly.
-    if os.path.exists(pem) and username == config.BOOTSTRAP_USER:
-        params['key_filename'] = pem
+    if username == config.BOOTSTRAP_USER:
+        if os.path.exists(pem):
+            params['key_filename'] = pem
+        else:
+            raise RuntimeError("private key for the bootstrap user for this host does not exist. I looked here: %s" % pem)
     params.update(kwargs)
     return params
 


giorgiosironi:
If this was already a failure, then by all means make it a readable failure.

lsh-0:
> make it a readable failure.

The error happened in a parallel worker and was swallowed up with no output. This was part of the debugging.

giorgiosironi:
Just had this error in a `update_infrastructure` run, not sure if it's related:
```
Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
  File "/home/giorgio/code/builder/venv/local/lib/python2.7/site-packages/fabric/main.py", line 757, in main
    *args, **kwargs
  File "/home/giorgio/code/builder/venv/local/lib/python2.7/site-packages/fabric/tasks.py", line 426, in execute
    results['<local-only>'] = task.run(*args, **new_kwargs)
  File "/home/giorgio/code/builder/venv/local/lib/python2.7/site-packages/fabric/tasks.py", line 173, in run
    return self.wrapped(*args, **kwargs)
  File "/home/giorgio/code/builder/src/decorators.py", line 20, in wrap
    result = fn(*args, **kw)
  File "/home/giorgio/code/builder/src/cfn.py", line 109, in update_infrastructure
    buildvars.refresh(stackname, context)
  File "/home/giorgio/code/builder/src/decorators.py", line 97, in call
    return func(stackname, *args, **kwargs)
  File "/home/giorgio/code/builder/src/buildvars.py", line 152, in refresh
    keypair.download_from_s3(stackname)
  File "/home/giorgio/code/builder/src/buildercore/keypair.py", line 36, in download_from_s3
    expected_path = stack_pem(stackname, die_if_exists=True)
  File "/home/giorgio/code/builder/src/buildercore/core.py", line 244, in stack_pem
    raise EnvironmentError("keypair %r found at %r, not overwriting." % (stackname, expected_key))
EnvironmentError: keypair 'journal--continuumtest' found at '/home/giorgio/code/builder/.cfn/keypairs/journal--continuumtest.pem', not overwriting.
```

lsh-0:
yeah I get them too, I thought it might be related to the upgrade process for multi-node systems. Not sure if it's related to this ticket. If anything it might be related to the extra download of the keypair in the recent PR to fix that other problem. I'll take a closer look when I get it again.

lsh-0:
just for future reference, Giorgio's problem was unrelated to this and was solved and merged. 

This is still a viable PR but the build failed. Kicking

giorgiosironi:
Makes https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Fapi-dummy/detail/PR-152/1/pipeline fail. Checking if there is a quick solution or have to revert.

src/buildercore/core.py
```diff
@@ -246,12 +246,16 @@ def stack_pem(stackname, die_if_exists=False, die_if_doesnt_exist=False):
...
     pem = stack_pem(stackname)
     # handles cases where we want to establish a connection to run a task
     # when machine has failed to provision correctly.
-    if os.path.exists(pem) and username == config.BOOTSTRAP_USER:
-        params['key_filename'] = pem
+    if username == config.BOOTSTRAP_USER:
+        if os.path.exists(pem):
+            params['key_filename'] = pem
+        else:
+            raise RuntimeError("private key for the bootstrap user for this host does not exist. I looked here: %s" % pem)
```
giorgiosironi: I think that then this doesn't cover the use case of using your own `.ssh/id_rsa` key rather than the one in `builder`. If you are not the user that created the stack, probably won't have that one. Or should it be synced from S3?

giorgiosironi: In the case of https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Fapi-dummy/detail/PR-152/1/pipeline it's definitely on S3 so there is a missing auto-download step.

giorgiosironi:
Haven't found a quick fix because we would have to call `keypair.download_from_s3` but can't do so from `buildercore.core` as it would create a mutual dependency betwen the two modules. I guess somewhere in `buildercore.lifecycle` is the right place, but in the meantime I'd revert to fix the build.

src/buildercore/core.py
lsh-0: > Or should it be synced from S3?
>  it's definitely on S3 so there is a missing auto-download step.

yup, it should be automatically downloaded from S3. It's making that particular failure explicit. I encountered it a handful of times while doing the upgrades.

lsh-0:
I'll revert the commit

PR #509, opened 2019-05-31
fix, skips building fastly context when domain=false

lsh-0:


src/buildvars.py
```diff
@@ -124,6 +124,9 @@ def _update_remote_bvars(stackname, buildvars):
 def refresh(stackname, context):
     "(safely) replaces the buildvars file on the ec2 instance(s)"
 
+    from master import download_keypair
+    download_keypair(stackname)
```
lsh-0: so this hack has been hanging around in my git stash for a while now and needed to be committed. obviously it's not it's final form, but it solves an issue where you run `update_infrastructure` on a stack that your builder installation hasn't visited before.

giorgiosironi: `download_keypair` seems like a safe operation for an existing stack. Move up the `import` to the top?

giorgiosironi: Not sure why it is in `master` however, it's generic for any stack that has an `ec2` configuration.

src/buildercore/cfngen.py
```diff
@@ -409,7 +409,7 @@ def _parameterize_bigquerylogging(bigquerylogging):
         return bigquerylogging
 
     context['fastly'] = False
-    if pdata['aws'].get('fastly'):
+    if pdata['aws'].get('fastly') and pdata['domain']:
```
giorgiosironi: seems fair as `domain` is the top-level `elifesciences.org` used to build DNS entries; hence if it's not around it means no DNS entries are possible and hence no CDN can be setup.

I probably comment out `fastly` along with setting `"domain": False` when migrating infrastructure.

lsh-0: I'll include your comment in the code

src/buildercore/keypair.py
```diff
@@ -35,6 +36,7 @@ def download_from_s3(stackname):
     expected_path = stack_pem(stackname, die_if_exists=True)
     s3.download(s3_keypair_key(stackname), expected_path)
     stack_pem(stackname, die_if_doesnt_exist=True)
+    local('chmod 400 -R %s' % expected_path) # is the -R (recursive) necessary?
```
giorgiosironi: right, `-R` is not needed because it's a single file, can remove

lsh-0: done

PR #510, opened 2019-05-31
Feature: adds a 'builder' project to the projects file

lsh-0:
The formula is here: https://github.com/elifesciences/builder-formula

Intended only as a common environment when one is necessary.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -275,6 +275,11 @@ defaults:
         #    1239: 80
         #    1240: 8001
 
+builder:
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -275,6 +275,11 @@ defaults:
         #    1239: 80
         #    1240: 8001
 
+builder:
+    formula-repo: https://github.com/elifesciences/builder-formula
+    vagrant:
+        box: bento/ubuntu-16.04
```
lsh-0: to be consistent with your environment, @giorgiosironi

PR #513, opened 2019-06-07
projects, elife-bot, adds removal policy for s3 buckets in s1604 envs

lsh-0:
fyi @giorgiosironi

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -769,6 +769,19 @@ elife-bot:
             # elife-bot-event-property--prod, elife-bot-event-property--end2end, etc.
             - elife-bot-event-property--{instance}
     aws-alt:
+        s1604:
+            s3:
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -769,6 +769,19 @@ elife-bot:
             # elife-bot-event-property--prod, elife-bot-event-property--end2end, etc.
             - elife-bot-event-property--{instance}
     aws-alt:
+        s1604:
```
giorgiosironi: this is the same as `standalone` so correct solution. We should consider dropping it to just `s3: false`? It's not very useful to create buckets when the goal is to test Salt formulas

lsh-0: it was honestly just a face-roll of a change to get something working. 

I agree though. If highstate in standalone doesn't require those buckets to be there, get rid of them

PR #514, opened 2019-06-07
[wip] Feature: projects, adds a 'bioprotocol' project

lsh-0:


projects/elife.yaml
```diff
@@ -2169,3 +2169,27 @@ microsimulation-demo:
...
+            - 22
+            - 80
+            - 443
+        sns:
+            # endpoint for BioProtocol to publish updates to article protocol data
+            - bioprotocol-updates--{instance}
+        sqs:
+            # single queue that receives messages about:
+            # - updates to articles (internally)
+            # - updates to article protocol data (from BioProtocol)
```
giorgiosironi: these will be wildly different messages, but if the application is smart enough

projects/elife.yaml
```diff
@@ -2169,3 +2169,27 @@ microsimulation-demo:
...
+    repo: https://github.com/elifesciences/bioprotocol-service
+    aws:
+        type: t2.micro
+        ports:
+            - 22
+            - 80
+            - 443
+        sns:
+            # endpoint for BioProtocol to publish updates to article protocol data
+            - bioprotocol-updates--{instance}
```
giorgiosironi: https://github.com/elifesciences/issues/issues/4930#issuecomment-498144947 states they would be writing to SQS; it doesn't make much of a difference but has this changed?

projects/elife.yaml
```diff
@@ -2169,3 +2169,27 @@ microsimulation-demo:
         ports:
             1280: 80
             1281: 8081
+
+bioprotocol:
+    repo: https://github.com/elifesciences/bioprotocol-service
```
giorgiosironi: no formula mentioned? This influences some `builder` behavior, I think

projects/elife.yaml
lsh-0: I thought that was a mistake. Writing to a queue doesn't seem correct to me. Writing to a topic and having queues subscribed to those topics seems more-correct. Or I could just write an API endpoint they can PUT/POST to directly. I'll follow this point up with them

projects/elife.yaml
lsh-0: dispatching a message based on the shape of the data seems like a pretty basic thing to do.

projects/elife.yaml
lsh-0: I hadn't gotten to that, I'll do it today.

lsh-0: done

projects/elife.yaml
lsh-0: I've contacted Dong. He seems like a cheerful guy.

giorgiosironi: Queues (SQS) are meant for 1-to-1 communication, whereas topics/fanouts (SNS) are meant for 1-to-N communication.

lsh-0: I know, but that seems so very limited to almost be a side effect of the larger design AWS has with SNS+SQS. There is an awful lot I can do while I wait for Dong to confirm so no urgency.

giorgiosironi: With different names, it is a general pattern of messaging systems:
https://www.enterpriseintegrationpatterns.com/patterns/messaging/PointToPointChannel.html
https://www.enterpriseintegrationpatterns.com/patterns/messaging/PublishSubscribeChannel.html

lsh-0: yes, thank you Giorgio. (this is getting a little condescending, please stop.)

lsh-0: they're happy to use POST. I've removed the SNS topic and SQS subscription

PR #517, opened 2019-06-18
Apply T2 unlimited to all search instances

giorgiosironi:
After another case of `search--prod` running out of credits, let's just roll it out everywhere.


giorgiosironi:
Applied to `end2end`, `continuumtest`; https://alfred.elifesciences.org/job/test-search/327/ passes.

lsh-0:
poor Search really does get hammered. elife-metrics too, but it manages to keep up

PR #518, opened 2019-06-26
updates several project descriptions with elastic ip allocations

lsh-0:


projects/elife.yaml
```diff
@@ -1053,6 +1053,7 @@ elife-api:
     repo: https://github.com/elifesciences/elife-api
     formula-repo: https://github.com/elifesciences/elife-api-formula
     aws:
+        # ip: 52.72.250.79 # legacy--api.elifesciences.org
```
giorgiosironi: these should only be in the `aws-alt.prod` or `aws-alt.legacy` section? Since it only applies to one instance.
I don't think we have any support for automatically attaching the ips unfortunately.

lsh-0: no, we don't have support right now. they're just comments and just for reference should we find the time to add elastic IPs

lsh-0: I discovered the other day that elastic IPs only cost money if they're associated, otherwise they're free

giorgiosironi: They cost money if they are _not_ associated?

lsh-0: yeah, I'll rustle up a link

lsh-0: sorry - I made a mistake in my comment

PR #519, opened 2019-07-15
'stop' command now stops ec2 instances only by default, as intended

lsh-0:


lsh-0:
the problem is `services` is a tuple and not a list

PR #521, opened 2019-07-15
Update AMI for 1804 to ami-0cb0ec9b5651415e7

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-builder-ami-update/6/display/redirect which resulted in this pull request.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -45,7 +45,7 @@ defaults:
             suppressed: []
             # find more here: http://cloud-images.ubuntu.com/releases/
             # Ubuntu 18.04
-            ami: ami-00aedeb287fe8c9d6 # GENERATED created from basebox--1804
+            ami: ami-0cb0ec9b5651415e7 # GENERATED created from basebox--1804
                                        # us-east-1, build date 20180814, hvm:ebs-ssd, AMI built on 20180828
             # use a master server or go ronin?
             masterless: false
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Do we want to do this on a schedule?

lsh-0:
the creation of new AMIs? I think we absolutely should. Extra points if we can also:

* update the project file afterwards
* automatically open a PR and assign a reviewer (just like this)
* add the new LTS when it comes out

giorgiosironi:
Current status:

- [x] update the project file afterwards
- [x] automatically open a PR 
- [ ] and assign a reviewer (just like this)
- [ ] add the new LTS when it comes out

but https://alfred.elifesciences.org/job/process/job/process-basebox-1804/ doesn't run automatically, only on demand.

What should the frequency be? This is implemented for 18.04 but we do support 16.04 so the process should be duplicated for that.

lsh-0:
weekly? monthly? whenever builder-base changes? whenever the base ami changes? I'm not sure. It doesn't need to happen often I think, it's going to affect new instances only which means formula PRs 

I would go with weekly to be honest. That seems just often enough for it to become routine and to shake out the issues peculiar to it.

giorgiosironi:
Set https://alfred.elifesciences.org/job/process/job/process-basebox-1804/ to run weekly on Monday at 3 AM.

PR #522, opened 2019-07-16
Feature: adds a 'restart all' command

lsh-0:
adds `restart_all_running_ec2` command that will restart all ec2 instances that are running

* maintains state so problems that force the task to quit won't see instances restarted again
* supports project ordering, so certain projects go first
* supports project exclusion, so certain projects are not restarted 
* uses `buildercore.lifecycle.restart` that stops nodes in order
* prompt to continue restarting instances if an error occurs
* skips nodes that aren't running
* adds `askmaster.update_kernel` to update the kernel images on 16.04 and 18.04 machines

### DIFF
--- a/src/buildercore/lifecycle.py
+++ b/src/buildercore/lifecycle.py
@@ -42,12 +42,12 @@ def push(lst, rec):
     if not lst or lst[-1] != rec:
         lst.append(rec)
 
-def restart(stackname):
+def restart(stackname, initial_states='pending|running|stopping|stopped'):
     """for each ec2 node in given stack, ensure ec2 node is stopped, then start it, then repeat with next node.
     rds is started if stopped (if *exists*) but otherwise not affected"""
-    start_rds_nodes(stackname)
+    #start_rds_nodes(stackname) # something a bit buggy here
 
-    node_list = find_ec2_instances(stackname, state='pending|running|stopping|stopped')
+    node_list = find_ec2_instances(stackname, state=initial_states, allow_empty=True)
 
     history = []
 
@@ -74,7 +74,9 @@ def restart(stackname):
                 update_msg="waiting for nodes to be networked",
                 done_msg="all nodes have public ips"
             )
-        update_dns(stackname)
+        if history:
+            # only update dns if we have a history of affected nodes
+            update_dns(stackname)
         return history
     except Exception:
         LOG.info("Partial restart history of %s: %s", stackname, pformat(history))
--- a/src/buildercore/lifecycle.py
+++ b/src/buildercore/lifecycle.py
@@ -45,7 +45,7 @@ def push(lst, rec):
 def restart(stackname, initial_states='pending|running|stopping|stopped'):
     """for each ec2 node in given stack, ensure ec2 node is stopped, then start it, then repeat with next node.
     rds is started if stopped (if *exists*) but otherwise not affected"""
-    #start_rds_nodes(stackname) # something a bit buggy here
+    # start_rds_nodes(stackname) # something a bit buggy here
 
     node_list = find_ec2_instances(stackname, state=initial_states, allow_empty=True)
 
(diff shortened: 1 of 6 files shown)


lsh-0:
it's still a WIP and a bit rough, but I'm running it now without issue. `src/lifecycle.py` probably not it's final home

src/lifecycle.py
```diff
@@ -24,8 +30,75 @@ def stop(stackname, *services):
...
 @requires_aws_stack
 @timeit
+@echo_output
 def restart(stackname):
-    lifecycle.restart(stackname)
+    return lifecycle.restart(stackname)
+
+
+@task
+def restart_all(statefile):
```
giorgiosironi: as for location, yes, a separate `tasks` module would help as it's not as generic as the rest of `lifecycle`

src/buildercore/lifecycle.py
```diff
@@ -42,12 +42,12 @@ def push(lst, rec):
     if not lst or lst[-1] != rec:
         lst.append(rec)
 
-def restart(stackname):
+def restart(stackname, initial_states='pending|running|stopping|stopped'):
```
giorgiosironi: why not an actual data structure (list? set?) here rather than a `|`-separated string?

src/lifecycle.py
giorgiosironi: thanks to `statefile` being passed it's very clear where the local state is and how to reset it

src/lifecycle.py
```diff
@@ -24,8 +30,75 @@ def stop(stackname, *services):
...
+    todo = filter(lambda stackname: pname(stackname) not in dont_do, todo)
+
+    print('todo:')
+    print(json.dumps(todo, indent=4))
+
+    with open(statefile, 'r') as fh:
+        done = fh.read().split('\n')
+
+    with open(statefile, 'a') as fh:
+        print('writing state to', fh.name)
```
giorgiosironi: would use a logger over `print` in general, even after closing a terminal an audit remains in the log file just in case

src/buildercore/lifecycle.py
lsh-0: it's how it's handled higher up in buildercore. I dimly recall it having something to do with boto2 but could be mistaken. I'm not going to change it for this PR but I think there are lots of little places where the libraries have improved that we could revisit and trim some code.

src/lifecycle.py
lsh-0: could you see this or a generic version of this being used to automatically restart instances weekly?

if they ran in parallel, or batches ran in parallel, it could all be done very quickly.

src/lifecycle.py
lsh-0: done

src/lifecycle.py
lsh-0: moved to `tasks`

src/buildercore/lifecycle.py
```diff
@@ -42,12 +42,12 @@ def push(lst, rec):
...
     """for each ec2 node in given stack, ensure ec2 node is stopped, then start it, then repeat with next node.
     rds is started if stopped (if *exists*) but otherwise not affected"""
-    start_rds_nodes(stackname)
+    # start_rds_nodes(stackname) # something a bit buggy here
```
lsh-0: not sure if this should be kept disabled. I've seen the command fail here but don't have a stacktrace to confirm it.

... I'm not sure if anything else actually uses `restart` to be honest

lsh-0:
related pr: https://github.com/elifesciences/it-admin/pull/8

src/buildercore/lifecycle.py
giorgiosironi: as far as I'm aware we don't use it in Jenkins, only separate `stop` and `start` commands in there

PR #523, opened 2019-07-16
Retry failed instance boot

giorgiosironi:
For https://github.com/elifesciences/elife-xpub/issues/2283

### DIFF
--- a/src/buildercore/config.py
+++ b/src/buildercore/config.py
@@ -113,6 +113,7 @@ def get_logger(name):
 BUILDER_BUCKET = 'elife-builder'
 BUILDER_REGION = 'us-east-1'
 BUILDER_NON_INTERACTIVE = 'BUILDER_NON_INTERACTIVE' in os.environ and os.environ['BUILDER_NON_INTERACTIVE']
+BUILDER_TIMEOUT = 'BUILDER_TIMEOUT' in os.environ and int(os.environ['BUILDER_TIMEOUT'])
 KEYPAIR_PREFIX = 'keypairs/'
 CONTEXT_PREFIX = 'contexts/'
 
--- a/src/buildercore/config.py
+++ b/src/buildercore/config.py
@@ -113,7 +113,10 @@ def get_logger(name):
 BUILDER_BUCKET = 'elife-builder'
 BUILDER_REGION = 'us-east-1'
 BUILDER_NON_INTERACTIVE = 'BUILDER_NON_INTERACTIVE' in os.environ and os.environ['BUILDER_NON_INTERACTIVE']
-BUILDER_TIMEOUT = 'BUILDER_TIMEOUT' in os.environ and int(os.environ['BUILDER_TIMEOUT'])
+if 'BUILDER_TIMEOUT' in os.environ:
+    BUILDER_TIMEOUT = int(os.environ['BUILDER_TIMEOUT'])
+else:
+    BUILDER_TIMEOUT = 600
 KEYPAIR_PREFIX = 'keypairs/'
 CONTEXT_PREFIX = 'contexts/'
 
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -274,7 +274,7 @@ def stack_conn(stackname, username=config.DEPLOY_USER, node=None, **kwargs):
 class NoPublicIps(Exception):
     pass
 
-def stack_all_ec2_nodes(stackname, workfn, username=config.DEPLOY_USER, concurrency=None, node=None, **kwargs):
+def stack_all_ec2_nodes(stackname, workfn, username=config.DEPLOY_USER, concurrency=None, node=None, instance_ids=None, **kwargs):
     """Executes work on all the EC2 nodes of stackname.
     Optionally connects with the specified username"""
     work_kwargs = {}
@@ -286,6 +286,9 @@ def stack_all_ec2_nodes(stackname, workfn, username=config.DEPLOY_USER, concurre
     nodes = {ec2['InstanceId']: int(tags2dict(ec2['Tags'])['Node']) if 'Node' in tags2dict(ec2['Tags']) else 1 for ec2 in data}
     if node:
         nodes = {k: v for k, v in nodes.items() if v == int(node)}
+        public_ips = {k: v for k, v in public_ips.items() if k in nodes.keys()}
+    elif instance_ids:
+        nodes = {k: v for k, v in nodes.items() if k in instance_ids}
         public_ips = {k: v for k, v in public_ips.items() if k in nodes.keys()}
 
     params = _ec2_connection_params(stackname, username)
--- a/src/buildercore/lifecycle.py
+++ b/src/buildercore/lifecycle.py
@@ -135,7 +135,7 @@ def start(stackname):
         # in case of botched boot and/or inability to
         # access through SSH, try once to stop and
         # start the instances again
-        LOG.info("Retrying boot on %s", ec2_to_be_checked)
+        LOG.info("Boot failed (instance(s) not accessible through SSH). Attempting boot one more time: %s", ec2_to_be_checked)
         _ec2_connection(stackname).instances.filter(InstanceIds=ec2_to_be_checked).stop()
         _wait_ec2_all_in_state(stackname, 'stopped', ec2_to_be_checked)
         _ec2_connection(stackname).instances.filter(InstanceIds=ec2_to_be_checked).start()
--- a/src/buildercore/lifecycle.py
+++ b/src/buildercore/lifecycle.py
@@ -17,6 +17,9 @@ from .context_handler import load_context
 
 LOG = logging.getLogger(__name__)
 
+class EC2Timeout(RuntimeError):
+    pass
+
 def _node_id(node):
     name = core.tags2dict(node.tags)['Name']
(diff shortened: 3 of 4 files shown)


giorgiosironi:
I broke everything, but it's a proof of concept.

giorgiosironi:
To test this I run:
```
$ bldr cmd:elife-libraries--ci,"sudo systemctl stop sshd"
```
which convinces builder the instance is `running` (hasn't changed on the EC2 console) but not accessible through SSH and hence needs a reboot.

giorgiosironi:
Accidentally when doing the first draft I turned off all EC2 instances in the region:
https://elifesciences.slack.com/archives/C6N559E2F/p1563283559000200

lsh-0:
holy hell - all those alerts were legitimate? too funny. I thought it was a NewRelic reporting issue as they sent an alert afterwards.

I think you need to get that :hankey: off of @diversemix ;)

There were 67 instances running yesterday as I was rebooting everything, there are 64 running now. Investigating ...

src/buildercore/lifecycle.py
```diff
@@ -111,17 +116,30 @@ def start(stackname):
...
         )
-    else:
-        LOG.info("EC2 nodes are all running")
+    try:
+        _wait_for_ec2_steady_state()
+    except EC2Timeout:
+        # in case of botched boot and/or inability to
+        # access through SSH, try once to stop and
+        # start the instances again
+        LOG.info("Retrying boot on %s", ec2_to_be_checked)
```
lsh-0: perhaps `"Boot failed. Attempting boot one more time: %s"`

src/buildercore/lifecycle.py
```diff
@@ -111,17 +116,30 @@ def start(stackname):
...
+        _ec2_connection(stackname).instances.filter(InstanceIds=ec2_to_be_checked).stop()
+        _wait_ec2_all_in_state(stackname, 'stopped', ec2_to_be_checked)
+        _ec2_connection(stackname).instances.filter(InstanceIds=ec2_to_be_checked).start()
```
lsh-0: we're not doing any caching of the connection and I don't think the boto library would be re-using connections either. I wonder if there is any overhead or extra noise because of the new connections. And if it's even worth a second thought..

lsh-0:
:+1:, apart from the message. I think it could be clearer that boot failed and it's going to attempt it just one last time.

lsh-0:
> To test this I run:
> $ bldr cmd:elife-libraries--ci,"sudo systemctl stop sshd"

that's pretty clever. I was just looking at the code for that `cmd` task. I thought it was maybe sending the command to the salt master, in which case you could turn sshd on and off as you wish, but in this case you could only ever turn it off but not connect to turn it back on again, tricking builder etc etc.

We have `askmaster.py` as well which does what I thought was happening above.

giorgiosironi:
Scary to work on this stuff when a test suite could turn off all the `prod` environments due to a bug. We'll do the post mortem today and hopefully come up with a way to make this safer.

giorgiosironi:
Also not sure about the failures that seem related to `restart`, which shouldn't have been touched here.

giorgiosironi:
Ah, the timeout has been changed, probably reduced.

giorgiosironi:
Needs some testing on `end2end` and similar before merging.

src/buildercore/config.py
```diff
@@ -113,6 +113,10 @@ def get_logger(name):
 BUILDER_BUCKET = 'elife-builder'
 BUILDER_REGION = 'us-east-1'
 BUILDER_NON_INTERACTIVE = 'BUILDER_NON_INTERACTIVE' in os.environ and os.environ['BUILDER_NON_INTERACTIVE']
+if 'BUILDER_TIMEOUT' in os.environ:
+    BUILDER_TIMEOUT = int(os.environ['BUILDER_TIMEOUT'])
+else:
+    BUILDER_TIMEOUT = 600
```
lsh-0: it's pedantic, but `BUILDER_TIMEOUT = int(os.environ.get('BUILDER_TIMEOUT', 600))` if you're in the whole brevity thing.

src/buildercore/core.py
```diff
@@ -287,6 +287,9 @@ def stack_all_ec2_nodes(stackname, workfn, username=config.DEPLOY_USER, concurre
     if node:
         nodes = {k: v for k, v in nodes.items() if v == int(node)}
         public_ips = {k: v for k, v in public_ips.items() if k in nodes.keys()}
+    elif instance_ids:
+        nodes = {k: v for k, v in nodes.items() if k in instance_ids}
+        public_ips = {k: v for k, v in public_ips.items() if k in nodes.keys()}
```
lsh-0:
```python
    if node:
        nodes = {k: v for k, v in nodes.items() if v == int(node)}
    elif instance_ids:
        nodes = {k: v for k, v in nodes.items() if k in instance_ids}
    public_ips = {k: v for k, v in public_ips.items() if k in nodes.keys()}
```

src/buildercore/lifecycle.py
```diff
@@ -111,23 +116,38 @@ def start(stackname):
...
+        wait_for_ec2_steady_state(stackname, ec2_to_be_checked)
+
     if rds_to_be_started:
         _wait_rds_all_in_state(stackname, 'available', rds_to_be_started)
 
     update_dns(stackname)
 
+def wait_for_ec2_steady_state(stackname, ec2_to_be_checked):
+    _wait_ec2_all_in_state(stackname, 'running', ec2_to_be_checked)
+    # TODO: this should only look at ec2_to_be_checked
```
lsh-0: comment still necessary?

lsh-0:
> Scary to work on this stuff when a test suite could turn off all the prod environments due to a bug.

with-great-power-comes-great-responsibility

two sets of credentials for builder devs with one for development that excludes everything on prod environment. 

we could also gate access to the stop/start/restart type commands in builder that checks how qualified our request is.

giorgiosironi:
See https://github.com/elifesciences/issues/issues/5000 for what came out of the post-mortem.

src/buildercore/core.py
giorgiosironi: Not sure if equivalent, because `public_ips` is not present for the missing `else` case but it's executed in the second version.

giorgiosironi:
All tests successful.

src/buildercore/core.py
lsh-0: you're right, ignore me

giorgiosironi:
Seems this feature was triggered in https://alfred.elifesciences.org/blue/organizations/jenkins/test-elife-bot/detail/test-elife-bot/761/pipeline but failed because `bioprotocol` is not accessible to `elife-alfred`?

giorgiosironi:
`end2end` can't work without all the stacks being turned on, so merging the fix.

giorgiosironi:
Not ideal as  `elife-alfred` will happily turn off and on again an instance it can't access. This is the error it sees, so that the restart could be narrowed down more:
```
16:14:50  INFO - buildercore.utils - waiting for nodes to complete boot
16:14:52  2019-08-06 15:14:52,479 botocore.vendored.requests.packages.urllib3.connectionpool [INFO] Starting new HTTPS connection (1): ec2.us-east-1.amazonaws.com
16:14:52  INFO - botocore.vendored.requests.packages.urllib3.connectionpool - Starting new HTTPS connection (1): ec2.us-east-1.amazonaws.com
16:14:52  INFO - buildercore.core - Executing on ec2 nodes ({'i-01fa171f9bbd8119b': '100.24.46.133'}), concurrency None
16:14:52  [100.24.46.133] Executing task 'single_node_work'
16:14:53  ERROR - buildercore.core - Needed to prompt for a connection or sudo password (host: 100.24.46.133), but input would be ambiguous in parallel mode
16:14:53  
16:14:53  INFO - buildercore.lifecycle - Generic failure of _daemons_ready execution: One or more hosts failed while executing task 'single_node_work'
```

lsh-0:
> end2end can't work without all the stacks being turned on, so merging the fix.

was this some bioprotocol fix that was merged? 

And isn't this a familiar problem when one machine hasn't seen the other before? I ran the below and got a prompt for input:

```
elife@prod--alfred:/ext/srv/builder$ sudo su jenkins
jenkins@prod--alfred:/ext/srv/builder$ ./bldr ssh:bioprotocol--end2end

...

Stack not running. Should it be started? (should_start)
(1, '-', True)
(2, '-', False)
> 1

...

[localhost] local: ssh elife@52.87.245.48 -i ~/.ssh/id_rsa
The authenticity of host '52.87.245.48 (52.87.245.48)' can't be established.
ECDSA key fingerprint is SHA256:xlplNeKfam9Auz1X3qgFP9R9v/o5R7KN72B1vTWCfbk.
Are you sure you want to continue connecting (yes/no)? yes
Warning: Permanently added '52.87.245.48' (ECDSA) to the list of known hosts.

```

lsh-0:
I stopped the instance and am running spectrum tests again here: https://alfred.elifesciences.org/job/test-elife-spectrum/352/console

and it looks like it could bring up all the end2end nodes.

giorgiosironi:
> was this some bioprotocol fix that was merged?

Just https://github.com/elifesciences/builder-configuration/pull/74 nothing new, was fixed since 15:24 UTC yesterday.

PR #524, opened 2019-07-22
Update AMI for 1804 to ami-0ddf1938f3461690b

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-builder-ami-update/7/display/redirect which resulted in this pull request.

PR #526, opened 2019-07-22
Update basebox--1604

giorgiosironi:
Spawned, manually, from https://github.com/elifesciences/builder/pull/524

Should eventually produce a 16.04 AMI if we want to still support it? Many projects are officially on 16.04 at the moment.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -303,7 +303,7 @@ basebox:
                 ami: ami-9eaa1cf6 # trusty
         1604:
             ec2:
-                ami: ami-1d4e7a66 # xenial, build 20170811, hvm:ebs-ssd
+                ami: ami-0cfee17793b08a293 # xenial, build 20190628, amd64, hvm:ebs-ssd, us-east-1
         1804:
             ec2:
                 # should match aws-alt 'fresh'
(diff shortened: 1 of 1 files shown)


lsh-0:
yes, lets support 16.04 for now. Once the last projects are off of 14.04, I'll start a new long-running ticket to upgrade the 16.04 projects. We'll have a new LTS in ~9 months ...

projects/elife.yaml
```diff
@@ -320,16 +320,13 @@ basebox:
         standalone1604:
             description: isolated from the master-server and uses Ubuntu 16.04 (Xenial)
             ec2:
-                ami: ami-1d4e7a66 # xenial, build 20170811, hvm:ebs-ssd
+                ami: ami-0cfee17793b08a293 # xenial, build 20190628, amd64, hvm:ebs-ssd, us-east-1
         # 18.04
-        standalone18.04:
+        standalone1804:
```
lsh-0: nice catch.

giorgiosironi:
We do support it but is it worth automating the 16.04 AMI creation as we do for 18.04?

lsh-0:
I think so. Was it particularly onerous to support automated 18.04? Can the two sets of changes be rolled into one PR? When 20.04 comes out some projects will migrate to that.

giorgiosironi:
I can just replicate the same scripts/pipelines and extract the common steps. I'll prefer two different PRs as they can fail independently and one shouldn't stop the other from running.

lsh-0:
no worries then. All these lists of checkboxes during the ec2 upgrades has me thinking about automation all day.

PR #529, opened 2019-08-07
Immediately fail boot if SSH access is denied

giorgiosironi:
Should solve https://github.com/elifesciences/builder/pull/523#issuecomment-518715341 but unsure if there are false positive when servers that are booting would temporarily refuse access in this way.

Tested by removing my key from `medium--ci:/home/ubuntu/.ssh/authorized_keys` which causes the `start` code to fail when trying to connect to check that boot has been completed.

### DIFF
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -326,10 +326,8 @@ def stack_all_ec2_nodes(stackname, workfn, username=config.DEPLOY_USER, concurre
                     raise err
             except config.FabricException as err:
                 LOG.error(str(err).replace("\n", "    "))
-                # not useful to specify more, this will be printed out
-                # but we are already logging it (and printing it on stderr)
-                # which is better
-                raise SystemExit("")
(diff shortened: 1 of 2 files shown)


giorgiosironi:
Very bad, https://alfred.elifesciences.org/blue/organizations/jenkins/test-elife-spectrum/detail/test-elife-spectrum/353/pipeline failed with exactly this error. Was probably temporary as `profiles--end2end` is surely configured to be accessible through SSH.

giorgiosironi:
From `fabric.network` it seems that either Fabric or the protocol itself cannot distinguish between these errors, so will need to transform this final exception into something that is retried again. _However_, I think we can be confident that if a timeout is reached with this exception, it is a real authentication problem rather than something that should be addressed with a reboot.

giorgiosironi:
Reverted in https://github.com/elifesciences/builder/commit/92aa6fd64d70d891dd3ceb5eeb04b32e8c3a5150

PR #531, opened 2019-08-13
Do not reboot on persistent SSH access problems

giorgiosironi:
New attempt for https://github.com/elifesciences/builder/pull/529 which got reverted.

Intentions:

- a login problem allows the polling to continue
- a generic SSH problem allows the polling to continue
- a persistent login problem doesn't cause a reboot
- a persistent generic SSH problem causes a reboot

Needs more test coverage for `buildercore.lifecycle`, I'll see what can be stubbed and covered there.

### DIFF
--- a/src/buildercore/lifecycle.py
+++ b/src/buildercore/lifecycle.py
@@ -165,7 +165,12 @@ def _some_node_is_not_ready(stackname, **kwargs):
         LOG.info("No running instances yet: %s", e)
         return True
     except config.FabricException as e:
-        LOG.info("Generic failure of _daemons_ready execution: %s", e)
+        # login problem is a legitimate error for booting servers,
+        # but also a signal the SSH private key is not allowed if it persists
+        if "Needed to prompt for a connection or sudo password" in e.message:
+            LOG.error("SSH access problem: %s", e)
+            return True
+        LOG.info("Generic failure of _some_node_is_not_ready execution: %s (class %s)", e, type(e))
         return True
     return False
 
--- a/src/buildercore/lifecycle.py
+++ b/src/buildercore/lifecycle.py
@@ -124,7 +124,7 @@ def start(stackname):
     try:
         wait_for_ec2_steady_state(stackname, ec2_to_be_checked)
     except EC2Timeout as e:
-        # a persistent login problem won't be solved be a reboot
+        # a persistent login problem won't be solved by a reboot
         if "Needed to prompt for a connection or sudo password" in e.message:
             raise
         # in case of botched boot and/or inability to
--- a/src/buildercore/lifecycle.py
+++ b/src/buildercore/lifecycle.py
@@ -125,7 +125,7 @@ def start(stackname):
         wait_for_ec2_steady_state(stackname, ec2_to_be_checked)
     except EC2Timeout as e:
         # a persistent login problem won't be solved by a reboot
-        if "Needed to prompt for a connection or sudo password" in e.message:
+        if "Needed to prompt for a connection or sudo password" in str(e):
             raise
         # in case of botched boot and/or inability to
         # access through SSH, try once to stop and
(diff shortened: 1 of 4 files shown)


src/buildercore/lifecycle.py
```diff
@@ -165,8 +167,14 @@ def _some_node_is_not_ready(stackname, **kwargs):
         LOG.info("No running instances yet: %s", e)
         return True
     except config.FabricException as e:
-        LOG.info("Generic failure of _daemons_ready execution: %s", e)
-        return True
+        # login problem is a legitimate error for booting servers,
+        # but also a signal the SSH private key is not allowed if it persists
+        if "Needed to prompt for a connection or sudo password" in e.message:
```
lsh-0: :+1:

src/tests/test_buildercore_utils.py
```diff
@@ -127,6 +127,36 @@ def test_call_while_timeout(self, sleep):
...
+            self.fail("Should not return normally")
+        except OSError as e:
+            self.assertEqual("Reached timeout 15 while waiting ...", e.message)
+
+    @patch('time.sleep')
+    def test_call_while_custom_message(self, sleep):
+        check = MagicMock()
+        check.return_value = True
+        try:
+            utils.call_while(check, interval=5, timeout=15, update_msg="waiting for Godot")
```
lsh-0: > Estragon: Nothing to be done.

giorgiosironi:
Some problem with the tests on Python 3.

giorgiosironi:
@GiancarloFusiello today I learned that `e.message` (where `e` is an exception) is not a field anymore on Python 3, so I'm using `str(e)`. Sounds alright?

lsh-0:
```python
Python 2.7.16 (default, Mar 11 2019, 18:59:25) 
[GCC 8.2.1 20181127] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> str(AssertionError("foo"))
'foo'
>>> str(RuntimeError("foo"))
'foo'
>>> RuntimeError("foo").message
'foo'
>>> AssertionError("foo").message
'foo'
```

looks ok to me. Not all exceptions/errors are structured the same though.

giorgiosironi:
`.use-python-3`, the correct flag is called `.use-python-3.flag` and is in `.gitignore`.

lsh-0:
The problem with Paramiko+Python3 has been figured out but is stuck in some sort of CI purgatory at the moment while the author figures out how to get it to pass. I hope that we can switch over to py3 for builder soon

src/buildercore/lifecycle.py
```diff
@@ -123,7 +123,10 @@ def start(stackname):
 
     try:
         wait_for_ec2_steady_state(stackname, ec2_to_be_checked)
-    except EC2Timeout:
+    except EC2Timeout as e:
```
GiancarloFusiello: Checking an error string doesn't seem like a good idea. Typically you would create an exception for the specific problem.

Also `e` is not very descriptive. Perhaps rename the variable to `error` or `exception`?

lsh-0: `except EC2Timeout as exception` ... really?

GiancarloFusiello: Yes. `exception`, in this case, is a variable with a name that describes the object it represents, which in this case, is an Exception class object. You could go one step further and provide a variable name which describes the type of exception. The point being that when someone reads the code further down the script, it's easy for the reader to know what that variable contains. Using single character variables doesn't provide this and is generally considered a bad practice.

giorgiosironi: Explained in person the error string, which stems from the fact that it's Fabric raising the exception without using more specific types that we can catch.

giorgiosironi: No opinions on the variable name, so by default leaving it consistent with the rest of the project and separating that decision if we want to make them more readable.

PR #533, opened 2019-08-19
Log error if bootstrap user key is not found

giorgiosironi:
Partially comments out https://github.com/elifesciences/builder/pull/503 to help https://github.com/elifesciences/api-dummy/pull/152

### DIFF
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -255,7 +255,8 @@ def _ec2_connection_params(stackname, username, **kwargs):
         if os.path.exists(pem):
             params['key_filename'] = pem
         else:
-            raise RuntimeError("private key for the bootstrap user for this host does not exist. I looked here: %s" % pem)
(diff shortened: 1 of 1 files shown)


src/buildercore/core.py
```diff
@@ -255,7 +255,8 @@ def _ec2_connection_params(stackname, username, **kwargs):
         if os.path.exists(pem):
             params['key_filename'] = pem
         else:
-            raise RuntimeError("private key for the bootstrap user for this host does not exist. I looked here: %s" % pem)
+            LOG.error("private key for the bootstrap user for this host does not exist. I looked here: %s" % pem)
```
giorgiosironi: though it is debatable whether this is actually an error, as the SSH key from `~` could be used instead and seems to have worked so far. I'd agree it's cleaner to use the `builder`-generated one for the boostrap user (`ubuntu`) used instead of `elife`.

PR #535, opened 2019-08-27
Feature: upgrades paramiko to remove cryptography warning

lsh-0:
Unfortunately this is not the 'make paramiko fast' patch we've all been waiting for, it just fixes all the deprecated cryptography warnings we were getting

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,7 @@ boto==2.48.0
 boto3==1.6.8
 botocore==1.9.8
 coverage==4.5.1
+paramiko==2.6.0
 Fabric3==1.14.post1
 green==2.12.1
 ipython==5.5.0


lsh-0:
Fabric3 (the unofficial Python3 fabric fork) has been deprecated now that Fabric supports Python3, however the major version bump of Fabric changes a lot.

requirements.txt
```diff
@@ -4,6 +4,7 @@ boto==2.48.0
 boto3==1.6.8
 botocore==1.9.8
 coverage==4.5.1
+paramiko==2.6.0
```
giorgiosironi: currently this is already `2.6.0` locally for me, I guess this forces upgrades for other installation?

giorgiosironi: would be wary of major versions but anything `2.6.*` should be pretty much ok

lsh-0: hmm, not sure. let me destroy my venv and re-install ... yeah, there it is. I'll close this PR then

PR #536, opened 2019-08-27
Feature: shifts default RDS version to 10.6

lsh-0:
* shifts default RDS version to 10.6
* pins individual projects at previous version (9.4)


lsh-0:
there are later versions of the 10.x line of psql, but xpub had already opted for 10.6 so I figured I'd go with that. I do believe this can be merged as-is without affecting anything, so here we go ...

PR #537, opened 2019-08-27
Add IP blacklisting to Fastly

thewilkybarkid:
https://github.com/elifesciences/issues/issues/5055

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -656,6 +656,7 @@ journal:
                     fallbacks:
                         4xx: "4xx.html"
                         5xx: "5xx.html"
+                ip-blacklist: true
                 vcl:
                     - "original-host"
                     - "ping-status"
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -552,7 +552,7 @@ journal:
                     fallbacks:
                         4xx: "4xx.html"
                         5xx: "5xx.html"
-                ip-blacklist: true
+                ip-blacklist: true # maintained at https://manage.fastly.com/configure/services/1WzbLJKoGqzOF1IrhTNx2M/versions/19/data/acls
                 vcl:
                     - "original-host"
                     - "ping-status"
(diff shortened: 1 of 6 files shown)


src/buildercore/terraform.py
```diff
@@ -11,7 +11,7 @@
 only_if_managed_services_are_present = only_if(*MANAGED_SERVICES)
 
 EMPTY_TEMPLATE = '{}'
-PROVIDER_FASTLY_VERSION = '0.4.0',
+PROVIDER_FASTLY_VERSION = '0.9.0',
```
thewilkybarkid: New `activate` option means there's _always_ a change to apply... 🤔

thewilkybarkid:
Deployed to `prod` and have added IP addresses through the UI.

src/buildercore/terraform.py
thewilkybarkid: Hmm, doesn't seem to always be true.

projects/elife.yaml
```diff
@@ -591,6 +592,7 @@ journal:
                     fallbacks:
                         4xx: "4xx.html"
                         5xx: "5xx.html"
+                ip-blacklist: true
```
giorgiosironi: could do `end2end` as well, just keeping everything in sync

projects/elife.yaml
```diff
@@ -654,6 +656,7 @@ journal:
...
```
giorgiosironi: since ip addresses lists are managed outside of builder/Terraform, this is the place to put a link to the Fastly panel where they are listed; not meant to be a long term solution

projects/elife.yaml
```diff
@@ -551,6 +552,7 @@ journal:
                     fallbacks:
                         4xx: "4xx.html"
                         5xx: "5xx.html"
+                ip-blacklist: true # maintained at https://manage.fastly.com/configure/services/1WzbLJKoGqzOF1IrhTNx2M/versions/19/data/acls
```
thewilkybarkid: These URIs contain version numbers, but the ACL content isn't part of the version so I think this will still be usable in the future (otherwise you'd have to switch to whatever the active version number is).

src/buildercore/terraform.py
thewilkybarkid: Think this just applies when a service hasn't been updated with this new version. Can't see it happening again after that.

PR #538, opened 2019-08-30
Feature: default rds version is now 10.9

lsh-0:
* latest RDS PostgreSQL version is 10.9
* elife-xpub now pinned at 10.6

The PPA will install the latest version of PostgreSQL 10.x which at time of writing is 10.10. 

If new machines with new RDS are going to be brought up, they should do so with a version that most closely matches what is available in Ubuntu/CI


lsh-0:
took me a while to find this: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts

I suspect AWS will only upgrade 10.x support for security fixes. 10.8 was skipped

PR #551, opened 2019-09-27
Feature: rds major version upgrade fixes

lsh-0:
* changes default rds psql version to 11.1 (from 10.6). aws will not let us upgrade psql 9.4 to 10.x
* fixes a bug where a project's instance-id was being used instead of the actual alt-config value
* adds `rds.major-version-upgrade` support to project config and CFN template generation

todo:
* add test to ensure ad-hoc alt-config changes appear in the upgrade diff


projects/elife.yaml
```diff
@@ -76,7 +76,8 @@ defaults:
             engine: postgres # or 'MySQL'
             # ensure this matches the version of Postgres you install on server!
             # version: '9.4' # EOL Feb 2020
-            version: '10.9' # EOL 2022
+            version: '11.1' # EOL Nov 2023
+            allow-major-version-upgrade: False # do not alter unless you're doing a major version upgrade
```
giorgiosironi: the naming of this is consistent with RDS terminology, which is what we should use :+1:

src/buildercore/cfngen.py
```diff
@@ -711,7 +711,12 @@ def regenerate_stack(stackname, **more_context):
     download_cloudformation_template(stackname)
     (pname, instance_id) = core.parse_stackname(stackname)
     more_context['stackname'] = stackname # TODO: purge this crap
-    more_context['alt-config'] = instance_id
+    # lsh@2019-09-27: usage of `instance_id` here is wrong. `instance_id` looks like "foobar" in "journal--foobar"
+    # and is only correct when an alt-config matches. We typically have alt-configs for our common environments, like
+    # ci, end2end, prod, continuumtest and has thus worked stably for a while now.
+    # ad-hoc instances whose instance-id does not match an environment will have it's alt-config ignored.
+    # the alt-config used during instance creation is found in `current_context` (but may not have always been the case).
+    more_context['alt-config'] = current_context.get('alt-config', instance_id)
```
giorgiosironi: can give an headache, just take some sample stacks that fall into the category of custom `instance_id`s and test that `update_infrastructure` doesn't get crazy and shows an empty diff

src/tests/fixtures/projects/dummy-project.yaml
```diff
@@ -609,6 +609,12 @@ project-with-rds-encryption:
             storage: 5
             encryption: arn:aws:kms:us-east-1:1234:key/12345678-1234-1234-1234-123456789012
 
+project-with-rds-major-version-upgrade:
```
giorgiosironi: I think this needs to be added to https://github.com/elifesciences/builder/blob/master/src/tests/test_buildercore_project.py#L6

lsh-0: yup, already done, just working on that other test with the adhoc alt-config definition

src/buildercore/cfngen.py
lsh-0: wow - I've really gone down the rabbit hole on this one.

Uncovered a few bugs, fixed a few things but I can't figure out why the dummy fixtures aren't including 'allow-major-version-upgrades' in the final project configuration. I have to stop it here for the day because I've spent too much time on it.

src/buildercore/cfngen.py
```diff
@@ -711,7 +715,14 @@ def regenerate_stack(stackname, **more_context):
...
     more_context['stackname'] = stackname # TODO: purge this crap
-    more_context['alt-config'] = instance_id
+    # lsh@2019-09-27: usage of `instance_id` here is wrong. `instance_id` looks like "foobar" in "journal--foobar"
+    # and is only correct when an alt-config matches. We typically have alt-configs for our common environments, like
+    # ci, end2end, prod, continuumtest and has thus worked stably for a while now.
+    # ad-hoc instances whose instance-id does not match an environment will have it's alt-config ignored.
+    # the alt-config used during instance creation is found in `current_context` (but may not have always been the case).
+    #more_context['alt-config'] = instance_id
+    #more_context['alt-config'] = current_context.get('alt-config', instance_id)
+    more_context['alt-config'] = current_context['alt-config']
```
lsh-0: This is what should ideally be happening, but I don't know how backwards compatible it is. It's such a minute and easily forgettable thing that I think we should set it to the correct value here and not think about again until if/when it breaks.

src/buildercore/trop.py
```diff
@@ -282,7 +282,7 @@ def render_rds(context, template):
         'BackupRetentionPeriod': lu('rds.backup-retention'),
         'DeletionPolicy': lu('rds.deletion-policy'),
         "Tags": tags,
-        "AllowMajorVersionUpgrade": False, # default? not specified.
+        "AllowMajorVersionUpgrade": lu('rds.allow-major-version-upgrade', default=False),
```
lsh-0: todo: problem here, I shouldn't have to be using `default=False`, there should already be a default of `False` in the project file. So why isn't `rds.allow-major-version-upgrade` available?

src/tests/test_cfn.py
```diff
@@ -38,6 +39,47 @@ def test_generate_stack_from_input(self):
...
+        current_context = context_handler.load_context(stackname)
+
+        #print('current context')
+        # pprint.pprint(current_context)
+
+        # skip calling update_infrastructure, we just want to test the diff with any changes
+        more_context = {
+            # todo: not working
+            #'rds': {
+            #    'allow-major-version-upgrade': True
```
lsh-0: todo: this should be possible but isn't. I thought it was because this 'extra' context is simply being `.update`d over the regular context, overwriting any struct at `rds` already, but it's not happening *at all*.

src/buildercore/cfngen.py
```diff
@@ -623,10 +623,14 @@ def _title_is_removable(title):
...
-        if not title in old_template[section]:
-            return False
+        if section in old_template:
+            # title was there before with a deprecated name, leave it alone
+            # e.g. 'EC2Instance' rather than 'EC2Instance1'
+            if not title in old_template[section]:
+                return False
+        else:
+            LOG.warn("section %r not present in old template but is present in new: %s" % (section, title))
+            return False # can we handle this better?
```
lsh-0: todo: this came up when in `tests.test_cfn.test_altconfig_name_preserved` where the template being generated was missing `Output`s or something. I haven't investigated very much but is probably just an artifact of testing using data in an unexpected way

lsh-0:
green here, but locally I'm getting five errors along the lines of:

> AssertionError: Unknown project 'project-with-eks'. Known projects: ['dummy-eu', 'project-with-fastly-shield-aws-region']

> AssertionError: Unknown project 'project-with-fastly-minimal'. Known projects: ['dummy-eu', 'project-with-fastly-shield-aws-region']

> AssertionError: Unknown project 'project-with-eks-helm'. Known projects: ['dummy-eu', 'project-with-fastly-shield-aws-region']

> AssertionError: Unknown project 'project-on-gcp'. Known projects: ['dummy-eu', 'project-with-fastly-shield-aws-region']

> AssertionError: Unknown project 'project-with-fastly-shield-pop'. Known projects: ['dummy-eu', 'project-with-fastly-shield-aws-region']

haven't investigated though

src/buildercore/trop.py
giorgiosironi: Default is missing in the fixtures projects file, I'll add it.

lsh-0: sure, I went a bit bonkers over this yesterday and just tidied up what I could this morning and then left it for Friday

giorgiosironi:
> green here, but locally I'm getting five errors along the lines of:

I think different tests execution orders triggered a bug in `test_buildercore_core`, tried to fix it in one of these commits, let me know if it improves.

lsh-0:
> tried to fix it in one of these commits, let me know if it improves.

Unfortunately not. 

I've pruned the scope of what is being tested down to just what I need and added an uncomitted fix I had for the `regenerate-fixtures.sh` script. I'll merge it if it's green.

PR #552, opened 2019-09-27
Remove hardcoding of project names

giorgiosironi:
Unfortunately it seems changing the security groups requires to destroy and recreate the whole cluster, so can't be done immediately.


giorgiosironi:
Merging as part of https://github.com/elifesciences/issues/issues/5195

PR #555, opened 2019-10-02
Switch test cluster to use elifesciences.org

giorgiosironi:
For https://github.com/elifesciences/issues/issues/4572

Required for wildcard HTTPS certificate to be used.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -2144,7 +2144,7 @@ kubernetes-aws:
                 desired-capacity: 3
             helm: true
             external-dns:
-                domain-filter: "elifesciences.net"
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Had some problems in applying due to timeouts,

giorgiosironi:
For posterity, after an interruption I had this problem despite repeatedly clearing `.cfn/terraform/kubernetes-aws--test/out.plan`:
```
Terraform doesn't allow you to run plans that were created from older
states since it doesn't properly represent the latest changes Terraform
may have made, and can result in unsafe behavior.

Plan Serial:    128
Current Serial: 129
```
Run a `terraform apply` manually from that folder that skips generating a plan file.

PR #563, opened 2019-10-09
Feature: switches RDS storage-type default from 'Standard' to 'gp2'

lsh-0:


lsh-0:
feel free to merge if you approve

lsh-0:
The RDS console page also has a 'Recommendations' section that lists switching to gp2. I really think magnetic storage is going the way of 32-bit processors.

PR #565, opened 2019-10-09
elife-bot-decision-letter-input bucket config.

gnott:
Re issue https://github.com/elifesciences/issues/issues/5146

I suppose this doesn't require lots of manual actions to implement, but please let me know if I can do more here.

Thanks!

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -759,6 +759,7 @@ elife-bot:
                 cors: true
             "{instance}-elife-bot-sessions":
             "{instance}-elife-bot-digests-input":
+            "{instance}-elife-bot-decision-letter-input":
         sns:
             # elife-bot-event-property--prod, elife-bot-event-property--end2end, etc.
             - elife-bot-event-property--{instance}
@@ -773,6 +774,8 @@ elife-bot:
                     deletion-policy: delete
                 "{instance}-elife-bot-digests-input":
                     deletion-policy: delete
+                "{instance}-elife-bot-decision-letter-input":
+                    deletion-policy: delete
                 # commented out because buckets already exist outside of CloudFormation:
                 #"{instance}-elife-bot":
                 #    deletion-policy: delete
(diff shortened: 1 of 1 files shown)


gnott:
I forgot to note that notifications from this bucket going to the same queue as the other two existing buckets is all fine for the `elife-bot` side of things. There's one queue listener, and then depending on the bucket name in the notification it can be shunted to the workflow desired.

projects/elife.yaml
```diff
@@ -773,6 +774,8 @@ elife-bot:
                     deletion-policy: delete
                 "{instance}-elife-bot-digests-input":
                     deletion-policy: delete
+                "{instance}-elife-bot-decision-letter-input":
```
giorgiosironi: I think we should remove the `s1604` section now that `elife-bot` runs on 16.04 by default; or transform it into `s1804`. @lsh-0?

giorgiosironi:
Applied to `end2end`.

giorgiosironi:
`elife-bot` has full S3 permissions so not need to add this bucket to https://console.aws.amazon.com/iam/home?region=us-east-1#/policies/arn:aws:iam::512686554592:policy/ElifeBotServices which already has `*`.

giorgiosironi:
Applied to `prod`.

```
$ aws s3 ls | grep decision-letter-input
2019-10-10 09:10:27 continuumtest-elife-bot-decision-letter-input
2019-10-10 09:05:35 end2end-elife-bot-decision-letter-input
2019-10-10 09:17:00 prod-elife-bot-decision-letter-input
```

giorgiosironi:
Will remove the unused configuration `s1604`.

gnott:
Very exciting - thanks @giorgiosironi. I look forward to attaching this bucket to the bot for a yet-to-build ingest workflow.

There's another bucket critical to the decision letter project, which I think is on the board - hint hint :)

projects/elife.yaml
lsh-0: Probably remove it for now, I'll re-introduce it on a per-project basis for the ticket to upgrade all 16.04 projects to 18.04.

PR #569, opened 2019-10-14
Feature: 14.04 config removal

lsh-0:
* removes 14.04 configuration for all projects, excluding elife-api legacy

todo:

- [ ] investigate basebox. why was it still using 14.04?
- [x] look for any project formulas still using a 14.04 config in their `Jenkinsfile` (none found)

cc @giorgiosironi

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -221,13 +221,7 @@ defaults:
             ec2:
                 ami: ami-14a1d06e # GENERATED created from basebox--1604
                                   # us-east-1, build date 20170811, hvm:ebs-ssd, AMI built on 20171215
-        # deprecated
-        1404:
-            description: uses Ubuntu 14.04 (Trusty)
-            ec2:
-                ami: ami-92002785 # created from a 14.04 basebox instance "basebox.2016-11-03"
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -292,24 +277,12 @@ builder:
 basebox:
```
giorgiosironi: `basebox` is the last one to go; as long as there were projects on 14.04 we needed to be able to generate a AMI for them. [Current automation](https://github.com/elifesciences/elife-alfred-formula/blob/master/jenkinsfiles/Jenkinsfile.basebox-1804) targets an `aws-alt` configuration though.

lsh-0: this is helpful, thank you.

either Graham will figure out how to decouple Lens and elife-api--legacy, or I'll finish upgrading that project to 18.04 and fold it into the stable of maintained projects. I'd prefer the former.

lsh-0: Graham got Lens figured out.

lsh-0:
this is blocking: https://github.com/elifesciences/builder-base-formula/pull/265

PR #571, opened 2019-10-15
Feature: fabric "tasks" replacement 

lsh-0:
todo:
- [x] remove the `task` decorator, it doesn't seem to be necessary
- [x] investigate all extant incantations of builder that are out there (alfred, master-server, etc)
    - [github search results 1](https://github.com/search?q=org%3Aelifesciences+%2Fsrv%2Fbuilder%2Fbldr&type=Code)
    - [github search results 2](https://github.com/search?q=org%3Aelifesciences+.%2Fbldr&type=Code)
    - `./bldr -l`
    - `BUILDER_NON_INTERACTIVE=1 /srv/builder/bldr`
    - `sh "${env.BUILDER_PATH}bldr cmd:${stackname},'cd ${remoteTestArtifactFolder}/..; rm -rf artifact.tar; tar -cf artifact.tar $remoteTestArtifactFolderBasename'"`
- [x] tests
- [x] review (self)
- [x] review (others)

### DIFF
--- a/src/taskrunner.py
+++ b/src/taskrunner.py
@@ -1,5 +1,5 @@
 import shlex
-import sys, os
+import sys, os, traceback
 import cfn, lifecycle, masterless, vault, aws, metrics, tasks, master, askmaster, buildvars, project, deploy
 from buildercore.utils import splitfilter
 from decorators import echo_output
@@ -203,17 +203,17 @@ def exec_task(task_str, task_map_list):
         return return_map
 
     except BaseException as e:
-        # TODO: stderr?
-        print('exception while executing task %r: %s' % (task_name, str(e)))
-        # TODO: print stacktrace
-        return_map['rc'] = 2 # I guess?
+        print('exception while executing task %r: %s\n' % (task_name, str(e)))
+        print(traceback.format_exc())
+        return_map['rc'] = 2 # arbitrary
         return return_map
 
 def exec_many(command_string, task_map_list):
     "splits a string up into multiple command strings and passes each to `exec_task`"
 
     # note: I just could not get the lexer to work here,
-    # too much python convenience-magic going on
+    # too much python convenience-magic going on.
+    # this section tokenises the whole input by whitespace, preserving singly quoted values
 
     task_list = []
     skipping = False
@@ -237,7 +237,7 @@ def main(arg_list):
     task_map_list = generate_task_list(show_debug_tasks)
 
     if not arg_list or arg_list[1:]:
-        print("builder should be called from the 'bldr' script") # or with a single double quoted argument string
+        print("`taskrunner.main` must be called from the ./bldr script")
         return 1
 
     command_string = arg_list[0].strip()
--- a/src/taskrunner.py
+++ b/src/taskrunner.py
@@ -88,7 +88,7 @@ DEBUG_TASK_LIST = [
 def mk_task_map(task, qualified=True):
     path = "%s.%s" % (task.__module__.split('.')[-1], task.__name__)
     unqualified_path = task.__name__
-    description = (task.__doc__ or '').replace('    ', '').replace('\n', ' ')[:60]
+    description = (task.__doc__ or '').strip().replace('\n', ' ')[:60]
     return {
         "name": path if qualified else unqualified_path,
         "path": path,
@@ -204,7 +204,7 @@ def main(arg_list):
 
     command_string = arg_list[0].strip()
 
-    if not command_string:
+    if not command_string or command_string in ["-l", "--list", "-h", "--help", "-?"]:
         for tm in task_map_list:
             path_len = len(tm['name'])
             max_path_len = 35
@@ -212,7 +212,8 @@ def main(arg_list):
             print("%s%s%s" % (tm['name'], ' ' * offset, tm['description']))
         return 0
 
-    # TODO: do we have cases where we rely on running multiple tasks in one go?
+    # note: we don't seem to have any cases where we rely on running multiple tasks
+    # todo: this is a naive split, there may be quoted whitespace in the commands
     task_list = command_string.split(' ')
     task_result_list = [exec_task(task_str, task_map_list) for task_str in task_list]
 
--- a/src/taskrunner.py
+++ b/src/taskrunner.py
@@ -113,8 +113,8 @@ def mk_task_map(task, qualified=True):
 def generate_task_list(show_debug_tasks=False):
     """returns a collated list of maps with task information.
 
-    [{"name": "fn", "fn": pathto.fn1, "description": "foo bar baz"}, ...]
-     {"name": "pathto.fn", "fn": pathto.fn2, "description": "foo bar baz"}, ...]"""
+    [{"name": "ssh", "fn": cfn.ssh, "description": "foobar baz"}, ...]
+     {"name": "cfn.deploy", "fn": cfn.deploy, "description": "bar barbar"}, ...]"""
 
     def to_list(task_list, qualified=True):
         return [mk_task_map(task, qualified) for task in task_list]
--- a/src/taskrunner.py
+++ b/src/taskrunner.py
@@ -1,4 +1,4 @@
-from shlex import shlex
+import shlex
 import sys, os
 import cfn, lifecycle, masterless, vault, aws, metrics, tasks, master, askmaster, buildvars, project, deploy
 from buildercore.utils import splitfilter
@@ -6,8 +6,13 @@ from buildercore.utils import splitfilter
 def ping():
     return "pong"
 
+def echo(msg, *args, **kwargs):
+    if args or kwargs:
+        return "received: %s with args: %s and kwargs: %s" % (msg, args, kwargs)
+    return "received: %s" % (msg,)
+
 UNQUALIFIED_TASK_LIST = [
-    ping,
+    ping, echo,
 
     cfn.destroy,
     cfn.ensure_destroyed,
@@ -124,7 +129,7 @@ def generate_task_list(show_debug_tasks=False):
 def parse_kv_pairs(text, item_sep=",", value_sep="="):
     """Parse key-value pairs from a shell-like text."""
     # initialize a lexer, in POSIX mode (to properly handle escaping)
-    lexer = shlex(text, posix=True)
+    lexer = shlex.shlex(text, posix=True)
     # set ',' as whitespace for the lexer
     # (the lexer will use this character to separate words)
     lexer.whitespace = item_sep
@@ -134,7 +139,7 @@ def parse_kv_pairs(text, item_sep=",", value_sep="="):
     # (if your option key or value contains any unquoted special character, you will need to add it here)
     # https://docs.python.org/2/library/shlex.html#shlex.shlex.wordchars
     lexer.wordchars += value_sep
-    lexer.wordchars += "-"
+    lexer.wordchars += "!@$%^&*()-;?/"
 
     # then we separate option keys and values to build the resulting dictionary
     # (maxsplit is required to make sure that '=' in value will not be a problem)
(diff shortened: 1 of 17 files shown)


src/taskrunner.py
```diff
@@ -0,0 +1,260 @@
...
+
+    except KeyboardInterrupt:
+        print('\nStopped.') # mimic fabric
+        return_map['rc'] = 1
+        return return_map
+
+    except BaseException as e:
+        # TODO: stderr?
+        print('exception while executing task %r: %s' % (task_name, str(e)))
+        # TODO: print stacktrace
```
lsh-0: do

src/taskrunner.py
```diff
@@ -0,0 +1,260 @@
...
+        return_map['rc'] = 0
+        return return_map
+
+    except KeyboardInterrupt:
+        print('\nStopped.') # mimic fabric
+        return_map['rc'] = 1
+        return return_map
+
+    except BaseException as e:
+        # TODO: stderr?
```
lsh-0: no. remove comment

src/taskrunner.py
```diff
@@ -0,0 +1,260 @@
...
+        print('exception while executing task %r: %s' % (task_name, str(e)))
+        # TODO: print stacktrace
+        return_map['rc'] = 2 # I guess?
```
lsh-0: "arbitrary"

src/taskrunner.py
```diff
@@ -0,0 +1,260 @@
...
+        return return_map
+
+def exec_many(command_string, task_map_list):
+    "splits a string up into multiple command strings and passes each to `exec_task`"
+
+    # note: I just could not get the lexer to work here,
+    # too much python convenience-magic going on
+
```
lsh-0: explain what the next chunk of code is doing

src/taskrunner.py
```diff
@@ -0,0 +1,260 @@
...
+    task_result_list = [exec_task(task_str, task_map_list) for task_str in task_list]
+    return task_result_list
+
+def main(arg_list):
+    show_debug_tasks = os.environ.get("BLDR_ROLE") == "admin"
+    task_map_list = generate_task_list(show_debug_tasks)
+
+    if not arg_list or arg_list[1:]:
+        print("builder should be called from the 'bldr' script") # or with a single double quoted argument string
```
lsh-0: "builder must be called from the `bldr` script"
remove comment.

src/tests/test_taskrunner.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def tearDown(self):
+        pass
+
+    def test_list_tasks(self):
+        "printing the list of commands can be invoked in many different ways"
+        list_tasks_invocations = [
+            "", # no args
+            "-l", "--list",
+            "", "-h", "--help", "-?"]
```
lsh-0: duplicate empty string

src/tests/test_taskrunner.py
```diff
@@ -0,0 +1,169 @@
...
+        result_map = tr.exec_task("echo:hello-world!@$%^&*()-;?/", task_list)
+        expected = {
+            'rc': 0,
+            'result': 'received: hello-world!@$%^&*()-;?/',
+            'task': 'echo',
+            'task_args': ['hello-world!@$%^&*()-;?/'],
+            'task_kwargs': {}}
+        self.assertEqual(expected, result_map)
+
+    def test_commend_short_circuited_by_comment(self):
```
lsh-0: typo

lsh-0:
found a bug 

`./bldr cmd:lax--continuumtest,'echo hello world'` should do what you expect it to

and not

```
[3.82.35.211] run: echo
[3.82.35.211] out: 
[3.82.35.211] out: 
Command not found: 'hello'
Command not found: 'world'
```

lsh-0:
it's hanging on the docker stuff for some reason

giorgiosironi:
(haven't had time to review this yet anyway, but pretty central so important to do so)

lsh-0:
> (haven't had time to review this yet anyway, but pretty central so important to do so)

It's fine, I don't think there is any urgency.

src/taskrunner.py
```diff
@@ -0,0 +1,241 @@
...
+def ping():
+    return "pong"
+
+@echo_output
+def echo(msg, *args, **kwargs):
+    if args or kwargs:
+        return "received: %s with args: %s and kwargs: %s" % (msg, args, kwargs)
+    return "received: %s" % (msg,)
+
+UNQUALIFIED_TASK_LIST = [
```
giorgiosironi: some documentation of these long lists will help placing a new command in the right place

src/taskrunner.py
```diff
@@ -0,0 +1,241 @@
...
+    """
+    Parse string list into list of tuples: command, args, kwargs, hosts, roles.
+
+    See sites/docs/usage/fab.rst, section on "per-task arguments" for details.
+    """
+    cmds = []
+    for cmd in arguments:
+        args = []
+        kwargs = {}
+        if ':' in cmd:
```
giorgiosironi: this is a tricky part that needs several unit tests. See https://github.com/elifesciences/elife-jenkins-workflow-libs/search?utf8=%E2%9C%93&q=bldr&type= for good examples of calls from Jenkins

src/taskrunner.py
```diff
@@ -0,0 +1,241 @@
...
+
+    # bash hands us an escaped string value via ./bldr
+    command_string = arg_list[0].strip()
+
+    if not command_string or command_string in ["-l", "--list", "-h", "--help", "-?"]:
+        print("Available commands:\n")
+        indent = 4
+        for tm in task_map_list:
+            path_len = len(tm['name'])
+            max_path_len = 35
```
giorgiosironi: arbitrary? Just for output, I guess so not important if that's the case

src/taskrunner.py
lsh-0: this was taken from fabric itself, it's what they were using to parse arguments.

When I realised bash was stripping the quotes and that there was no way to preserve them I replaced my implementation with theirs. The tests I had didn't change except to replicate the quote removal that bash was doing.

Commas need to be escaped, but this was the same under Fabric as well. It uses commas to separate commands.

I'll take a look at that link and see if any further tests need to be added

src/taskrunner.py
lsh-0: it's the length of the longest task name ('path') in the list of tasks. you take an initial indent, plus that magic number, plus a dynamic offset so that the description strings that come next are aligned. I'll make sure this is a calculated value.

src/taskrunner.py
lsh-0: The cases in the link should all be handled provided the values of the variables don't contain commas - which was true before as well, we just may not have realised it.

lsh-0:
nuts, I committed the changes to the wrong branch: https://github.com/elifesciences/builder/pull/576/files

I think rather than undoing commits I'll leave the changes in that branch (which is branched from this one) and let it all get merged at once.

lsh-0:
we may also have been assuming that certain values being passed to tasks as parameters were being coerced into their correct types rather than just being left as strings. 

For example, the `cmd` function [here](https://github.com/elifesciences/builder/blob/master/src/cfn.py#L341) accepts a parameter `clean_output`, default `False` (boolean).

`./bldr cmd:lax--continuumtest,'echo hello\, world',clean_output=True`

behaves the same as

`./bldr cmd:lax--continuumtest,'echo hello\, world',clean_output=False`

but differently to

`./bldr cmd:lax--continuumtest,'echo hello\, world'`

with this PR we'll have control over the parsing of the command and can introduce type coercion if we want, but not right now.

src/tests/test_taskrunner.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def tearDown(self):
+        pass
+
+    def test_list_tasks(self):
+        "printing the list of commands can be invoked in many different ways"
+        list_tasks_invocations = [
+            "", # no args
+            "-l", "--list",
+            "-h", "--help", "-?"]
```
giorgiosironi: `-h` does not list the tasks currently:
```
$ ./bldr -h
Usage: fab [options] <command>[:arg1,arg2=val2,host=foo,hosts='h1;h2',...] ...

Options:
  -h, --help            show this help message and exit
  -d NAME, --display=NAME
...
```
I'm not aware of us using this option in any script.

src/tests/test_taskrunner.py
```diff
@@ -0,0 +1,123 @@
...
+            'task_args': ['zulu'],
+            'task_kwargs': {}}
+        self.assertEqual(expected, result_map)
+
+    def test_commands_with_args_and_kwargs_can_be_called(self):
+        task_list = tr.generate_task_list()
+        result_map = tr.exec_task("echo:zulu,gamma,a=b", task_list)
+        expected = {
+            'rc': 0,
+            'result': "received: zulu with args: ('gamma',) and kwargs: {'a': 'b'}",
```
giorgiosironi: `zulu` is also an arg here, so would change to `received: zulu with additional args: ...`

src/tests/test_taskrunner.py
```diff
@@ -0,0 +1,123 @@
...
+        expected = {
+            'task': 'echo',
+            'task_args': ['msg', 'echo hello world'],
+            'task_kwargs': {},
+            'result': "received: msg with args: ('echo hello world',) and kwargs: {}",
+            'rc': SUCCESS_RC}
+        self.assertEqual(expected, result_map)
+
+    def test_commands_with_special_chars_can_be_called_without_quoting(self):
+        task_list = tr.generate_task_list()
```
giorgiosironi: can you add a `this is what we give bash` comment here, too?

src/tests/test_taskrunner.py
```diff
@@ -0,0 +1,123 @@
...
+            'result': 'received: hello-world!@$%^&*()-;?/',
+            'task': 'echo',
+            'task_args': ['hello-world!@$%^&*()-;?/'],
+            'task_kwargs': {}}
+        self.assertEqual(expected, result_map)
+
+    def test_quoted_commands_can_be_called(self):
+        task_list = tr.generate_task_list()
+        # result_map = tr.exec_task("echo:'hello\, world'", task_list) # this is what we give bash
+        result_map = tr.exec_task("echo:hello\\, world", task_list) # this is what we see after bash
```
giorgiosironi: the pair of what we give bash and what we see doesn't check out for me; a single quoted `'hello\, world'` should preserve a single slash

src/taskrunner.py
```diff
@@ -0,0 +1,241 @@
...
+        "name": path if qualified else unqualified_path,
+        "path": path,
+        "description": description,
+        "fn": task,
+    }
+
+def generate_task_list(show_debug_tasks=False):
+    """returns a collated list of maps with task information.
+
+    [{"name": "fn", "fn": pathto.fn1, "description": "foo bar baz"}, ...]
```
giorgiosironi: example could be more realistic, using names like `ssh` or `deploy`

src/tests/test_taskrunner.py
lsh-0: > -h does not list the tasks currently:

neither does `--help`, I'll remove them both

src/tests/test_taskrunner.py
lsh-0: the `echo` task takes one required parameter, the message and zero or more args and kwargs. It's output is just the message, and if args or kwargs have been provided, those too.

in this case `zulu` is the message, `gamma` an arg and `a=b` a kwarg.

It could definitely be clearer though, I'll make it so

src/tests/test_taskrunner.py
lsh-0: sure thing

src/tests/test_taskrunner.py
lsh-0: you're right, test updated

lsh-0:
(deep breath)

\>**merge**<

PR #573, opened 2019-10-18
Add EFS IAM policy to Kubernetes workers

giorgiosironi:
Allows EFS usage, to share volumes between containers.

### DIFF
--- a/src/buildercore/terraform.py
+++ b/src/buildercore/terraform.py
@@ -808,6 +808,32 @@ def _render_eks_workers_role(context, template):
             'role': "${aws_iam_role.worker.name}",
         })
 
+    if context['eks']['efs']:
+        template.populate_resource('aws_iam_policy', 'kubernetes_efs', block={
+            'name': '%s--AmazonEFSKubernetes' % context['stackname'],
+            'path': '/',
+            'description': 'Allows management of EFS resources',
+            'policy': json.dumps({
+                "Version": "2012-10-17",
+                "Statement": [
+                    {
+                        "Effect": "Allow",
+                        "Action": [
+                            "elasticfilesystem:*",
+                        ],
+                        "Resource": [
+                            "*",
+                        ],
+                    },
+                ],
+            }),
+        })
+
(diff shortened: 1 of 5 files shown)


src/buildercore/terraform.py
```diff
@@ -808,6 +808,32 @@ def _render_eks_workers_role(context, template):
...
+            'name': '%s--AmazonEFSKubernetes' % context['stackname'],
+            'path': '/',
+            'description': 'Allows management of EFS resources',
+            'policy': json.dumps({
+                "Version": "2012-10-17",
+                "Statement": [
+                    {
+                        "Effect": "Allow",
+                        "Action": [
+                            "elasticfilesystem:*",
```
giorgiosironi: we can possibly be more specific here, what `Action`s are required? `Resource` is limited anyway to the AWS account and nothing else uses EFS, so not much to do on that field

giorgiosironi:
Already applied to `kubernetes-aws--test`.

src/buildercore/terraform.py
tayowonibi: from https://docs.aws.amazon.com/efs/latest/ug/efs-api-permissions-ref.html, i think these permissions below should be ok

elasticfilesystem:DescribeFileSystems 
elasticfilesystem:DescribeMountTargets
elasticfilesystem:DescribeMountTargetSecurityGroups
elasticfilesystem:DescribeTags

giorgiosironi:
Would like to know if we can assign fewer permissions than `elasticfilesystem:*` (only selected actions)? But not a blocker.

tayowonibi:
> Would like to know if we can assign fewer permissions than `elasticfilesystem:*` (only selected actions)? But not a blocker.


https://github.com/elifesciences/builder/pull/573#discussion_r336468558

src/buildercore/terraform.py
tayowonibi: yes

giorgiosironi:
Updated and reapplied.

PR #575, opened 2019-10-21
Use CloudFormation Outputs to discover the load balancer name

giorgiosironi:
For https://github.com/elifesciences/issues/issues/5137

The `ElasticLoadBalancer` name is generated by AWS, e.g.
```
Load balancer: elife-xpu-ElasticL-19CB72BN8E36S
DNS name: elife-xpu-ElasticL-19CB72BN8E36S-1360186739.us-east-1.elb.amazonaws.com
```

In order to perform blue-green deployment, we need to detach and re-attach instances from the load balancer. So we need to find out its name first, given an input like `elife-xpub--staging`.

Using the CloudFormation API when creating the infrastructure, we can add the generated name as an `Output` of the template. We then retrieve it at deployment time to use it.

Before this, we were listing all the load balancers and filtering them one by one checking their `Cluster` tag matched the input `elife-xpub--staging`. The ELB API we were using has a limit of 20 load balancers however, so it broke once we had too many.

Note: this requires `update_infrastructure` to be run on all ELB-related stacks to add the `Output`, otherwise deployments will still be failing because of the missing `Outputs` key.

### DIFF
--- a/src/buildercore/bootstrap.py
+++ b/src/buildercore/bootstrap.py
@@ -336,6 +336,7 @@ def master_data(region):
 def master(region, key):
     return master_data(region)[key]
 
+# TODO: move to buildercore.cloudformation?
 @core.requires_active_stack
 def template_info(stackname):
     "returns some useful information about the given stackname as a map"
--- a/src/buildercore/cloudformation.py
+++ b/src/buildercore/cloudformation.py
@@ -131,6 +131,7 @@ def _wait_until_in_progress(stackname):
 def read_template(stackname):
     "returns the contents of a cloudformation template as a python data structure"
     output_fname = os.path.join(config.STACK_DIR, stackname + ".json")
+    # TODO: use a context manager to close the file afterwards
     return json.load(open(output_fname, 'r'))
 
 def read_output(stackname, key):
@@ -169,6 +170,7 @@ def _merge_delta(stackname, delta):
 def write_template(stackname, contents):
     "writes a json version of the python cloudformation template to the stacks directory"
     output_fname = os.path.join(config.STACK_DIR, stackname + ".json")
+    # TODO: use a context manager to close the file afterwards
     open(output_fname, 'w').write(contents)
     return output_fname
 
--- a/src/buildercore/cloudformation.py
+++ b/src/buildercore/cloudformation.py
@@ -135,8 +135,10 @@ def read_template(stackname):
 
 def read_output(stackname, key):
     data = core.describe_stack(stackname).meta.data # boto3
+    ensure('Outputs' in data, "Outputs missing: %s" % data) 
     selected_outputs = [o for o in data['Outputs'] if o['OutputKey'] == key]
     ensure(len(selected_outputs) == 1, "Too many outputs selected: %s" % selected_outputs)
+    ensure('OutputValue' in selected_outputs[0], "Badly formed Output: %s" % selected_outputs[0])
     return selected_outputs[0]['OutputValue']
 
 def apply_delta(template, delta):
--- a/src/buildercore/cloudformation.py
+++ b/src/buildercore/cloudformation.py
@@ -133,6 +133,12 @@ def read_template(stackname):
     output_fname = os.path.join(config.STACK_DIR, stackname + ".json")
     return json.load(open(output_fname, 'r'))
 
+def read_output(stackname, key):
+    data = core.describe_stack(stackname).meta.data # boto3
+    selected_outputs = [o for o in data['Outputs'] if o['OutputKey'] == key]
+    ensure(len(selected_outputs) == 1, "Too many outputs selected: %s" % selected_outputs)
+    return selected_outputs[0]['OutputValue']
+
 def apply_delta(template, delta):
     for component in delta.plus:
         ensure(component in ["Resources", "Outputs", "Parameters"], "Template component %s not recognized" % component)
--- a/src/buildercore/cloudformation.py
+++ b/src/buildercore/cloudformation.py
@@ -136,7 +136,7 @@ def read_template(stackname):
 
 def read_output(stackname, key):
     data = core.describe_stack(stackname).meta.data # boto3
-    ensure('Outputs' in data, "Outputs missing: %s" % data) 
+    ensure('Outputs' in data, "Outputs missing: %s" % data)
     selected_outputs = [o for o in data['Outputs'] if o['OutputKey'] == key]
     ensure(len(selected_outputs) == 1, "Too many outputs selected: %s" % selected_outputs)
     ensure('OutputValue' in selected_outputs[0], "Badly formed Output: %s" % selected_outputs[0])
--- a/src/buildercore/concurrency.py
+++ b/src/buildercore/concurrency.py
@@ -21,4 +21,4 @@ def concurrency_for(stackname, concurrency_name):
     if concurrency_name is None:
         return 'parallel'
 
-    raise ValueError("Concurrency %s is not supported. Supported models: %s" % concurrency_name, concurrency_names)
+    raise ValueError("Concurrency %s is not supported. Supported models: %s" % (concurrency_name, concurrency_names))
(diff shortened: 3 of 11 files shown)


giorgiosironi:
Debugging tasks sample:
```
(venv) [12:57:25][giorgio@Newton:~/code/builder]$ bldr load_balancer_status:elife-xpub--staging
INFO - deploy - Load balancer name: elife-xpu-ElasticL-19CB72BN8E36S
INFO - deploy - Health: [{u'Description': 'N/A',
  u'InstanceId': 'i-08f0483b606b29d6c',
  u'ReasonCode': 'N/A',
  u'State': 'InService'},
 {u'Description': 'N/A',
  u'InstanceId': 'i-080c35729cf0e242d',
  u'ReasonCode': 'N/A',
  u'State': 'InService'}]
(venv) [12:59:51][giorgio@Newton:~/code/builder]$ bldr load_balancer_register_all:elife-xpub--staging
INFO - deploy - Load balancer name: elife-xpu-ElasticL-19CB72BN8E36S
INFO - deploy - Register all: {'nodes': {'i-080c35729cf0e242d': 1, 'i-08f0483b606b29d6c': 2},
 'public_ips': {'i-080c35729cf0e242d': '54.91.166.142',
                'i-08f0483b606b29d6c': '18.207.1.66'},
 'stackname': 'elife-xpub--staging',
 'user': 'elife'}
INFO - buildercore.bluegreen - Registering on elife-xpu-ElasticL-19CB72BN8E36S: ['i-08f0483b606b29d6c', 'i-080c35729cf0e242d']
INFO - buildercore.bluegreen - Waiting for registration of all on elife-xpu-ElasticL-19CB72BN8E36S: ['i-08f0483b606b29d6c', 'i-080c35729cf0e242d']
INFO - buildercore.bluegreen - InService: {'i-08f0483b606b29d6c': True, 'i-080c35729cf0e242d': True}
```

giorgiosironi:
@GiancarloFusiello can you sanity check the Python? The AWS part is complicated so no worries on that.

giorgiosironi:
```
bldr switch_revision_update_instance:elife-xpub--staging,approved,concurrency=blue-green
```
seems to work as expected:
Screenshot_2019-10-21 Load Balancers EC2 Management Console

src/buildercore/cloudformation.py
```diff
@@ -133,6 +133,12 @@ def read_template(stackname):
     output_fname = os.path.join(config.STACK_DIR, stackname + ".json")
     return json.load(open(output_fname, 'r'))
 
+def read_output(stackname, key):
+    data = core.describe_stack(stackname).meta.data # boto3
+    selected_outputs = [o for o in data['Outputs'] if o['OutputKey'] == key]
+    ensure(len(selected_outputs) == 1, "Too many outputs selected: %s" % selected_outputs)
```
GiancarloFusiello: Could be replaced with `assert` unless `ensure` has some error logic?

src/buildercore/cloudformation.py
```diff
@@ -133,6 +133,12 @@ def read_template(stackname):
     output_fname = os.path.join(config.STACK_DIR, stackname + ".json")
     return json.load(open(output_fname, 'r'))
 
+def read_output(stackname, key):
+    data = core.describe_stack(stackname).meta.data # boto3
+    selected_outputs = [o for o in data['Outputs'] if o['OutputKey'] == key]
```
GiancarloFusiello: could raise a `ValueError` if these keys do not exist. Ok if this is desired but thought I would check.

giorgiosironi: Yes, it's a low layer in the application so better make it robust

src/buildercore/concurrency.py
```diff
@@ -19,4 +21,4 @@ def concurrency_for(stackname, concurrency_name):
     if concurrency_name is None:
         return 'parallel'
 
-    raise ValueError("Concurrency %s is not supported" % concurrency_name)
+    raise ValueError("Concurrency %s is not supported. Supported models: %s" % concurrency_name, concurrency_names)
```
GiancarloFusiello: Needs parenthesis surrounding string formatting arguments.

giorgiosironi: Triggered the error and confirmed it, will fix

src/buildercore/core.py
```diff
@@ -279,6 +279,23 @@ def stack_conn(stackname, username=config.DEPLOY_USER, node=None, **kwargs):
 class NoPublicIps(Exception):
     pass
 
+def all_node_params(stackname):
+    data = stack_data(stackname)
+    public_ips = {ec2['InstanceId']: ec2.get('PublicIpAddress') for ec2 in data}
+    nodes = {ec2['InstanceId']: int(tags2dict(ec2['Tags'])['Node']) if 'Node' in tags2dict(ec2['Tags']) else 1 for ec2 in data}
```
GiancarloFusiello: Not sure about this line. Might need a comment for context or splitting over multiple lines for readability.

giorgiosironi:
@lsh-0 feel free to leave other comments, unblocking deployments.

src/buildercore/cloudformation.py
```diff
@@ -131,8 +131,17 @@ def is_updating(stackname):
 def read_template(stackname):
     "returns the contents of a cloudformation template as a python data structure"
     output_fname = os.path.join(config.STACK_DIR, stackname + ".json")
+    # TODO: use a context manager to close the file afterwards
```
lsh-0: should have just done it

src/buildercore/bootstrap.py
```diff
@@ -336,6 +336,7 @@ def master_data(region):
 def master(region, key):
     return master_data(region)[key]
 
+# TODO: move to buildercore.cloudformation?
```
lsh-0: definitely.

src/buildercore/cloudformation.py
lsh-0: not fear, assert is actually stripped out under some conditions.

`assert` is meant more of a developer sanity check (which is why they may be stripped out), however the way we use them here is more of a formal check: we don't want the normal flow of execution to proceed if our assertion about the environment is false in development, or while testing or in production or in a compiled format etc.

src/buildercore/cloudformation.py
```diff
@@ -161,6 +170,7 @@ def _merge_delta(stackname, delta):
 def write_template(stackname, contents):
     "writes a json version of the python cloudformation template to the stacks directory"
     output_fname = os.path.join(config.STACK_DIR, stackname + ".json")
+    # TODO: use a context manager to close the file afterwards
```
lsh-0: just do it in future

src/buildercore/core.py
```diff
@@ -287,6 +309,7 @@ def stack_all_ec2_nodes(stackname, workfn, username=config.DEPLOY_USER, concurre
         workfn, work_kwargs = workfn
 
     data = stack_data(stackname)
+    # TODO: reuse all_node_params?
```
lsh-0: perhaps `all_node_params` should have reused this? or this should have been extended to accommodate whatever usecase `all_node_params` needs. This is how a codebase grows gnarly.

src/deploy.py
```diff
@@ -1,8 +1,12 @@
 """module concerns itself with tasks involving branch deployments of projects."""
 
+from pprint import pformat
 from fabric.api import task
-from decorators import requires_aws_stack
-from buildercore import bootstrap
+from decorators import requires_aws_stack, debugtask
+# TODO: import modules rather than functions
```
lsh-0: I don't care, so long as it's explicit where the module or function came from. No `from blah import *`

src/tests/test_buildercore_trop.py
```diff
@@ -229,6 +229,8 @@ def test_clustered_template(self):
             resources['EC2Instance1']['Properties']['Tags']
         )
         outputs = data['Outputs']
+        self.assertIn('ElasticLoadBalancer', list(outputs.keys()))
```
lsh-0: does this work without the `list` wrapper? curious more than anything else

giorgiosironi: I think it was a pattern to make the error message clearer by dumping a `list` rather than a `dict_keys` object.

Will follow-up to the comments with a separate PR this week.

lsh-0: ah that makes sense

PR #577, opened 2019-10-28
peerscout, removes project configuration

lsh-0:


lsh-0:
> But first destroy the existing stacks

already done

lsh-0:
stuck on the docker stuff here as well :( grumble. I'll take a crack at debugging what is going on with them today.

PR #586, opened 2019-11-19
Feature: threadbare integration

lsh-0:
* adds a copy of threadbare
* adds ability to switch backend between fabric or threadbare
* wraps several Fabric commands to unify behaviour between the two libraries
* removes xdist for testing. It's asyncio usage was causing obscure errors with geventlet/multiprocessing

---

- [x] discrepancy between what threadbare `execute` is returning and what Fabric is returning (1)
- [x] `env` is being accessed as an object somewhere (see exception in results returned) (1)
- [x] `OSError` attempting to make a tempfile with threadbare while downloading. (2)
- [x] discrepancy between what threadbare `local` is returning when `capture=True` and what Fabric is returning (3)
- [x] `remote_list_files` is relying on `stdout` being returned instead of a dictionary (4)
- [x] `host` vs `host_string` confusion. investigation (5)
- [x] self review
- [ ] can `pytest-xdist` be re-introduced?

<details>
  <summary>outdated</summary>
---
(1)

```python
stackname = 'project-with-cluster-integration-tests--luke-20191204065616-172945'
kwargs = {'instance_ids': ['i-02ff6edd0df1cb414', 'i-0022ed6fe65bbe870']}
ip_to_ready = [AttributeError("'LockableDict' object has no attribute 'host'",), AttributeError("'LockableDict' object has no attribute 'host'",)]

    def _some_node_is_not_ready(stackname, **kwargs):
        try:
            ip_to_ready = stack_all_ec2_nodes(stackname, _daemons_ready, username=config.BOOTSTRAP_USER, **kwargs)
            LOG.info("_some_node_is_not_ready: %s", ip_to_ready)
>           return len(ip_to_ready) == 0 or False in ip_to_ready.values()
E           AttributeError: 'list' object has no attribute 'values'
```
---
(2)

```python
name = '/venv', mode = 511

    def makedirs(name, mode=0777):
        """makedirs(path [, mode=0777])
    
        Super-mkdir; create a leaf directory and all intermediate ones.
        Works like mkdir, except that any intermediate path segment (not
        just the rightmost) will be created if it does not exist.  This is
        recursive.
    
        """
        head, tail = path.split(name)
        if not tail:
            head, tail = path.split(head)
        if head and tail and not path.exists(head):
            try:
                makedirs(head, mode)
            except OSError, e:
                # be happy if someone already created the path
                if e.errno != errno.EEXIST:
                    raise
            if tail == curdir:           # xxx/newdir/. exists if xxx/newdir exists
                return
>       mkdir(name, mode)
E       OSError: [Errno 13] Permission denied: '/venv'
```
tracepath:

```python
src/decorators.py:80: in call
    return func(stackname, *args, **kwargs)
src/cfn.py:297: in download_file
    _download(path, destination)
venv/lib/python2.7/site-packages/backoff/_sync.py:99: in retry
    ret = target(*args, **kwargs)
src/cfn.py:295: in _download
    download(path, destination, use_sudo=True)
src/buildercore/threadbare/operations.py:438: in download
    local_path = _download_as_root_hack(remote_path, local_path, **kwargs)
src/buildercore/threadbare/operations.py:414: in _download_as_root_hack
    client.copy_remote_file(remote_tempfile, local_path)
venv/lib/python2.7/site-packages/pssh/clients/native/single.py:636: in copy_remote_file
    self._make_local_dir(destination)
venv/lib/python2.7/site-packages/pssh/clients/native/single.py:862: in _make_local_dir
    os.makedirs(dirpath)
```

---
(3)
```python
url = 'https://github.com/elifesciences/builder'

    def _git_remote_refs(url):
        cmd = "git ls-remote --heads %s" % url
        output = local(cmd, capture=True)
>       return [line.split() for line in output.splitlines()]
E       AttributeError: 'dict' object has no attribute 'splitlines'
```

---
(4)
```python
__________________________________________________ TestWithInstance.test_core_listfiles_remote ___________________________________________________

self = <integration_tests.test_with_instance.TestWithInstance testMethod=test_core_listfiles_remote>

    def test_core_listfiles_remote(self):
        with core.stack_conn(self.stackname, username=BOOTSTRAP_USER):
>           results = command.remote_listfiles('/')

src/integration_tests/test_with_instance.py:81: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

path = '/*', use_sudo = False

    def remote_listfiles(path=None, use_sudo=False):
        """returns a list of files in a directory at `path` as absolute paths"""
        if not path:
            raise AssertionError("path to remote directory required")
        with fab_api.hide('output'):
            runfn = remote_sudo if use_sudo else remote
            path = "%s/*" % path.rstrip("/")
            stdout = runfn("for i in %s; do echo $i; done" % path)
            if stdout == path: # some kind of bash artifact where it returns `/path/*` when no matches
                return []
>           return stdout.splitlines()
E           AttributeError: 'dict' object has no attribute 'splitlines'

src/buildercore/command.py:149: AttributeError
```
---
(5)
```python
___________________________________________________ TestWithManyNodes.test_restart_one_stopped ___________________________________________________

self = <integration_tests.test_with_many_nodes.TestWithManyNodes testMethod=test_restart_one_stopped>

    def test_restart_one_stopped(self):
        "multiple nodes can be restarted from a mixed state"
        node1 = core.find_ec2_instances(self.stackname)[0]
        node1.stop()
        node1.wait_until_stopped()
>       history = lifecycle.restart(self.stackname)

src/integration_tests/test_with_many_nodes.py:69: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/buildercore/lifecycle.py:78: in restart
    done_msg="all nodes have public ips, are reachable via SSH and have completed boot"
src/buildercore/utils.py:175: in call_while
    result = fn()
src/buildercore/lifecycle.py:74: in <lambda>
    lambda: _some_node_is_not_ready(stackname, node=node_id, concurrency='serial'),
src/buildercore/lifecycle.py:156: in _some_node_is_not_ready
    ip_to_ready = stack_all_ec2_nodes(stackname, _daemons_ready, username=config.BOOTSTRAP_USER, **kwargs)
src/buildercore/core.py:372: in stack_all_ec2_nodes
    return serial_work(single_node_work, params)
src/buildercore/core.py:384: in serial_work
    return execute(serial(single_node_work), hosts=list(params['public_ips'].values()))
src/buildercore/threadbare/execute.py:208: in execute_with_hosts
    results = execute(env, func, param_key="host_string", param_values=host_list)
src/buildercore/threadbare/execute.py:192: in execute
    return _serial_execution(env, func, param_key, param_values)
src/buildercore/threadbare/execute.py:139: in _serial_execution
    result_list.append(func())
src/buildercore/threadbare/execute.py:14: in inner
    return func(*args, **kwargs)
src/buildercore/core.py:345: in single_node_work
    return workfn(**work_kwargs)
src/buildercore/lifecycle.py:270: in _daemons_ready
    node_id = current_ec2_node_id()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def current_ec2_node_id():
        """Assumes it is called inside the 'workfn' of a 'stack_all_ec2_nodes'.
    
        Sticking to the 'node' terminology because 'instance' is too overloaded.
    
        Sample value: 'i-0553487b4b6916bc9'"""
    
>       ensure(env['host'] is not None, "This is supposed to be called with settings for connecting to an EC2 instance")
E       KeyError: 'host'

src/buildercore/core.py:397: KeyError
```
</details>

### DIFF
--- a/Dockerfile.py3
+++ b/Dockerfile.py3
@@ -1,9 +1,13 @@
 FROM python:3.5-alpine3.8
 
+# cmake + zlib-dev for parallel-ssh dependencies
+# paramiko is pure python and never needed it
 RUN apk add --no-cache \
     libressl-dev \
     musl-dev \
-    libffi-dev
+    libffi-dev \
+    cmake \
+    zlib-dev
 
 COPY requirements.txt /requirements.txt
 
@@ -17,6 +21,7 @@ RUN apk add --no-cache --virtual build-deps \
     pip install virtualenv && \
     mkdir /venv && \
     virtualenv --python=python3 /venv && \
+    /venv/bin/pip install parallel-ssh && \
     /venv/bin/pip install -r /requirements.txt && \
     apk del build-deps
 
--- a/Dockerfile.py3
+++ b/Dockerfile.py3
@@ -21,7 +21,6 @@ RUN apk add --no-cache --virtual build-deps \
     pip install virtualenv && \
     mkdir /venv && \
     virtualenv --python=python3 /venv && \
-    /venv/bin/pip install parallel-ssh && \
     /venv/bin/pip install -r /requirements.txt && \
     apk del build-deps
 
--- a/py2-requirements.txt
+++ b/py2-requirements.txt
@@ -22,4 +22,4 @@ PyYAML==4.2b2
 requests==2.20.0
 troposphere==2.3.3
 unittest2==1.1.0
-parallel-ssh
+parallel-ssh==1.9.1
--- a/py2-requirements.txt
+++ b/py2-requirements.txt
@@ -15,7 +15,6 @@ pylint==1.8.2
 pytest==3.5.0
 pytest-cov==2.5.1
 pytest-forked==0.2
-pytest-xdist==1.22.2
 python-dateutil==2.6.0
 python-slugify==1.2.1
 python-terraform==0.10.0
--- a/py2-requirements.txt
+++ b/py2-requirements.txt
@@ -24,3 +24,4 @@ PyYAML==4.2b2
 requests==2.20.0
 troposphere==2.3.3
 unittest2==1.1.0
+-e git+ssh://git@github.com/elifesciences/threadbare@develop#egg=threadbare
--- a/py2-requirements.txt
+++ b/py2-requirements.txt
@@ -24,4 +24,4 @@ PyYAML==4.2b2
 requests==2.20.0
 troposphere==2.3.3
 unittest2==1.1.0
--e git+ssh://git@github.com/elifesciences/threadbare@develop#egg=threadbare
+parallel-ssh
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,7 +16,6 @@ pylint==1.8.2
 pytest==3.5.0
 pytest-cov==2.5.1
 pytest-forked==0.2
-pytest-xdist==1.22.2
 python-dateutil==2.7.0
 python-slugify==1.2.4
 python-terraform==0.10.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,4 +23,4 @@ PyYAML==4.2b2
 requests==2.20.0
 troposphere==2.3.3
 unittest2==1.1.0
-parallel-ssh
+parallel-ssh==1.9.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,3 +25,4 @@ PyYAML==4.2b2
 requests==2.20.0
 troposphere==2.3.3
 unittest2==1.1.0
+-e git+ssh://git@github.com/elifesciences/threadbare@develop#egg=threadbare
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,4 +25,4 @@ PyYAML==4.2b2
 requests==2.20.0
 troposphere==2.3.3
 unittest2==1.1.0
--e git+ssh://git@github.com/elifesciences/threadbare@develop#egg=threadbare
+parallel-ssh
--- a/src/buildercore/bootstrap.py
+++ b/src/buildercore/bootstrap.py
@@ -51,7 +51,8 @@ def run_script(script_filename, *script_params, **environment_variables):
 
     env_string = ['%s=%s' % (k, v) for k, v in environment_variables.items()]
     cmd = ["/bin/bash", remote_script] + lmap(escape_string_parameter, list(script_params))
-    retval = remote_sudo(" ".join(env_string + cmd))
+    result = remote_sudo(" ".join(env_string + cmd))
+    retval = result['return_code']
     remote_sudo("rm " + remote_script) # remove the script after executing it
     end = datetime.now()
     LOG.info("Executed script %s in %2.4f seconds", script_filename, (end - start).total_seconds())
--- a/src/buildercore/command.py
+++ b/src/buildercore/command.py
@@ -75,25 +75,28 @@ NetworkError = fab_exceptions.NetworkError
 # api
 #
 
-def fab_api_local_wrapper(*args, **kwargs):
+def fab_api_local_remote_wrapper(fab_result):
     """Fabric returns the stdout of the command when capture=True, with stderr and some values also available as attributes.
     This modifies the behaviour of Fabric's `local` to return a dictionary of results."""
-    fab_result = fab_api.local(*args, **kwargs)
-    # https://github.com/mathiasertl/fabric/blob/master/fabric/operations.py#L1240-L1251
+    # local:
+    # - https://github.com/mathiasertl/fabric/blob/master/fabric/operations.py#L1240-L1251
+    # run/sudo:
+    # - https://github.com/mathiasertl/fabric/blob/master/fabric/operations.py#L898-L971
     result = fab_result.__dict__
     result['stdout'] = (fab_result or b"").splitlines()
     result['stderr'] = (fab_result.stderr or b"").splitlines()
     return result
 
-local = api(fab_api_local_wrapper, threadbare.operations.local)
-execute = api(fab_api.execute, partial(threadbare.execute.execute_with_hosts, env))
-parallel = api(fab_api.parallel, threadbare.execute.parallel)
-serial = api(fab_api.serial, threadbare.execute.serial)
+def fab_api_local_wrapper(*args, **kwargs):
+    return fab_api_local_remote_wrapper(fab_api.local(*args, **kwargs))
 
-hide = api(fab_api.hide, threadbare.operations.hide)
+def fab_api_run_wrapper(*args, **kwargs):
+    return fab_api_local_remote_wrapper(fab_api.run(*args, **kwargs))
+
+def fab_api_sudo_wrapper(*args, **kwargs):
+    return fab_api_local_remote_wrapper(fab_api.sudo(*args, **kwargs))
 
 # https://github.com/mathiasertl/fabric/blob/master/fabric/context_managers.py#L158-L241
-
 def fab_api_settings_wrapper(*args, **kwargs):
     "a context manager that alters mutable application state for functions called within it's scope"
 
@@ -105,13 +108,22 @@ def fab_api_settings_wrapper(*args, **kwargs):
 
     return fab_api.settings(*args, **kwargs)
 
+#
+
+local = api(fab_api_local_wrapper, threadbare.operations.local)
+execute = api(fab_api.execute, partial(threadbare.execute.execute_with_hosts, env))
+parallel = api(fab_api.parallel, threadbare.execute.parallel)
+serial = api(fab_api.serial, threadbare.execute.serial)
+
+hide = api(fab_api.hide, threadbare.operations.hide)
+
 settings = api(fab_api_settings_wrapper, threadbare.state.settings)
 
 lcd = api(fab_api.lcd, threadbare.operations.lcd) # local change dir
 rcd = api(fab_api.cd, threadbare.operations.rcd) # remote change dir
 
-remote = api(fab_api.run, threadbare.operations.remote)
-remote_sudo = api(fab_api.sudo, threadbare.operations.remote_sudo)
+remote = api(fab_api_run_wrapper, threadbare.operations.remote)
+remote_sudo = api(fab_api_sudo_wrapper, threadbare.operations.remote_sudo)
 upload = api(fab_api.put, threadbare.operations.upload)
 download = api(fab_api.get, threadbare.operations.download)
 remote_file_exists = api(fab_files.exists, threadbare.operations.remote_file_exists)
--- a/src/buildercore/command.py
+++ b/src/buildercore/command.py
@@ -82,7 +82,7 @@ def fab_api_local_wrapper(*args, **kwargs):
     # https://github.com/mathiasertl/fabric/blob/master/fabric/operations.py#L1240-L1251
     result = fab_result.__dict__
     result['stdout'] = (fab_result or b"").splitlines()
-    result['stderr'] = (fab_result.stderr or b"").splitlines() 
+    result['stderr'] = (fab_result.stderr or b"").splitlines()
     return result
 
 local = api(fab_api_local_wrapper, threadbare.operations.local)
--- a/src/buildercore/command.py
+++ b/src/buildercore/command.py
@@ -7,10 +7,7 @@ import fabric.state
 import fabric.network
 import logging
 from io import BytesIO
-from . import utils
-import threadbare.operations
-import threadbare.state
-import threadbare.execute
+from . import utils, threadbare
 from functools import partial
 
 THREADBARE = 'threadbare'
--- a/src/buildercore/command.py
+++ b/src/buildercore/command.py
@@ -75,7 +75,17 @@ NetworkError = fab_exceptions.NetworkError
 # api
 #
 
-local = api(fab_api.local, threadbare.operations.local)
+def fab_api_local_wrapper(*args, **kwargs):
+    """Fabric returns the stdout of the command when capture=True, with stderr and some values also available as attributes.
+    This modifies the behaviour of Fabric's `local` to return a dictionary of results."""
+    fab_result = fab_api.local(*args, **kwargs)
+    # https://github.com/mathiasertl/fabric/blob/master/fabric/operations.py#L1240-L1251
+    result = fab_result.__dict__
+    result['stdout'] = (fab_result or b"").splitlines()
+    result['stderr'] = (fab_result.stderr or b"").splitlines() 
+    return result
+
+local = api(fab_api_local_wrapper, threadbare.operations.local)
 execute = api(fab_api.execute, partial(threadbare.execute.execute_with_hosts, env))
 parallel = api(fab_api.parallel, threadbare.execute.parallel)
 serial = api(fab_api.serial, threadbare.execute.serial)
--- a/src/buildercore/command.py
+++ b/src/buildercore/command.py
@@ -1,5 +1,4 @@
 import os
-import time
 import fabric.api as fab_api
 import fabric.contrib.files as fab_files
 import fabric.exceptions as fab_exceptions
@@ -8,7 +7,6 @@ import fabric.network
 import logging
 from io import BytesIO
 from . import utils, threadbare
-from functools import partial
 
 THREADBARE = 'threadbare'
 FABRIC = 'fabric'
@@ -22,29 +20,6 @@ def api(fabric_fn, threadbare_fn):
     return fabric_fn if BACKEND == FABRIC else threadbare_fn
 
 COMMAND_LOG = []
-
-_default_env = {}
-_default_env.update(fab_api.env)
-
-def envdiff():
-    "temporary. returns only the elements that are different between the default Fabric env and the env as it is right now"
-    return {k: v for k, v in fab_api.env.items() if _default_env.get(k) != v}
-
-def spy(fn):
-    "temporary. wrapper to inspect inputs to commands"
-    # Fabric 1.14 documentation: https://docs.fabfile.org/en/1.14/
-    def _wrapper(*args, **kwargs):
-        timestamp = time.time()
-        funcname = getattr(fn, '__name__', '???')
-        lst = [timestamp, funcname, args, kwargs]
-        print(lst)
-        COMMAND_LOG.append(lst)
-        with open('/tmp/command-log.jsonl', 'a') as fh:
-            msg = utils.json_dumps({"ts": timestamp, "fn": funcname, "args": args, "kwargs": kwargs, 'env': envdiff()}, dangerous=True)
-            fh.write(msg + "\n")
-        result = fn(*args, **kwargs)
-        return result
-    return _wrapper
 
 def no_match(msg):
     def fn(*args, **kwargs):
@@ -111,7 +86,7 @@ def fab_api_settings_wrapper(*args, **kwargs):
 #
 
 local = api(fab_api_local_wrapper, threadbare.operations.local)
-execute = api(fab_api.execute, partial(threadbare.execute.execute_with_hosts, env))
+execute = api(fab_api.execute, threadbare.execute.execute_with_hosts)
 parallel = api(fab_api.parallel, threadbare.execute.parallel)
 serial = api(fab_api.serial, threadbare.execute.serial)
 
--- a/src/buildercore/command.py
+++ b/src/buildercore/command.py
@@ -17,18 +17,16 @@ BACKEND = os.environ.get('BLDR_BACKEND', DEFAULT_BACKEND)
 assert BACKEND in [FABRIC, THREADBARE]
 
 def api(fabric_fn, threadbare_fn):
+    "accepts two functions and returns the one matching the currently set BACKEND"
     return fabric_fn if BACKEND == FABRIC else threadbare_fn
 
-COMMAND_LOG = []
-
 def no_match(msg):
+    "used in rare circumstances when either a function in Fabric or Threadbare doesn't have a corollary in the other"
     def fn(*args, **kwargs):
         return None
     return fn
 
 LOG = logging.getLogger(__name__)
-
-env = api(fab_api.env, threadbare.state.ENV)
 
 #
 # exceptions
@@ -37,12 +35,11 @@ env = api(fab_api.env, threadbare.state.ENV)
 class CommandException(Exception):
     pass
 
-# no un-catchable errors from Fabric
-# env.abort_exception = CommandException # env is just a dictionary with attribute access
-
-# TODO: how to handle this ...
-# with initial_settings() ... ? we could explicitly go from an empty environment to default settings
-#env['abort_exception'] = CommandException
+if BACKEND == FABRIC:
+    # no un-catchable errors from Fabric
+    fab_api.env['abort_exception'] = CommandException
+else:
+    threadbare.state.set_defaults({"abort_exception": CommandException})
 
 NetworkError = fab_exceptions.NetworkError
 
@@ -84,6 +81,8 @@ def fab_api_settings_wrapper(*args, **kwargs):
     return fab_api.settings(*args, **kwargs)
 
 #
+
+env = api(fab_api.env, threadbare.state.ENV)
 
 local = api(fab_api_local_wrapper, threadbare.operations.local)
 execute = api(fab_api.execute, threadbare.execute.execute_with_hosts)
--- a/src/buildercore/command.py
+++ b/src/buildercore/command.py
@@ -50,8 +50,11 @@ class CommandException(Exception):
     pass
 
 # no un-catchable errors from Fabric
-#env.abort_exception = CommandException
-env['abort_exception'] = CommandException # env is just a dictionary with attribute access
+#env.abort_exception = CommandException # env is just a dictionary with attribute access
+
+# TODO: how to handle this ...
+# with initial_settings() ... ? we could explicitly go from an empty environment to default settings
+#env['abort_exception'] = CommandException
 
 NetworkError = fab_exceptions.NetworkError
 
--- a/src/buildercore/command.py
+++ b/src/buildercore/command.py
@@ -9,35 +9,38 @@ import fabric.network
 import logging
 from io import BytesIO
 from . import utils
-import threadbare
-
+import threadbare.operations
+import threadbare.state
+import threadbare.execute
+from functools import partial
 COMMAND_LOG = []
 
 _default_env = {}
 _default_env.update(fab_api.env)
 
 def envdiff():
-    "returns only the elements that are different between the default Fabric env and the env as it is right now"
+    "temporary. returns only the elements that are different between the default Fabric env and the env as it is right now"
     return {k: v for k, v in fab_api.env.items() if _default_env.get(k) != v}
 
 def spy(fn):
+    "temporary. wrapper to inspect inputs to commands"
     def _wrapper(*args, **kwargs):
-        result = fn(*args, **kwargs)
-
         timestamp = time.time()
-        funcname = fn.__name__
-        
-        COMMAND_LOG.append([timestamp, funcname, args, kwargs])
+        funcname = getattr(fn, '__name__', '???')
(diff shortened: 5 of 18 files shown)


lsh-0:
(will let it die from linting for the time being - can Alfred distinguish between a draft PR and regular PR?)

example output of spying while running tests: [command-log.jsonl.indented.txt](https://github.com/elifesciences/builder/files/3862307/command-log.jsonl.indented.txt)

requirements.txt
```diff
@@ -25,3 +25,4 @@ PyYAML==4.2b2
 requests==2.20.0
 troposphere==2.3.3
 unittest2==1.1.0
+-e git+ssh://git@github.com/elifesciences/threadbare@develop#egg=threadbare
```
giorgiosironi: (I am not supposed to look at this yet; but in the long term between the two routes of installing the library and pasting it into the `builder` project, the latter seems preferable as it makes fixes and refactorings much easier)

lsh-0: you can totally look at it if you like ;) I'm quite proud of how well it's working.

I knew going in that separating it out into another project would troublesome, but the speed of development and isolation have been really good. Of course, now the two must be joined somehow that lets us make fixes quickly.

The `-e` flag means 'editable' and I've been working on it in `builder/venv/src/threadbare/`. Theoretically this should be good enough, but I suspect it won't be.

Another approach might be a sub-repository.

Copying and pasting threadbare into builder may be the least clever approach but probably the one with fewer problems. I'll investigate all three.

I'm able to completely launch and bootstrap an ec2 instance now with no problems. 

Exception handling has had some attention to make it consistent with Fabric but I need to take a second look, same with keyboard interrupts during parallel/multiprocessing operations.

Works in python2+3

Slices, dices.

lsh-0: Went with a simple copy for now. Will sync changes back and forth manually until it's stable and gets a proper pypi release.

lsh-0:
so with the default backend still pointed to Fabric, `./.run-tests-ci.sh py35` pass successfully for me in 20mins and `./.run-tests-ci.sh py27` pass successfully in 21mins.

switching backend to threadbare to see what happens

lsh-0:
python27 had 5 failures and 316 passes with a threadbare backend in 11mins 

python35 had 5 failures and 316 passes with a threadbare backend in 10mins

unless the three previous tests, this one didn't exit with a gevent error and I have some useful output on what failed

lsh-0:
I can see two instances running that shouldn't be, probably from the failed tests:

* dummy1--luke-20191204063841-469764--1
* dummy1--luke-20191204065438-393877--1

destroying

lsh-0:
@giorgiosironi , I can see an adhoc instance still running that probably shouldn't be:

* basebox--2019-11-adhoc--1

I *think* I might have created it, but don't remember. I'm happy to tear it down if you are

giorgiosironi:
> I think I might have created it, but don't remember. I'm happy to tear it down if you are.

I'm not working on baseboxes at the moment, destroy :+1:

requirements.txt
giorgiosironi: > Slices, dices.

But does it make julienne?

lsh-0: no. julienne requires python 4 and a mysterious fifth python parallel programming library that can only be detected when a thread within a subprocess attempts to run a geventlet in an asyncio future.

src/buildercore/core.py
```diff
@@ -392,11 +394,11 @@ def current_ec2_node_id():
 
     Sample value: 'i-0553487b4b6916bc9'"""
 
-    ensure(env.host is not None, "This is supposed to be called with settings for connecting to an EC2 instance")
-    current_public_ip = env.host
+    ensure(env['host'] is not None, "This is supposed to be called with settings for connecting to an EC2 instance")
```
lsh-0: difference in behaviour between fabric and threadbare - 'host' may already exist in fabric as it has loads of defaults, threadbare has zero defaults and an empty dict.

change this to `ensure('host' in env and env['host'], ...)`

lsh-0: (self-reviewing, ignore me)

lsh-0:
woo ! `./.run-tests-ci.sh py35` passes locally in 26mins.

alfred just hangs indefinitely.

investigating just what is going on with the test runner to make it so unhappy

lsh-0:
removing pytest-xdist and execnet the py27 tests complete without issue in 25mins

I'm expecting something similar for py35

lsh-0:
py27 looks like it passed all tests but something else failed afterwards ('invocation error').

this doesn't seem related to threadbare/parallel-ssh, switching backend to threadbare now tests are (ostensibly) passing

lsh-0:
interestingly I'm getting the `/venv` error on Alfred as well. I thought it was just the crazy invocation I was doing.

at least I can replicate it locally

lsh-0:
yes! **YES!** Your time has come Fabric !  

bff

Super green.

py2-requirements.txt
```diff
@@ -23,3 +22,4 @@ PyYAML==4.2b2
 requests==2.20.0
 troposphere==2.3.3
 unittest2==1.1.0
+parallel-ssh
```
giorgiosironi: no version constraint?

Dockerfile.py3
```diff
@@ -17,6 +21,7 @@ RUN apk add --no-cache --virtual build-deps \
     pip install virtualenv && \
     mkdir /venv && \
     virtualenv --python=python3 /venv && \
+    /venv/bin/pip install parallel-ssh && \
```
giorgiosironi: isn't this already in `requirements.txt`?

src/buildercore/bootstrap.py
```diff
@@ -51,7 +51,8 @@ def escape_string_parameter(parameter):
 
     env_string = ['%s=%s' % (k, v) for k, v in environment_variables.items()]
     cmd = ["/bin/bash", remote_script] + lmap(escape_string_parameter, list(script_params))
-    retval = remote_sudo(" ".join(env_string + cmd))
+    result = remote_sudo(" ".join(env_string + cmd))
+    retval = result['return_code']
```
giorgiosironi: was this wrong while using the original Fabric `remote_sudo`, too? As it seems the backend can be switched from Fabric to Threadbare and back, I assume this code supports both

src/buildercore/command.py
```diff
@@ -20,23 +35,41 @@
...
 #
 # api
 #
 
-local = fab_api.local
-execute = fab_api.execute
-parallel = fab_api.parallel
-serial = fab_api.serial
-hide = fab_api.hide
+def fab_api_local_remote_wrapper(fab_result):
```
giorgiosironi: the name is a bit confusing given there is also `fab_api_local_wrapper`; this is one layer down from other `fab_api_*_wrapper` functions

src/buildercore/command.py
```diff
@@ -47,15 +80,30 @@ def settings(*args, **kwargs):
...
+remote_file_exists = api(fab_files.exists, threadbare.operations.remote_file_exists)
 
-remote = fab_api.run
-remote_sudo = fab_api.sudo
-upload = fab_api.put
-download = fab_api.get
-remote_file_exists = fab_files.exists
-network_disconnect_all = fabric.network.disconnect_all
+network_disconnect_all = api(fabric.network.disconnect_all,
+                             no_match("threadbare automatically closes ssh closes connections"))
```
giorgiosironi: would call this `no_op()` as the implementation does nothing (there could be one that throws an exception for example)

src/buildercore/core.py
```diff
@@ -243,10 +243,12 @@ def stack_pem(stackname, die_if_exists=False, die_if_doesnt_exist=False):
     return expected_key
 
 def _ec2_connection_params(stackname, username, **kwargs):
-    "returns a dictionary of settings to be used with Fabric's api.settings context manager"
+    "returns a dictionary of settings to be used with `command.settings` context manager"
     # http://docs.fabfile.org/en/1.14/usage/env.html
     params = {'user': username}
     pem = stack_pem(stackname)
+    # todo: handle other non-rsa private keys
```
giorgiosironi: ignores `pem`? What's the preference between the two?

lsh-0:
> Only got to half of this today

that's fine, take your time, there's a lot

src/buildercore/bootstrap.py
lsh-0: It wasn't wrong, I've made [the return value consistent](https://github.com/elifesciences/builder/pull/586/files/155b9bbbc6129889bbdbfa98fa4c4f1c63badab7#diff-18a6b93c14cdde804a9939c596c81067R62-R69) across backends. The Fabric backend now returns a dictionary of results for `remote`, `local` and `remote_sudo` versus just the return code with stdout, stderr etc attached as attributes to an object.

src/buildercore/command.py
lsh-0: I've renamed if `fab_api_results_wrapper`, converted it to a decorator and removed the intermediate functions.

src/buildercore/command.py
lsh-0: renamed

src/buildercore/core.py
lsh-0: do you mean `id_rsa.pem` and `id_rsa.pub` ? `ssh-keygen` doesn't output a `.pem` file.

I think this change to `_ec2_connection_params` can be reverted and the `key_filename` pushed back into `command.py` and the call to `threadbare.state.set_defaults`. `set_defaults` was implemented only recently.

lsh-0: > ignores pem? What's the preference between the two?

oh, I think I see what you mean now. It looks like I was setting a default. Fabric has *loads* of defaults, the path to a private key among them. Threadbare starts with no defaults. There is a case here where the bootstrap user's `pem` key isn't added because the expected file doesn't exist. Fabric would default to the user's key, Threadbare would fail.

I've moved this particular default into `command.py`. I'll also expand it to check several other common keys. I've seen implementations in both Fabric and PSSH for that

src/buildercore/threadbare/common.py
```diff
@@ -0,0 +1,114 @@
...
+    pass
+
+
+def first(x):
+    "returns the first element in an collection of things"
+    if x is None:
+        return x
+    try:
+        if isinstance(x, dict):
+            return list(x.items())[0]
```
lsh-0: this is suspect. native ordered dicts were only introduced in python3.7 and the docker containers are using python3.5

src/buildercore/command.py
```diff
@@ -1,17 +1,32 @@
...
 import logging
 from io import BytesIO
-from . import utils
+from . import utils, threadbare
 
-LOG = logging.getLogger(__name__)
+THREADBARE = 'threadbare'
+FABRIC = 'fabric'
+
+DEFAULT_BACKEND = THREADBARE
```
giorgiosironi: could merge with Fabric first, and rollout Threadbare as the second step?

src/utils.py
```diff
@@ -31,8 +31,8 @@ def git_remote_branches(url):
 
 def _git_remote_refs(url):
     cmd = "git ls-remote --heads %s" % url
-    output = local(cmd, capture=True)
-    return [line.split() for line in output.splitlines()]
+    output = local(cmd, capture=True)['stdout']
```
giorgiosironi: `stdout` is always split in lines? Not all commands do work on multiple lines e.g. binary output

src/buildercore/threadbare/__init__.py
```diff
@@ -0,0 +1,19 @@
+# gevent is used by parallel-ssh which interferes with Python multiprocessing and futures
+# it can cause indefinite blocking.
+# this bit of magic appears to make everything work nicely with each other.
+# - http://www.gevent.org/api/gevent.monkey.html
+from gevent import monkey
+
+monkey.patch_all()
```
giorgiosironi: I've [had this failing before I think](https://github.com/elifesciences/elife-spectrum/pull/44#issuecomment-297313791), because of a library that was doing it when it shouldn't, or wasn't doing it when it should have. Keep an eye on it during rollout.

src/buildercore/threadbare/common.py
```diff
@@ -0,0 +1,114 @@
...
+    sudo_prefix = "sudo --non-interactive"
+    space = " "
+    return sudo_prefix + space + command
+
+
+def pwd_wrap_command(command, working_dir):
+    "adds a 'cd' prefix to command"
+    prefix = 'cd "%s" &&' % working_dir
+    space = " "
+    return prefix + space + command
```
giorgiosironi: any form of string formatting/interpolation would be safer than `+` e.g. producing better errors if `None` is passed in as `command` accidentally

src/buildercore/threadbare/common.py
```diff
@@ -0,0 +1,114 @@
...
+    # https://github.com/mathiasertl/fabric/blob/master/fabric/state.py#L374-L376
+    # note: differs from Fabric. they support interactive input of password, users and groups
+    # we use it exclusively to run commands as root
+    sudo_prefix = "sudo --non-interactive"
+    space = " "
+    return sudo_prefix + space + command
+
+
+def pwd_wrap_command(command, working_dir):
+    "adds a 'cd' prefix to command"
```
giorgiosironi: names seem misleading, isn't `cd_wrap_command`?

src/buildercore/threadbare/state.py
```diff
@@ -0,0 +1,130 @@
+import copy
+import contextlib
+
+CLEANUP_KEY = "_cleanup"
+
+
+class LockableDict(dict):
```
giorgiosironi: in a context of concurrency this name is scary as it evokes locks as in semaphores. `FreezableDict`?

src/buildercore/threadbare/state.py
```diff
@@ -0,0 +1,130 @@
...
+    def update(self, new_dict):
+        if self.read_only:
+            raise ValueError(
+                "dictionary is locked attempting to `update` with %r" % new_dict
+            )
+        dict.update(self, new_dict)
+
+    def __setitem__(self, key, val):
+        # I suspect multiprocessing isn't copying the custom 'read_only' attribute back
+        # from the child process results. be aware of this weirdness
```
giorgiosironi: I am not sure I know all the stack of libraries that provides multiple processes here, but it's usually safe to say that once you have multiple processes variables diverge and are not in communication with each other. Whatever communication channel is there would be built on stdin/stdout?

giorgiosironi: Assuming the results are encoded on stdout somehow.

src/buildercore/threadbare/state.py
```diff
@@ -0,0 +1,130 @@
...
+
+def _add_cleanup(state, fn):
+    cleanup_fn_list = state.get(CLEANUP_KEY, [])
+    cleanup_fn_list.append(fn)
+    state[CLEANUP_KEY] = cleanup_fn_list
+
+
+def add_cleanup(fn):
+    "add a function to a list of functions that are called after leaving the current scope of the context manager"
+    return _add_cleanup(ENV, fn)
```
giorgiosironi: needs a better differentiation from `add_cleanup()` in terms of naming

src/buildercore/threadbare/execute.py
```diff
@@ -0,0 +1,225 @@
...
+
+    # poll the processes until all are complete
+    # remove process from pool when it is complete
+    while len(pool) > 0:
+        for idx, running_p in enumerate(pool):
+            result = process_status(running_p)
+            if not result["alive"]:
+                result_map[result["name"]] = result
+                del pool[idx]
+        time.sleep(0.1)
```
giorgiosironi: magic sleep? what does it do?

giorgiosironi: ah, polling interval; usually I ask for a timeout here but this could be arbitrarily long, deploys that take tens of minutes due to backup being made, etc.

src/buildercore/threadbare/operations.py
```diff
@@ -0,0 +1,638 @@
+import tempfile
+import contextlib
+import subprocess
+from threading import Timer
```
giorgiosironi: :policeman: mixing processes and threads - will get to the point where this is used

src/buildercore/threadbare/operations.py
```diff
@@ -0,0 +1,638 @@
...
+            # builder: https://github.com/elifesciences/builder/blob/master/src/buildercore/core.py#L345-L347
+            # pssh: https://github.com/ParallelSSH/parallel-ssh/blob/8b7bb4bcb94d913c3b7da77db592f84486c53b90/pssh/clients/native/parallel.py#L272-L274
+            pssh_exceptions.Timeout: "Timed out trying to connect.",
+            # builder: https://github.com/elifesciences/builder/blob/master/src/buildercore/core.py#L348-L350
+            # fabric: https://github.com/mathiasertl/fabric/blob/master/fabric/network.py#L601-L605
+            # pssh: https://github.com/ParallelSSH/parallel-ssh/blob/2e9668cf4b58b38316b1d515810d7e6c595c76f3/pssh/exceptions.py
+            pssh_exceptions.SSHException: "Low level socket error connecting to host.",
+            pssh_exceptions.SessionError: "Low level socket error connecting to host.",
+            pssh_exceptions.ConnectionErrorException: "Low level socket error connecting to host.",
+        }
```
giorgiosironi: I suspected there was code relying on Fabric's error messages rather than exception types. The exception types were likely easy to find, and these messages match perfectly what we depend on.

src/buildercore/threadbare/operations.py
```diff
@@ -0,0 +1,638 @@
...
+    )
+
+    shell = False  # handled ourselves
+    sudo = False  # handled ourselves
+    user = None  # user to sudo to
+    encoding = "utf-8"  # used everywhere
+
+    try:
+        # https://parallel-ssh.readthedocs.io/en/latest/native_single.html#pssh.clients.native.single.SSHClient.run_command
+        channel, host_string, stdout, stderr, stdin = client.run_command(
```
giorgiosironi: that's a lot of arguments on the input and output side; any data structure like a dictionary where they could be stored together? Easier to give the whole concept a name e.g. a `dict` `commandOptions` (terrible name but you get the point) for all the other arguments than `command`

src/buildercore/threadbare/operations.py
```diff
@@ -0,0 +1,638 @@
...
+        local_bytes.seek(0)  # reset internal pointer
+        temp_file, local_path = tempfile.mkstemp(suffix="-threadbare")
+        with os.fdopen(temp_file, "wb") as fh:
+            fh.write(local_bytes.getvalue())
+        cleanup = lambda: os.unlink(local_path)
+        return local_path, cleanup
+    return local_path, None
+
+
+def upload(local_path, remote_path, use_sudo=False, **kwargs):
```
giorgiosironi: Hoping this file won't need much maintenance as it is a lot of (forked from Fabric/written/rewritten?) code

giorgiosironi:
Finished last round on the `threadbare` module and approved. Go through the comments and feel free to (slowly) rollout.

src/buildercore/command.py
lsh-0: there is already a PR for that: https://github.com/elifesciences/builder/pull/593

I opened to ensure the Fabric backend continued to work while I juggled the two. We can merge that one first if you'd like, it's a branch of this branch.

src/utils.py
lsh-0: There is an option in Fabric I haven't implemented (many actually), but this one is called [`linewise`](https://github.com/mathiasertl/fabric/blob/master/sites/docs/usage/env.rst#linewise)

> Forces buffering by line instead of by character/byte, typically when running in parallel mode. This option is implied by `env.parallel` -- even if linewise is False, if parallel is True then linewise behavior will occur.

which is in contrast to parallel-ssh that *only* [gives you lines](https://github.com/ParallelSSH/parallel-ssh/blob/5f4a4137e58cd7d7716e8193b4de12e52ffb7832/pssh/clients/native/single.py#L401-L406). That `output_buffer` comes from [this function](https://github.com/ParallelSSH/parallel-ssh/blob/379b8713fa79592b7e45ee4c9b8715efa36e525e/pssh/native/_ssh2.pyx#L33-L67) that is reading the bytes directly until it encounters a line separator then yielding. At that point it's quite close to the C++ wrapper.

I really can appreciate the need to occasionally have access to the raw bytes as they're coming out but we're just not doing that anywhere. In the cases where we were doing that we ended up reading by line anyway.

There may be gremlins here with pssh+parallel code, I'm not sure.

src/buildercore/threadbare/__init__.py
lsh-0: Parallelism in Python is a shambles if you ask me and likely to cause a headache later. I had to remove xdist because of it.

I'm eyeing ParallelSSH and wondering if we can't fork it and strip out all of the parallel stuff. It was originally using Paramiko until they developed their own backend called Python-SSH and now they offer either backend to use, so there is a Paramiko dependency it's pulling in as well. Another headache and maintenance burden I don't want.

I was using Python-SSH to begin with but it was too low level and I didn't want to be messing about with sockets and control codes.

src/buildercore/threadbare/state.py
lsh-0: another victim burnt by deadlocks ;) fair enough, I'll rename it

src/buildercore/threadbare/state.py
lsh-0: > Whatever communication channel is there would be built on stdin/stdout?

in this case, it's Python's `multiprocessing` that is doing the encoding of data and passing it to child processes [here](https://github.com/elifesciences/threadbare/blob/1.0-prep/threadbare/execute.py#L30-L57). I was having unexpected behaviour here as new dicts should have been locked and empty, instead they were empty and unlocked, which to me meant that the the `state` module wasn't being loaded properly somehow as the `env` is initialised and locked in the global scope. 

I ended up resetting the state before executing the `worker_func` and this seems to work fine.

src/buildercore/threadbare/state.py
lsh-0: why?

lsh-0: not "why do we need to give functions clear names", but "why does the wrapper need to clearly differentiate itself from it's wrapped private function?".

it seems clear as crystal to me

src/buildercore/threadbare/execute.py
lsh-0: yes, it introduces the slightest of delays so that we're not manically polling the process every microsecond. I'm happy to implement a longer poll duration if you're expecting a long running process. Failed processes would obviously take longer to reveal themselves.

timeouts are also available in `local`, `remote` and `remote_sudo`. `local` timeouts use Python's `Timer` to kill the process. Remote timeouts rely on pssh's implementation.

I'll add a note as to why we're sleeping here

src/buildercore/threadbare/operations.py
lsh-0: Threadbare has a well defined scope and an established project (Fabric) to take lessons from. I have no intention of swapping multiprocessing for threading.

That said, you shouldn't be afraid of threading but of programming languages and paradigms that let you shoot yourself in the foot with threads.

src/buildercore/threadbare/operations.py
lsh-0: exactly. A little gross but it gets the job done.

src/buildercore/threadbare/operations.py
lsh-0: > as it is a lot of (forked from Fabric/written/rewritten?) code

I wrote all of this myself. It wasn't a clean room reverse engineering process and I had to do a lot of code spelunking to figure out Fabric idioms we were relying on, but what we have here in Threadbare and these three modules is quite lean and straightforward. It has excellent test coverage, including of remote commands.

But yes - definitely a lot of code and ideas to digest and review at once. I don't think you should approve it until you can do so without hesitation or regret. I'm happy to take any part of it and explain it in-depth.

src/buildercore/threadbare/operations.py
lsh-0: `_execute` is the exception here. I made the function signature more verbose so there wouldn't be any ambiguity while writing tests. Outputs can't be helped and hiding them in a named tuple or fancy dict would obfuscate it. `_execute` shouldn't be used directly anyway. Threadbare *is* a library. I should look into generating some docs for it.

`operations` *does* suffer from the many-options problem in general though. Fabric defined it's `env` full of defaults that these operations would then pull values out of and then compare them with in an inconsistent adhoc way. The trade-off I've made in Threadbare is that these additional options are in a `**kwargs` bucket for each function. Unsupported args are removed and argument precedence is handled in a consistent way (see `handle`). The downside to this is you need to inspect the code to see if an argument is supported or not.

lsh-0:
> I don't think you should approve it until you can do so without hesitation or regret

ah, I see you've already approved it. Well. I'll merge [that branch](https://github.com/elifesciences/builder/pull/593) of this branch which has Fabric as the default backend and keep this PR open. The changes will be reduced to just the backend switching to Threadbare.

At that point you'll be able to run commands like:

    BLDR_BACKEND=threadbare ./bldr launch:foo,bar,baz

And get a feel for the slight changes in output that have been introduced. Maybe run some commands in parallel.

src/utils.py
lsh-0: By gremlins I mean interleaving of output across parallel processes. I haven't noticed any but that could be because it's well handled elsewhere

src/buildercore/threadbare/common.py
lsh-0: support for dicts removed

src/utils.py
lsh-0: >  IIRC Fabric tarnished the output anyway e.g. by inserting IP addresses of the server on every line

we have more control over this now. I disabled the PSSH logger so it wouldn't do the exact same thing. Once support is merged you can change the backend with an environment variable and decide what sort of output you would prefer

src/buildercore/threadbare/common.py
lsh-0: That's not possible here. For one thing it would die earlier in `_shell_escape`. I'll ensure `command` is a string.

lsh-0:
urgh. it's too late and too hot to be dealing with weird python string issues. I'll handle it tomorrow

src/buildercore/threadbare/state.py
giorgiosironi: Not really sure what the separation is doing as this seems to be the only call to `_add_cleanup` and given the `_` I'd consider it private to this module :thinking: Nitpicking, do not want to change it either way

giorgiosironi:
I am not up to speed with all the branching but I'm ok for this to get into `master` with:
- tests passing
- Fabric as default backend (may go through syncing with the other PR)

Then I can local experiments with a canonical version on `master`, run commands on Jenkins etc. with the Threadbare backend.

src/buildercore/threadbare/state.py
lsh-0: it follows a general pattern I've taken on with Clojure where side effecting code is pushed to the fringes, typically because testing in Clojure is slightly harder than in Python so the less scaffolding, the better.

In this case however my instincts are wrong as the 'pure' function is modifying a global state dict and not returning anything. This module underwent a lot of churn late in the process as I removed a big feature and added `set_defaults`. I'll take a second look at it today as I fix up the py2 typing issue breaking CI

lsh-0:
yup, no worries

PR #593, opened 2019-12-10
Threadbare integration with Fabric backend as default

lsh-0:
just a (very) simple means to test Fabric is still working while threadbare gets finished off


lsh-0:
hrm - that weird issue again where everything passes but ultimately fails with an 'invocation error'

lsh-0:
ok, figured it out.

the "InvocationError" I'm guessing is coming from `tox`, and the error is that the py27 environment is run with `coverage` configuration and will fail if less than 69% (it's 64% locally).

interestingly enough, the py3 environment doesn't have these coverage options.

the coverage has dropped way down because threadbare is being copied from it's package into builder

lsh-0:
> the coverage has dropped way down because threadbare is being copied from it's package into builder

FWIW, threadbare coverage is at 97%

lsh-0:
I've merged the changes from the review feedback into `feat-replace-fabric1` and then those changes in to this branch (`feat-replace-fabric1--fabric`). This branch has Fabric as the default backend so once merged it shouldn't cause a fuss and we can test threadbare manually if we like.

PR #594, opened 2019-12-11
bootstrap, python2.7 removal

lsh-0:
* ~changes salt version to `2018.3.4`~ separate PR
* adds command to clone formula repositories locally. Used for checking formulas for instances of python2-isms and changed states
* adds switch to bootstrap script to cleanly disable installation of python2 in future
* adds python library `docker` so salt can use `docker_*` states

prepares the path for switch away from py2 and salt 2018.3 that uses py3

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1,7 +1,6 @@
 defaults:
     description: defaults for all projects in this file
-    #salt: '2017.7.8' # the version of salt these project use
-    salt: '2018.3.4'
+    salt: '2018.3.4' # the version of salt these project use
     # use false with a subdomain to assign internal addresses only
     domain: elifesciences.org
     # addressing within VPC
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1,6 +1,7 @@
 defaults:
     description: defaults for all projects in this file
-    salt: '2018.3.4' # the version of salt these project use
+    #salt: '2018.3.4' # the version of salt these project use
+    salt: '2017.7.8'
     # use false with a subdomain to assign internal addresses only
     domain: elifesciences.org
     # addressing within VPC
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1,6 +1,7 @@
 defaults:
     description: defaults for all projects in this file
-    salt: '2017.7.8' # the version of salt these project use
+    #salt: '2017.7.8' # the version of salt these project use
+    salt: '2018.3.4'
     # use false with a subdomain to assign internal addresses only
     domain: elifesciences.org
     # addressing within VPC
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -7,6 +7,8 @@ set -e # everything must pass
 set -u # no unbound variables
 set -xv  # output the scripts and interpolated steps
 
+export DEBIAN_FRONTEND=noninteractive # no ncurses prompts
+
 echo "-----------------------------"
 
 if [ ! "$#" -ge 3 ]; then
@@ -96,7 +98,7 @@ if $upgrade_python; then
 
     apt-get install python3 python3-dev -y
 
-    # virtual envs have to be recreated
+    # virtualenvs have to be recreated
     find /srv /opt -depth -type d -name venv -exec rm -rf "{}" \;
 
     # install/upgrade pip+setuptools
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -154,7 +154,8 @@ if ($installing || $upgrading); then
     # -c  Temporary configuration directory
     # -M  Also install master
     # https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh
-    sh salt_bootstrap.sh -x python3 -P -F -c /tmp stable "$version"
+    #sh salt_bootstrap.sh -x python3 -P -F -c /tmp stable "$version"
+    sh salt_bootstrap.sh -P -F -c /tmp stable "$version"
 else
     echo "Skipping minion bootstrap, found: $(salt-minion --version)"
 fi
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -123,9 +123,6 @@ if $upgrade_python3; then
     apt-get install python3 python3-dev python3-pip python3-setuptools -y --no-install-recommends
     python3 -m pip install pip setuptools --upgrade
 
-    # virtualenvs have to be recreated
-    find /srv /opt -depth -type d -name venv -exec rm -rf "{}" \;
-
     # some libraries need to be installed *before* calling Salt
     python3 -m pip install "docker[tls]==4.1.0"
 
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -166,7 +166,8 @@ if [ "$install_master" = "true" ]; then
     # salt is not installed or the version installed is old
     if ! (command -v salt-master > /dev/null && salt-master --version | grep "$version"); then
         # master not installed
-        sh salt_bootstrap.sh -x python3 -P -F -M -c /tmp stable "$version"
+        #sh salt_bootstrap.sh -x python3 -P -F -M -c /tmp stable "$version"
+        sh salt_bootstrap.sh -P -F -M -c /tmp stable "$version"
     else
         echo "Skipping master bootstrap, found: $(salt-master --version)"
     fi
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -108,7 +108,7 @@ if $upgrade_python2; then
         apt-get install python2.7 python2.7-dev -y
 
         # virtualenvs have to be recreated
-        find /srv /opt -depth -type d -name venv -exec rm -rf "{}" \;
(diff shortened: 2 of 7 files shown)


projects/elife.yaml
```diff
@@ -1,6 +1,7 @@
 defaults:
     description: defaults for all projects in this file
-    salt: '2017.7.8' # the version of salt these project use
+    #salt: '2017.7.8' # the version of salt these project use
```
lsh-0: rm

src/project.py
```diff
@@ -31,6 +32,34 @@ def context(pname, output_format=None):
     formatter = formatters.get(output_format)
     return formatter(cfngen.build_context(pname, stackname=core.mk_stackname(pname, "test")))
 
+@requires_project
+def clone_project_formulas(pname):
+    """clones a project's list of formulas to `cloned-projects/$formulaname` if it doesn't already exist.
```
lsh-0: comma

src/project.py
```diff
@@ -31,6 +32,34 @@ def context(pname, output_format=None):
...
+        fpath = os.path.join(destination, os.path.basename(furl))
+        if os.path.exists(fpath):
+            cmd = "cd %s; git pull" % (fpath,)
+        else:
+            cmd = "cd %s; git clone %s" % (destination, furl)
+
+        try:
+            result = local(cmd)
+        except CommandException:
+            pass # any error is printed to stdout/stderr
```
lsh-0: use `warn_only=True`

lsh-0: actually, fabric doesn't support `warn_only` as a keyword parameter, so use `settings` (Threadbare supports both)

lsh-0:
(the build problems are lint issues. keeping it that way to stop unnecessary builds while I fix things)

lsh-0:
I found a subtle state bug in the bootstrap script so that python2 and it's libraries (and thus python3, as I had it) were not upgraded on first run. It was because the default installed Python2 in Ubuntu 18.04 is Python 2.17 so `upgrade_python` never became `true`

lsh-0:
this PR contains a salt version upgrade, which means the master needs to be upgraded then all of the minions. There is [a doc for it here](https://github.com/elifesciences/builder/blob/master/docs/salt.md#upgrading-salt). 

Once this PR is approved I'll that as a :+1: for upgrading Salt.

The state of that doc and those commands should be addressed so I'll do that afterwards in another builder PR.

I imagine there will be bugs in formulas I haven't caught yet and will fix those as I go along.

The loris upgrade is dependent on this PR.

scripts/bootstrap.sh
```diff
@@ -66,35 +68,79 @@ else
...
+elife_depends_on_python2=true
+
+if $upgrade_python2; then
+
+    if $elife_depends_on_python2; then
+        echo "eLife still has formulas that depend on Python2!"
+        apt-get install python2.7 python2.7-dev -y
 
+        # virtualenvs have to be recreated
+        find /srv /opt -depth -type d -name venv -exec rm -rf "{}" \;
```
giorgiosironi: sounds a bit dangerous as there could be other software in `/opt`, though there is `venv` limiting it; probably ok

scripts/bootstrap.sh
```diff
@@ -66,35 +68,79 @@ else
...
+    # record an entry about when python3 was installed/upgraded
+    # presence of this entry is used to skip this section in future, unless forced with a flag
+    if [ -f /root/upgrade-python3.flag ]; then
+        echo "$(date -I) -- installed/upgraded python3 (forced)" >> /root/events.log;
+    else
+        echo "$(date -I) -- installed/upgraded python3" >> /root/events.log;
+    fi
+
+    # remove 'force upgrade' flag, if it exists
+    rm -f /root/upgrade-python3.flag
```
giorgiosironi: could be inside the `then`?

scripts/bootstrap.sh
```diff
@@ -105,12 +151,13 @@ if ($installing || $upgrading); then
...
     wget -O salt_bootstrap.sh https://bootstrap.saltstack.com --no-verbose
 
+    # -x  Changes the Python version used to install Salt.
     # -P  Allow pip based installations.
     # -F  Allow copied files to overwrite existing(config, init.d, etc)
     # -c  Temporary configuration directory
     # -M  Also install master
     # https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh
-    sh salt_bootstrap.sh -P -F -c /tmp stable "$version"
+    sh salt_bootstrap.sh -x python3 -P -F -c /tmp stable "$version"
```
giorgiosironi: strictly required to roll this out at the same time as the new bootstrap script and its possible actions? Maybe yes because of dependencies, just checking

giorgiosironi: (if not it could be a follow-up PR)

scripts/bootstrap.sh
```diff
@@ -149,7 +196,7 @@ done < <(env | grep '^grain_.*')
 echo "$grains" > /etc/salt/grains
 
 # restart salt-minion. necessary as we may have changed minion's configuration
-systemctl restart salt-minion 2> /dev/null || service salt-minion restart
```
giorgiosironi: :+1:

src/buildercore/config.py
```diff
@@ -116,11 +116,13 @@ def get_logger(name):
 
 #
 # settings
-# believe it or not but buildercore.config is NOT the place for user config
+# buildercore.config is NOT the place for user config
 #
 
 PROJECTS_FILES = ['projects/elife.yaml']
 
+PROJECT_FORMULAS = os.path.join(PROJECT_PATH, 'cloned-projects') # same path as used by Vagrant
```
giorgiosironi: unclear name, looks like a list of formulas but it's actually a path to their folder

src/project.py
```diff
@@ -31,6 +32,32 @@ def context(pname, output_format=None):
...
 
+@requires_project
+def clone_project_formulas(pname):
+    """clones a project's list of formulas to `cloned-projects/$formulaname`, if it doesn't already exist.
+    if it does exist, it attempts to update it with a `git pull`."""
+    destination = config.PROJECT_FORMULAS
+    pdata = project.project_data(pname)
+
+    formula_url_list = [pdata.get('formula-repo')]
+    formula_url_list.extend(pdata.get('formula-dependencies', []))
```
giorgiosironi: sounds like this does a lot of `git pull` in `builder-base-formula`; perhaps keep track of what has already been pulled, though don't want to make this method stateful?

scripts/bootstrap.sh
lsh-0: (it's not my code ;)

I'm happy to drop this. IIRC there were issues with old/incompatible cryptography/certificates between 2.7.9 and 2.7.12 and then other things between 2.7.12 and 2.7.17

scripts/bootstrap.sh
```diff
@@ -66,35 +68,79 @@ else
...
 
+if $upgrade_python3; then
+    apt-get install python3 python3-dev python3-pip python3-setuptools -y --no-install-recommends
+    python3 -m pip install pip setuptools --upgrade
+
+    # virtualenvs have to be recreated
+    find /srv /opt -depth -type d -name venv -exec rm -rf "{}" \;
+
+    # some libraries need to be installed *before* calling Salt
+    python3 -m pip install "docker[tls]==4.1.0"
```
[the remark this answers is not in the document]
lsh-0: it's a bootstrap script, it installs and configures what is necessary for Salt to run. For docker I prefer using Salt states over bash (or docker compose) scripts. `cmd.run` is a fallback if all the other alternatives won't do.

scripts/bootstrap.sh
lsh-0: It could be. The purpose of that flag is to force an upgrade of python2. It's unlikely, but if we set it and `elife_depends_on_python2` was `false`, the flag would hang around forever.

That whole block will gone soon enough. It's barely worth thinking about.

scripts/bootstrap.sh
lsh-0: That's an interesting point. I could remove that `-x python3` and the salt version bump so the other changes could go through and be tested - I have made *a lot* of them - so that the follow up PR is just two small changes without all the other things that could go wrong.

lsh-0: I've reverted those two changes now. This PR can be merged a bit more safely now.

scripts/bootstrap.sh
giorgiosironi: comment out and reinstate if something breaks?

src/buildercore/project/__init__.py
```diff
@@ -94,19 +92,15 @@ def ec2_projects(*args, **kwargs):
...
 #
 
 def known_formulas():
-    "a simple list of all known project formulas (excluding the private-repo)"
-    lst = utils.unique(utils.shallow_flatten(project_formulas().values()))
-    if None in lst:
-        lst.remove(None)
-    return lst
+    "a simple list of all known project formulas"
+    return utils.lfilter(None, utils.unique(utils.shallow_flatten(project_formulas().values())))
```
giorgiosironi: :+1:

scripts/bootstrap.sh
lsh-0: I can do that

lsh-0:
all green, here we go ...

PR #597, opened 2020-01-07
bootstrap, prevent ncurses prompt when installing libssl1.1

lsh-0:


lsh-0:
surprised we haven't encountered this before now

scripts/bootstrap.sh
```diff
@@ -1,11 +1,14 @@
 #!/bin/bash
 # *ALL INSTANCES*
-# copied into the virtual machine and executed. DO NOT run on your host machine.
+# copied/uploaded onto the virtual machine and executed. DO NOT run on your host machine.
+# run as root
```
giorgiosironi: important, as environment variables won't survive `sudo` usage

lsh-0: precisely

PR #605, opened 2020-01-28
Feature: next-salt alt-config

lsh-0:
* adds new alt-config 'next-salt' to base config

todo:

- [x] bootstrap script is receiving old version of Salt still.
- [x] hide pip progress bar in this call `python3 -m pip install pip setuptools --upgrade`
- [x] address this `ERROR: requests 2.22.0 has requirement idna<2.9,>=2.5, but you'll have idna 2.0 which is incompatible.` (see log)

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -216,8 +216,8 @@ defaults:
 
         next-salt:
             description: uses the next version of Salt to test formula for problems upgrading. OS agnostic.
+            salt: '2018.3.4'
             ec2:
-                salt: '2018.3.4'
                 masterless: True
 
         1804:
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -216,8 +216,8 @@ defaults:
 
         next-salt:
             description: uses the next version of Salt to test formula for problems upgrading. OS agnostic.
-            salt: '2018.3.4'
             ec2:
+                salt: '2018.3.4'
                 masterless: True
 
         1804:
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -214,7 +214,7 @@ defaults:
                 # should be the same as basebox '18.04' and 'standalone18.04' aws-alt configurations
                 ami: ami-0b425589c7bb7663d # bionic, build 20180814, hvm:ebs-ssd
 
-        next-salt:
+        standalone-next-salt:
             description: uses the next version of Salt to test formula for problems upgrading. OS agnostic.
             ec2:
                 salt: '2018.3.4'
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -213,6 +213,13 @@ defaults:
             ec2:
                 # should be the same as basebox '18.04' and 'standalone18.04' aws-alt configurations
                 ami: ami-0b425589c7bb7663d # bionic, build 20180814, hvm:ebs-ssd
+
+        next-salt:
+            description: uses the next version of Salt to test formula for problems upgrading. OS agnostic.
+            ec2:
+                salt: '2018.3.4'
+                masterless: True
+
         1804:
             description: Ubuntu 18.04 (Bionic)
             ec2: true # explicitly specify an 18.04 ami when 18.04 is no longer the default (2023?)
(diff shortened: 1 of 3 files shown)


lsh-0:
Running a build now. 

Even if it were using the next version of salt (it's not for some reason), it's being installed using python2.

lsh-0:
so this works, but without switching bootstrap over to python3 it's not raising the error I'm encountering in vagrant

```
elife@next-salt-test-3--libraries:~$ salt-call --versions
Salt Version:
           Salt: 2018.3.4
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: 1.3.7
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Oct  8 2019, 14:14:10)
   python-gnupg: 0.3.8
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-1030-aws
         system: Linux
        version: Ubuntu 16.04 xenial
```

projects/elife.yaml
```diff
@@ -213,6 +213,13 @@ defaults:
             ec2:
                 # should be the same as basebox '18.04' and 'standalone18.04' aws-alt configurations
                 ami: ami-0b425589c7bb7663d # bionic, build 20180814, hvm:ebs-ssd
+
+        next-salt:
```
giorgiosironi: maybe `standalone-next-salt` since it has `masterless`?

src/buildercore/bootstrap.py
```diff
@@ -486,7 +486,15 @@ def _update_ec2_node():
         # write out environment config (/etc/cfn-info.json) so Salt can read CFN outputs
         write_environment_info(stackname, overwrite=True)
 
+        # bit of a hack, but project config merging doesn't apply to top-level values
+        # in this case we look for an alternate salt version under a project's 'ec2' section
         salt_version = context['project']['salt']
+        alt_salt_version = context['ec2'].get('salt')
+
+        # and we only use it if we're going masterless
```
giorgiosironi: why the restriction to masterless? Could eventually be relaxed?

src/buildercore/bootstrap.py
```diff
@@ -486,7 +486,15 @@ def _update_ec2_node():
         # write out environment config (/etc/cfn-info.json) so Salt can read CFN outputs
         write_environment_info(stackname, overwrite=True)
 
+        # bit of a hack, but project config merging doesn't apply to top-level values
+        # in this case we look for an alternate salt version under a project's 'ec2' section
         salt_version = context['project']['salt']
+        alt_salt_version = context['ec2'].get('salt')
```
giorgiosironi: eventually it will be correct for `salt` to be under `ec2` as it's not used without EC2 servers involved (e.g. the `generic-cdn` project). Would support that change, in a separate PR

src/buildercore/bootstrap.py
lsh-0: it would require the master server to be running the same version of Salt. I don't recall any problems with mismatches in Salt versions between client and server but we've always done it with master first.

it could definitely be relaxed if we wanted to test that aspect as well

src/buildercore/bootstrap.py
lsh-0: yes, I started going down that route but it got complex quickly. Another option might be to revisit the configuration merging and get rid of aws-alt, gcp-alt, vagrant-alt and simply have 'alt' configs where any part of the structure could be overridden. That's a bit more ambitious though ;)

lsh-0:
> Seems incomplete according to the checklist

yes, I noticed pip complaining while bringing up an instance and thought if I don't fix it here it won't get fixed ever

projects/elife.yaml
lsh-0: weren't we deprecating the long alt-config names because they screwed with something in CI?

`sNextSalt` perhaps?

lsh-0:
pip error is tied to Ubuntu 16.04 and not present on 18.04 instances.

Projects still on 16.04 will be dealt with early this year

PR #616, opened 2020-02-11
tox, updates references to python3.5 to python3 in general

lsh-0:


lsh-0:
after we're both satisfied with the threadbare backend, we can remove python2 support from builder entirely. It will help reduce test times as well

PR #620, opened 2020-02-17
Allow update of IntDomainName

giorgiosironi:
`search--continuumtest` wasn't showing the hanging problem solved and
described by
https://github.com/elifesciences/search/pull/247
because its lack of `IntDomainName` caused
https://github.com/elifesciences/builder-base-formula/blob/beb38e50407acc55458839be27d51b2eb66f4af9/elife/hostname.sls#L16-L23
not to be executed.

Wouldn't merge this straight away until that is cleared up, as it may break random software that relies on `localhost` in the same mysterious way as Gearman.

### DIFF
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -553,7 +553,7 @@ def generate_stack(pname, **more_context):
 
 
 # can't add ExtDNS: it changes dynamically when we start/stop instances and should not be touched after creation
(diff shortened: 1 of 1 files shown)


lsh-0:
> Wouldn't merge this straight away until that is cleared up, as it may break random software that relies on localhost in the same mysterious way as Gearman.

Indeed. The change looks innocuous and I'm happy for it to be merged, but we should run highstate on the critical path of apps once it's merged to check for shenanigans

giorgiosironi:
When running `update_infrastructure`, by default also the update of EC2 containing highstate runs so that should surface the problems.

giorgiosironi:
Tried to run `update_infrastructure` on various critical path projects, on `prod` and `continuumtest`. No outstanding changes after this merge.

PR #622, opened 2020-02-18
alt-configs, adds sNextSalt

lsh-0:
* identical to standalone-next-salt, but shorter name,
* marked latter as deprecated


lsh-0:
fyi @giorgiosironi

PR #623, opened 2020-02-18
lowercase bucket names when using mixedcase alt-configs

lsh-0:
* min and max length to s3 bucket names
* ~make a s3 **bucket name** all lower case (not key path)~

### DIFF
--- a/src/tests/fixtures/projects/dummy-project.yaml
+++ b/src/tests/fixtures/projects/dummy-project.yaml
@@ -266,10 +266,6 @@ project-with-s3:
                 public: true
             widgets-encrypted-{instance}:
                 encryption: arn:aws:kms:us-east-1:1234:key/12345678-1234-1234-1234-123456789012
-            # not recommended, however 'instance' here may be derived from an alt-config name
-            # that may in turn contain MixedCase alphanumeric
-            widgets-MixedCase-{instance}:
-                public: true
 
 project-with-ext:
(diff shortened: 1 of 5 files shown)


src/tests/fixtures/projects/dummy-project.yaml
```diff
@@ -266,6 +266,10 @@ project-with-s3:
                 public: true
             widgets-encrypted-{instance}:
                 encryption: arn:aws:kms:us-east-1:1234:key/12345678-1234-1234-1234-123456789012
+            # not recommended, however 'instance' here may be derived from an alt-config name
+            # that may in turn contain MixedCase alphanumeric
+            widgets-MixedCase-{instance}:
```
giorgiosironi: can we consider just refusing this and tell the user to change their mind? I understand this can be automatically generated from an alt config name, but shouldn't that also stick to the simplest possible combination of characters as it gets into hostnames, DNS, etc?

lsh-0: I'll enforce all lowercase and change the name of the alt config

giorgiosironi: Thank you, it's a great simplification when seeing this from the outside

lsh-0:
I've removed the support for mixed case alt-config names and s3 bucket normalisation. It leaves the validation of the bucket name in though. Very long bucket names have been a problem in the past

PR #626, opened 2020-02-27
adds 's-next-salt' alt-config

lsh-0:
the other two alt-configs are deprecated and will be removed shortly

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -214,21 +214,21 @@ defaults:
                 # should be the same as basebox '18.04' and 'standalone18.04' aws-alt configurations
                 ami: ami-0b425589c7bb7663d # bionic, build 20180814, hvm:ebs-ssd
 
-        # deprecated, use `s-next-salt` instead
+        # deprecated, use `snsalt` instead
         standalone-next-salt:
             description: uses the next version of Salt to test formula for problems upgrading. OS agnostic.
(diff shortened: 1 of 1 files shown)


lsh-0:
@giorgiosironi , here is your opportunity to use a name you prefer. Whatever we go with here I'll update all other usages to it.

projects/elife.yaml
```diff
@@ -205,26 +205,34 @@ defaults:
...
+        # note! only use lowercase and hyphens for alt-config names.
 
         fresh:
             description: uses a public Ubuntu AMI instead of an elife generated basebox.
             ec2:
                 # should be the same as basebox '18.04' and 'standalone18.04' aws-alt configurations
                 ami: ami-0b425589c7bb7663d # bionic, build 20180814, hvm:ebs-ssd
 
-        # deprecated, use `sNextSalt` instead
+        # deprecated, use `s-next-salt` instead
```
giorgiosironi: My main constraint when using e.g. `s1804` was length of hostnames that get generated, for example `pr-27-base-update--personalised-covers.elifesciences.org` or similar. I'd try to keep it as short as possible, maybe `snextsalt`? `snsalt`?

lsh-0: > snsalt

ha. snail salt. I can live with that

PR #631, opened 2020-03-05
microsimulation-demo, removes project definition

lsh-0:


giorgiosironi:
Nothing is running (hence still to be destroyed) so I'll merge.

PR #637, opened 2020-03-30
search shouldn't listen for updates to subjects

giorgiosironi:
Discovered while deploying https://github.com/elifesciences/issues/issues/4608

We know `search` doesn't index `subject`s as they are not in the possible results.

https://github.com/elifesciences/search/blob/856d3931264783e78ed120a79b76d6ced4cca27c/src/Search/GearmanTransformer.php#L21-L28 shows the mapping of types for bus messages coming in, and there's no mention of `subject`.

When a `subject` message is in the queue, its processing fails even before that:
https://github.com/elifesciences/bus-sdk-php/blob/3af0739fac110e79ea7f21aadf8d17b891f3b918/src/Queue/BasicTransformer.php#L57

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1235,7 +1235,6 @@ search:
                 subscriptions:
                     - bus-articles--{instance}
                     - bus-podcast-episodes--{instance}
-                    - bus-subjects--{instance}
                     - bus-collections--{instance}
                     - bus-interviews--{instance}
                     - bus-blog-articles--{instance}


lsh-0:
(looked like a transitory error, kicked build, all green)

giorgiosironi:
Applied to `search--end2end` and there was 1 remaining subscriptions at https://console.aws.amazon.com/sns/v3/home?region=us-east-1#/topic/arn:aws:sns:us-east-1:512686554592:bus-subjects--end2end as originally there were two. Will clean up manually.
Permissions were correctly removed from https://console.aws.amazon.com/sqs/home?region=us-east-1#

giorgiosironi:
Applied to `ci` and `continuumtest` too.

giorgiosironi:
Also cleaned up some spurious subscriptions from `recommendations--*` which doesn't have any SQS queue anymore.
As it stands, nothing should listen to `bus-subjects--*` SNS topics.

PR #639, opened 2020-04-07
threadbare, sync with latest changes (SCP backend)

lsh-0:
* switches downloads and uploads to use SCP from using SFTP
* adds support for running local commands with `sudo`

### DIFF
--- a/src/buildercore/threadbare/operations.py
+++ b/src/buildercore/threadbare/operations.py
@@ -1,3 +1,4 @@
+from functools import wraps
 from datetime import datetime
 import tempfile
 import contextlib
@@ -7,6 +8,7 @@ import getpass
 from pssh import exceptions as pssh_exceptions
 import os, sys
 from pssh.clients.native import SSHClient as PSSHClient
+import gevent
 import logging
 from . import state, common
 from .common import (
@@ -18,6 +20,7 @@ from .common import (
     sudo_wrap_command,
     cwd_wrap_command,
     shell_wrap_command,
+    ensure,
 )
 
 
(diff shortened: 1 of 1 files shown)


lsh-0:
The big change here is that you can specify SFTP or SCP to transfer files. 

Paramiko uses SFTP so that is what I went with with threadbare, however Paramiko uses it's own SFTP implementation which is faster than native. I'm not sure if it's native SFTP or the Parallel-SSH library's usage of it, but performance degrades as the file sizes get larger. SCP on the other hand outperforms Paramko SFTP and native SFTP.

I encountered one issue with uploading via SCP during CI that I couldn't replicate locally, possibly because I'm still using magnetic spinning discs, but there was a delay between uploading a file, having it report that is was successfully completed, checking the file and finding it missing. Performing a check that the remote file exists after the upload successfully completed seemed to fix this. SFTP never suffered from this.

Another factor that may have contributed to the original issue is that Parallel-SSH was returning a pool of `gevent` 'greenlets' that weren't being forced to join the main process. They now are. My gut feeling is this wasn't the real cause but may have obfuscated stacktraces.

lsh-0:
once this is merged in could you attempt to replicate that problem you were having in Alfred, @giorgiosironi ?

giorgiosironi:
As far as I know reproducing the problem just means running https://alfred.elifesciences.org/job/process/job/process-journal-cms-restore-continuumtest/ which has been modified to pass in the `BLDR_BACKEND=threadbase` environment variable.

lsh-0:
> which has been modified to pass in the BLDR_BACKEND=threadbase environment variable.

alright then, I'll give that a go once Alfred has finished updating itself.

lsh-0:
It's a bust. Debugging these alfred scripts to figure out what they're doing

PR #654, opened 2020-04-24
threadbare, sync with proper to include hanging fix

lsh-0:
A fix for an issue that only presented itself while downloading a 6gb file remotely. 

Python's Multiprocessing never acknowledged the worker having finished and just kept it alive indefinitely. Now we monitor the queue for results and once we have the expected number of results, we stop blocking and terminate any processes that are still alive. If we find any, a `warning` is issued.


lsh-0:
fyi @giorgiosironi

PR #663, opened 2020-05-15
threadbare, set default backend to sftp

lsh-0:


src/buildercore/threadbare/operations.py
```diff
@@ -558,7 +558,10 @@ def _transfer_fn(client, direction, **kwargs):
         # sftp is *exceptionally* slow so SCP is used by default.
```
lsh-0, line 558: 'used by default'

.run-tests-ci.sh
```diff
@@ -17,7 +17,7 @@ else
     coverage_options=
 fi
 
-pytest \
+./patched_pytest \
```
lsh-0, line 20: explain

lsh-0: should this script be used in the other test scripts (if it isn't already?)

PR #665, opened 2020-05-18
Salt 3000 upgrade branch

lsh-0:
builder doesn't need it's own special branch for the Salt 3000 upgrade, but I've re-discovered [an unsafe operation](https://github.com/elifesciences/builder/pull/489/files#r268079733) that slipped past code review and it should be dealt with for the next Salt upgrade.


lsh-0:
(just linting issues, all good)

PR #675, opened 2020-06-19
fix(update.sh): make compatible with zsh

erkannt:
quotes around grep pattern for ssh-agent ps aux fix things for zsh


lsh-0:
(I wouldn't hang around waiting for a review from Giorgio or Chris before merging)

PR #682, opened 2020-07-07
logging, tweaks to quieten certain loggers

lsh-0:
- [ ] unhelpful 'debug' noise is pushed down to debug level
- [ ] stderr handler only prints 'info' and above log messages
- [ ] file handler still captures 'debug' and above log messages
- [ ] `./logs/app.log` is rotated at a certain file size


lsh-0:
(note to self: look at existing log file to find unhelpful debug level noise that is at info or above levels)

PR #686, opened 2020-07-14
Allow users other than the creator to run `update_infrastructure` on Kubernetes clusters

giorgiosironi:
The `user` IAM role generated with the cluster is how the cluster authorizes access to read e.g. ConfigMaps during `terraform plan`. Normally we pass it to `aws eks update-kubeconfig`.

When using the Kubernetes Terraform provider, we need to assume that role explicitly: 
https://github.com/elifesciences/issues/issues/5775

The `providers.tf` file now features a list rather than a map, to allow the addition of an additional `aws` provider.

The additional `aws` provider has a `eks_assume_role` alias and so any resource may choose to use it over the default one.


erkannt:
This works for me. :partying_face:

PR #696, opened 2020-08-03
containers, OS upgrade

lsh-0:


projects/elife.yaml
```diff
@@ -277,8 +277,7 @@ defaults:
 
 builder:
     formula-repo: https://github.com/elifesciences/builder-formula
-    vagrant:
-        box: bento/ubuntu-16.04
+    vagrant: {}
```
lsh-0, line 280: unrelated, but sneaking it in

PR #698, opened 2020-08-07
python3, re-enable py3 build

lsh-0:
the python3 docker build is failing and had to be disabled to unblock other work.


lsh-0:
bumping base container to python3.5-alpine3.12 from alpine3.8

PR #700, opened 2020-08-10
add elife-libero-reviewer--staging rds

erkannt:
- copied basic project structure from `error-pages`
- copied rds settings from `elife-xpub--`staging`
- reusing encryption key from xpub

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1687,6 +1687,24 @@ digests:
         ports:
             1290: 80
 
+
+elife-libero-reviewer:
+    description: replacement for xpub
+    default-branch: null
+    domain: null
+    intdomain: null
+    aws:
+        ec2: false
+    aws-alt:
+        staging:
+            rds:
+                engine: postgres # or 'MySQL'
+                version: '10.6'
+                type: db.t2.small
+                storage: 5 # GB
+                backup-retention: 2 # days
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -1687,6 +1687,24 @@ digests:
...
+    default-branch: null
+    domain: null
+    intdomain: null
+    aws:
+        ec2: false
+    aws-alt:
+        staging:
+            rds:
+                engine: postgres # or 'MySQL'
+                version: '10.6'
```
lsh-0, line 1702: any flexibility in getting this bumped to postgresql 11 ? I'm aware the versions of mysql/psql that we're using are aging and will need to be upgraded soon.

projects/elife.yaml
```diff
@@ -1687,6 +1687,24 @@ digests:
...
+    description: replacement for xpub
+    default-branch: null
+    domain: null
+    intdomain: null
+    aws:
+        ec2: false
+    aws-alt:
+        staging:
+            rds:
+                engine: postgres # or 'MySQL'
```
lsh-0, line 1701: no todo, but prefer postgresql over mysql

projects/elife.yaml
erkannt: Yes. This will be part of our testing of xpub to reviewer migration.

PR #730, opened 2020-11-03
threadbare, sync from main repository

lsh-0:
fixes divergence in behaviour with fabric where parallel tasks re-raise any errors after execution results become available


src/buildercore/threadbare/operations.py
```diff
@@ -659,7 +659,7 @@ def _transfer_fn(client, direction, **kwargs):
         # - https://github.com/ParallelSSH/parallel-ssh/issues/177
         # however, SCP is buggy and may randomly hang or complete without uploading anything.
         # take slow and reliable over fast and buggy.
-        "transfer_protocol": "rsync", # "sftp",  # "scp"
```
lsh-0, line 662: this never made it into threadbare. revert

PR #738, opened 2020-12-02
feat: create aws resources for libero editor

NuclearRedeye:
Update the builder config to create staging and prod resources for libero editor.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1694,30 +1694,21 @@ elife-libero-reviewer:
                     encryption: arn:aws:kms:us-east-1:512686554592:key/b6f44e77-46d8-41bb-a2e5-75c0588a2b20
 
 elife-libero-editor:
-    default-branch: null
-    domain: null
-    intdomain: null
     aws:
         ec2: false
-    aws-alt:
-        staging:
+        s3:
+            "{instance}-elife-libero-editor": {}
+        sqs:
+            # Monitor the production process bucket for new deposits from Exeter.
+            elife-libero-editor-import--{instance}:
+                subscriptions:
+                    - "elife-production-processes"
+        docdb: {}
+    aws-alt: 
+        prod: 
             s3:
                 "{instance}-elife-libero-editor":
-            sqs:
-                # Monitor the production process bucket for new deposits from Exeter.
-                elife-libero-editor-import--{instance}:
-                    subscriptions:
-                        - "elife-production-processes"
-            docdb: {}
-        prod:
-            s3:
-                "{instance}-elife-libero-editor":
-            sqs:
-                # Monitor the production process bucket for new deposits from Exeter.
-                elife-libero-editor-import--{instance}:
-                    subscriptions:
-                        - "elife-production-processes"
-            docdb: {}
+                    deletion-policy: delete
 
 task-adept:
     description: production data dumps from TaskAdept
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1703,9 +1703,21 @@ elife-libero-editor:
         staging:
             s3:
                 "{instance}-elife-libero-editor":
+            sqs:
+                # Monitor the production process bucket for new deposits from Exeter.
+                elife-libero-editor-import--{instance}:
+                    subscriptions:
+                        - "elife-production-processes"
+            docdb: {}
         prod:
             s3:
                 "{instance}-elife-libero-editor":
+            sqs:
+                # Monitor the production process bucket for new deposits from Exeter.
+                elife-libero-editor-import--{instance}:
+                    subscriptions:
+                        - "elife-production-processes"
+            docdb: {}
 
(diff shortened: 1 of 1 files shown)


projects/elife.yaml
```diff
@@ -1693,6 +1693,32 @@ elife-libero-reviewer:
                 "{instance}-elife-libero-reviewer":
                     encryption: arn:aws:kms:us-east-1:512686554592:key/b6f44e77-46d8-41bb-a2e5-75c0588a2b20
 
+elife-libero-editor:
+    default-branch: null
+    domain: null
+    intdomain: null
+    aws:
+        ec2: false
+    aws-alt:
```
NuclearRedeye, line 1702: @lsh-0 Wasn't sure what the difference between the `aws` and `aws-alt` keys are? At a guess wondered if the later is for Libero?

Cooryd:
@lsh-0 this was our attempt to complete the requirements defined in libero/editor#585. Could you check that we've approached this in the right way and there isn't anything missing / broken with our implementation please.

lsh-0:
sure thing, I'll take a look at this today

projects/elife.yaml
```diff
@@ -1693,6 +1693,32 @@ elife-libero-reviewer:
                 "{instance}-elife-libero-reviewer":
                     encryption: arn:aws:kms:us-east-1:512686554592:key/b6f44e77-46d8-41bb-a2e5-75c0588a2b20
 
+elife-libero-editor:
```
lsh-0, line 1696: the 'elife' prefix isn't necessary. I think 'libero-editor' is fine.

projects/elife.yaml
```diff
@@ -1693,6 +1693,32 @@ elife-libero-reviewer:
...
+    default-branch: null
```
lsh-0, line 1697: obsolete, unused. I would get rid of it.

projects/elife.yaml
lsh-0: I see 'elife-libero-reviewer' is already in there. You might want to keep it for consistency, I don't know. It just isn't necessary.

projects/elife.yaml
lsh-0: `aws-alt` and `gcp-alt` are 'alternate configurations'.

What happens, roughly, is that the 'default' section at the top of the file has the top-level project section merged over it to give us our project configuration. We can create tweaked 'alt' versions of this project config under `aws-alt`. 

In your example, the S3 bucket could live in the regular project config like:

```yaml
elife-libero-reviewer:
    aws:
        ec2: false
        s3:
            "{instance}-elife-libero-reviewer":
                encryption: arn:aws:kms:us-east-1:512686554592:key/b6f44e77-46d8-41bb-a2e5-75c0588a2b20
    aws-alt:
        prod:
            s3:
                "{instance}-elife-libero-reviewer":
                    deletion-policy: retain
```

and then the `prod` environment would ensure the production version of that bucket isn't deleted when the stack is deleted. Or prod could have extra buckets or no buckets, etc.

You can see the final configuration using 

    ./bldr project.data

and picking a project, or in this case:

    ./bldr project.data:elife-libero-editor

default output format is Python, but you can also use JSON or YAML with:

    ./bldr project.data:elife-libero-editor,output_format=yaml

etc.

From what I can see in your proposal the whole thing could live in the top-level with no alt configs:

```
elife-libero-editor:
    aws:
        ec2: false
        s3:
            "{instance}-elife-libero-editor": {}
        sqs:
            # Monitor the production process bucket for new deposits from Exeter.
            elife-libero-editor-import--{instance}:
                subscriptions:
        docdb: {}
```

I would recommend production buckets have an explicit deletion policy though. If it's ok to delete production data, have `deletion-policy: delete` set.

lsh-0:
it needs some tweaks so have a read of my comments.

builder and the project file grew out of defining ec2 instances so there are some top-level values that should live under `ec2` and a few other warts and things. And then we added Google Cloud support and they got their own heading.

I'd love to properly spec the project file out and revisit it's implementation, I think I have a ticket for that, but for now it works. 

Let me know if you have more questions. I'd also prefer to give it a formal github review before it gets merged.

projects/elife.yaml
lsh-0: I remember some other developers coming along once and thinking the 'elife-' prefix was some sort of company policy. It really isn't ;)

Cooryd:
Hi @lsh-0, thanks for the feedback. I've made some changes and I think we're ready for a formal review before this goes in. Hopefully I understood everything correctly. Have left the `elife-` in just to keep it uniform with reviewer for now.

lsh-0 (approved):
looks good.

DocumentDB has two attributes you might be interested in (you can find all properties in the 'default' section): `deletion-protection` and `backup-retention-period`.

I'm not sure how you're going to be using docdb, but you might consider setting those properties as well.

PR #740, opened 2020-12-10
command, 'env' access is now via a function call rather than dict

lsh-0:
during multiprocessing there is a case where an indirect reference to the state isn't updated.

- [x] review

### DIFF
--- a/src/buildercore/command.py
+++ b/src/buildercore/command.py
@@ -79,7 +79,14 @@ def fab_api_settings_wrapper(*args, **kwargs):
 
 #
 
-env = api(fab_api.env, threadbare.state.ENV)
+# lsh@2020-12-10: last minute change to how `env` is accessed.
+# worker processes during multiprocessing seem to hang on to old imported references of `env`.
+# this means `buildercore.command.env` is missing values, but `buildercore.threadbare.state.ENV` is fine.
+# force proper reference by enclosing in a function.
+#env = api(fab_api.env, threadbare.state.ENV)
+def env(key=None):
+    _env = api(fab_api.env, threadbare.state.ENV)
+    return _env[key] if key is not None else _env
 
 local = api(fab_api_results_wrapper(fab_api.local), threadbare.operations.local)
 execute = api(fab_api.execute, threadbare.execute.execute_with_hosts)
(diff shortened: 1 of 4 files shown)


src/buildercore/command.py
```diff
@@ -79,7 +79,14 @@ def fab_api_settings_wrapper(*args, **kwargs):
 
 #
 
-env = api(fab_api.env, threadbare.state.ENV)
+# lsh@2020-12-10: last minute change to how `env` is accessed.
+# worker processes during multiprocessing seem to hang on to old imported references of `env`.
+# this means `buildercore.command.env` is missing values, but `buildercore.threadbare.state.ENV` is fine.
+# force proper reference by enclosing in a function.
+#env = api(fab_api.env, threadbare.state.ENV)
+def env(key=None):
```
lsh-0, line 87: docstr

src/buildercore/lifecycle.py
```diff
@@ -86,7 +86,7 @@ def restart(stackname, initial_states='pending|running|stopping|stopped'):
         raise
 
 def start(stackname):
-    "Puts all EC2 nodes of stackname into the 'started' state. Idempotent"
+    "Puts all EC2 nodes and RDS instances of given `stackname` into the 'started' state. Idempotent"
```
lsh-0, line 89: 'for given'

src/buildercore/lifecycle.py
```diff
@@ -345,7 +345,7 @@ def _select_nodes_with_state(interesting_state, states):
     return [instance_id for (instance_id, state) in states.items() if state == interesting_state]
 
 def _ec2_nodes_states(stackname, node_ids=None):
-    """dictionary from instance id to a string state.
+    """returns a dictionary of instance-id to a string state.
```
lsh-0, line 348: ticks around instance-id

src/buildercore/threadbare/execute.py
```diff
@@ -51,9 +51,8 @@ def _parallel_execution_worker_wrapper(env, worker_func, name, queue):
         result = worker_func()
         queue.put({"name": name, "result": result})
     except BaseException as unhandled_exception:
-        # kept for debugging
-        # import traceback
-        # traceback.print_exc()
+        import traceback
```
lsh-0, line 54: top of file

PR #752, opened 2021-01-21
Editor processes notifications

Cooryd:
For context:

eLife has an S3 bucket called `elife-production-processes` which doesn't appear in the builder configuration so may have been created manually at some point. We need `Put` events within that bucket to be picked up by the ` elife-libero-editor-import--staging` and ` elife-libero-editor-import--prod` SQS instances created at the end of last year. They already have the configuration 

```
sqs:
    elife-libero-editor-import--{instance}:
        subscriptions:
```

- Does adding the config in this PR allow me to alter the config for an existing bucket not created by builder?
- Will not including other bucket config in this file (deletion policies for example) overwrite any config that may already be set on the bucket or will this be left as is?
- Have I missed anything to achieve what I described above?

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1704,9 +1704,11 @@ elife-libero-editor:
         ec2: false
         s3:
             "{instance}-elife-libero-editor": {}
-            "elife-production-processes":
-                sqs-notifications:
-                    elife-production-processes: {}
+            # This bucket is required by editor but was created manually pre-editor so can't be configured via builder.
+            # Config would have been:
+            # "elife-production-processes":
+            #     sqs-notifications:
+            #         elife-production-processes: {}
         sqs:
             # Monitor the production process bucket for new deposits from Exeter.
             elife-libero-editor-import--{instance}:
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1704,6 +1704,9 @@ elife-libero-editor:
         ec2: false
         s3:
             "{instance}-elife-libero-editor": {}
+            "elife-production-processes":
+                sqs-notifications:
+                    elife-production-processes: {}
         sqs:
             # Monitor the production process bucket for new deposits from Exeter.
             elife-libero-editor-import--{instance}:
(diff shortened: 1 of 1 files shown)


lsh-0:
> Does adding the config in this PR allow me to alter the config for an existing bucket not created by builder?

probably not.

> Will not including other bucket config in this file (deletion policies for example) overwrite any config that may already be set on the bucket or will this be left as is?

Assuming it did what you're expecting it to do, which it doesn't, it would probably be destroyed.

> Have I missed anything to achieve what I described above?

Yeah. So, builder only 'recently' got support for creating buckets and there were a tonne of buckets in existence before that support was added. Builder uses Cloudformation for buckets and Cloudformation can be fickle and destructive when it comes to naming conflicts. If the resource is managed by Cloudformation then it will probably be destructive unless you've explicitly added a 'no-delete' policy (assuming the resource supports it), but if the resource already exists and it's *not* managed by Cloudformation then it will roll the changes back and complain about conflicts.

I haven't tried to use Cloudformation to take over the management of already existing infrastructure but [it's supported, apparently](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import-new-stack.html).

I'll take a closer at the code and your changes and see if I can be more helpful.

lsh-0 (commented):
I've taken a closer and I don't think you can do what you want to do.

Because this elife-production-processes bucket is outside of builder management I would configure the PUT notifications manually. I took a quick look and it's pretty straightforward to do, let me or Joel or even Giorgio know if you need a hand.

If you want to document what you've done in a helpful way then perhaps (and this is just a suggestion), keep what you have but comment it out and a line that explains what is happening.

It would mean the configuration change was documented *somewhere* and we might come across it in future while grepping for references to that bucket or SQS queue.

NuclearRedeye (approved):

Cooryd:
Thanks Luke, I've commented out so that it's there documented as suggested. Adding manually seems to be fairly straight forward and I'd requested perms during investigation so should be fine from here.

PR #753, opened 2021-01-21
fix: create security group for docdb access

NuclearRedeye:
Updates to create a security group and open port 27017 for inbound traffic so that clients can connect to documentdb clusters.

This fixes libero/editor#635


NuclearRedeye:
@lsh-0 Sorry I meant to open this as a draft PR, but pressed the wrong button. DocumentDB resources need to create a security group that opens port 27017 hence I had a go at making some changes that I _think_ will do what is needed. Appreciate your thoughts.

src/buildercore/trop.py
```diff
@@ -132,6 +132,16 @@ def rds_security(context):
                           ingress_ports,
                           "RDS DB security group")
 
+def docdb_security(context):
+    """returns a security group for the docdb instance.
+    this security group only allows access within the subnet, not because of the ip address range but because this is dealt with in the subnet configuration"""
+    ingress_data = [27017]
+    ingress_ports = _convert_ports_to_dictionary(ingress_data)
+    return security_group("VPCSecurityGroup",
```
giorgiosironi, line 140: I think this name doesn't need to be unique, so it doesn't clash with the RDS security group

giorgiosironi: (it does need to be unique within the same project, but I don't think projects usually contain both RDS and DocumentDB at the same time)

NuclearRedeye: Ah, yes you are correct. I've updated it to a more meaning full name.

src/tests/test_buildercore_trop.py
```diff
@@ -1217,6 +1217,11 @@ def test_docdb(self):
                                  'Value': 'project-with-docdb--prod'},
                                 {'Key': 'Project',
                                  'Value': 'project-with-docdb'}]},
+                      'VPCSecurityGroups': {
```
NuclearRedeye, line 1220: Not sure what the exact value for this object should be, or if it should be omitted and fed in using the `with patch` statement below. Any help or pointers appreciated.

src/buildercore/trop.py
```diff
@@ -132,6 +132,16 @@ def rds_security(context):
                           ingress_ports,
                           "RDS DB security group")
 
+def docdb_security(context):
```
NuclearRedeye, line 135: Is a unit test needed for this function? I can't see one for the `rds_security` function so I'm assuming just updating the template in `test_docdb` is sufficient.

lsh-0: it's output will be covered in the test you've already touched upon

src/tests/test_buildercore_trop.py
lsh-0: that specific test can be run with:

```bash
cd /path/to/builder
source venv/bin/activate
./.run-tests.sh ./src/tests/test_buildercore_trop.py::TestBuildercoreTrop::test_docdb
```

Scripts with a preceding '.' typically assume the activation of a python virtualenv.

`VpcId` is pulled from the test project file. The test project file has a lot of values similar to the project file in normal use but some are dummied up or missing etc. You can find them in the test fixtures.

lsh-0 (approved):

PR #754, opened 2021-01-22
Remove unused sns subscription

Cooryd:
These SQS queues need to take event notifications directly from the target S3 bucket (`elife-production-processes`) I miss-understood the subscription config option originally and added:

```
subscriptions:	
```

This added a permissions policy rule on the SQS queue of something like: 

```
"Condition": {
  "StringLike": {
    "aws:SourceArn": "arn:aws:sns:us-east-1:11111111111111:elife-production-processes"
  }
}
```
I think it may be overkill to require an SNS service as well as an SQS service at this stage for this project as there are no plans for other services needing to be subscribed to notifications.

What I actually need is something like (have manually set this in for staging queue and it's worked):

```
"Condition": {
  "StringLike": {
    "aws:SourceArn": "arn:aws:s3:::elife-production-processes"
  }
}
```

How might I achieve this through this project file?

Thanks 👍

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1711,9 +1711,7 @@ elife-libero-editor:
             #         elife-production-processes: {}
         sqs:
             # Monitor the production process bucket for new deposits from Exeter.
-            elife-libero-editor-import--{instance}:
-                subscriptions:
-                    - "elife-production-processes"
+            elife-libero-editor-import--{instance}: {}
         docdb: {}
     aws-alt: 
         prod: 


lsh-0:
If I'm understanding you correctly:

1. you thought your config was doing `s3 PUT -> sns -> sqs`
2. you actually want `s3 PUT -> sqs`

It looks like support for this was added to builder at some point because I think [elife-bot is using it](https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L774-L777).

I'll take a look at the implementation and get back to you as there is [this helpful comment](https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L128) in the defaults that contradicts it.

lsh-0:
cc @gnott, do you know what is going on with the SQS notifications attached to buckets in the elife-bot project configuration?

From looking at the code and the commit history it's never been implemented, just placeholders. Within AWS I can see `prod-elife-silent-corrections` bucket is configured to send all creation events to `incoming-queue`. I'm guessing this is done manually?

lsh-0:
There is an [old ticket here](https://github.com/elifesciences/issues/issues/4523) mentioning S3 and SQS and Terraform, but from what I can see Terraform isn't doing any S3 duties unless it's to deposit something into one. I don't think it was ever implemented.

@giorgiosironi , do you know why this wasn't implemented? Is there something very difficult hiding in there?

@Cooryd , if this is urgent I would suggest hooking it up manually like elife-bot is doing. I'll create a ticket to implement this.

lsh-0 (approved):
The change is fine. Perhaps put the PR's description into a ticket in future or hold off on asking for review until it's ready to go.

Cooryd:
I'll merge this in so theres no reference to a non-existent SNS in the SQS's permissions policy. As the source bucket isn't defined in this config file it may be that what I want is already in builder but the policy would have gotten added by: https://github.com/elifesciences/builder/blob/9cc54c471ad38274f06fde1427adf8cb59806b1f/projects/elife.yaml#L1710 either way I'll set it all up manually for now and keep an eye on the services to ensure builders not wiping the manual policy rules out.

Cooryd:
I've just discovered that you can't notify two queues with the same prefix / suffix configuration from a single bucket (must be a technical limitation of S3, learning as I go here). Because we're using a single source bucket for staging & prod it looks like we'll need a single SNS which both SQS can subscribe to to pick up these event notifications. This would remove the requirement of editor to have S3 buckets being able to notify SQS directly.

gnott:
> do you know what is going on with the SQS notifications attached to buckets in the elife-bot project configuration?

I didn't implement these directly, going from memory I believe adding S3 notifications to existing buckets was not possible because it would have meant deleting the buckets (perhaps) in order to add notificaitons, so the actual configuration was applied manually, the traits of them recorded in builder for visibility.

I think the `-elife-bot-decision-letter-input` bucket was configured using builder formulas because it was entirely new.

This may not help, since it sounds like if you need SNS and the builder formulas don't do that, then other examples of SQS are off-topic? Maybe my comments still have value for general knowledge.

PR #757, opened 2021-01-25
Add an SNS for elife-production-processes bucket to notify

Cooryd:
Due to limitation of S3 buckets not being able to notify multiple SQS queues with the same configuration, this adds an SNS which the S3 bucket will notify (manually as the bucket isn't configured via builder) which will then notify SQS. 

https://github.com/libero/editor/issues/644


NuclearRedeye (approved):
Seem OK to me

PR #759, opened 2021-01-26
Move creation of elife-production-processes-events to aws-alt

Cooryd:
We only need a single `elife-production-processes-events` SNS topic, which both of our `prod` and `staging` `elife-libero-editor-import--{instance}` SQS queues will be subscribed to. Having the `elife-production-processes-events` SNS created as it was causes an error when trying to `update_infrastructure:elife-libero-editor prod` after running the `update_infrastructure :elife-libero-editor staging` (I had to request Daniel H to run these for me as the newer version of OSX don't like installing builder deps). The error thrown was: `elife-production-processes-events already exists in stack arn:aws:cloudformation:us-east-1:512686554592:stack/elife-libero-editor--staging/c21115f0-3879-11eb-b050-0e6115325dc3`

I'm hoping that the change I've made would mean someone could bring up staging which would create the SNS then when they bring up prod the SNS required by its instance SQS queue will already be available. 

Maybe I've came at this from the wrong direction but both @NuclearRedeye and I had looked for docs / existing examples for something similar and this as the best we could come up with.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1711,7 +1711,7 @@ elife-libero-editor:
                     - elife-production-processes-events
         docdb: {}
     aws-alt: 
-        production-processes:
+        staging:
             # This bucket is required by editor but was created manually pre-editor so can't be configured via builder.
             # It should also notify the bus-elife-libero-editor on Put events
             # s3:
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1704,11 +1704,6 @@ elife-libero-editor:
         ec2: false
         s3:
             "{instance}-elife-libero-editor": {}
-            # This bucket is required by editor but was created manually pre-editor so can't be configured via builder.
-            # It should also notify the bus-elife-libero-editor on Put events
-            # "elife-production-processes":
-        sns:
-            - "elife-production-processes-events"
         sqs:
             # Monitor the production process bucket for new deposits from Exeter.
             elife-libero-editor-import--{instance}:
@@ -1716,6 +1711,13 @@ elife-libero-editor:
                     - elife-production-processes-events
         docdb: {}
     aws-alt: 
+        production-processes:
+            # This bucket is required by editor but was created manually pre-editor so can't be configured via builder.
+            # It should also notify the bus-elife-libero-editor on Put events
+            # s3:
+            #   "elife-production-processes":
+            sns:
+                - "elife-production-processes-events"
         prod: 
             s3:
                 "{instance}-elife-libero-editor":


lsh-0:
> We only need a single elife-production-processes-events SNS topic, 
> which both of our prod and staging elife-libero-editor-import--{instance} SQS queues will be subscribed to. 

Understood.

> Having the elife-production-processes-events SNS created as it was causes an error when trying to 
> update_infrastructure:elife-libero-editor--prod after running the update_infrastructure:elife-libero-editor--staging 
> The error thrown was: elife-production-processes-events already exists in stack arn:aws:cloudformation:us-east-1:512686554592:stack/elife-libero-editor--staging/c21115f0-3879-11eb-b050-0e6115325dc3

I've investigated and I think I have a handle on the current state.

The topic configuration was added, then `update_infrastructure` was called on the *staging* instance and the SNS topic for "elife-production-processes-events" was attached to it.

You went to run `update_infrastructure` on the *prod* instance and it failed with the 'already exists' error.

This was because that topic was using a fixed name and in the common project configuration.

You moved the topic config under `staging` and this is what I'm looking at now.

> I'm hoping that the change I've made would mean someone could bring up staging which would create the SNS then when they bring up prod the SNS required by its instance SQS queue will already be available.

Correct. `staging` would have to be created first and then `prod`. Should `staging` ever be destroyed `prod` would have the target of it's subscription destroyed.

Just a suggestion, but I recommend pushing that SNS topic under the `prod` configuration. The migrations to get there would be:

1. commit this current PR
2. move that SNS configuration from `staging` to `prod` in another PR
3. `update_infrastructure` on `staging` to destroy the SNS topic
4. `update_infrastructure` on `prod` to create the SNS topic
5. `update_infrastructure` on `staging` again to re-subscribe `staging` to the topic 
6. merge

> Maybe I've came at this from the wrong direction but both @NuclearRedeye and I had looked for docs / existing examples for something similar and this as the best we could come up with.

There have been many good intentioned attempts at documentation. I'm not sure if there is a magic bullet. I suspect it would require a lot of time and effort.

> (I had to request Daniel H to run these for me as the newer version of OSX don't like installing builder deps). 

Open a ticket, please. I don't use macOS but others do and may benefit from a fix. I have a dodgy little mac mini I can do some adhoc testing on if need be.

lsh-0 (approved):

Cooryd:
I've raised https://github.com/elifesciences/issues/issues/6426 to try to fix my issue with running builder. I'll raise another issue to migrate the SNS from staging to production. Thanks

lsh-0:
> I'm hoping once this is in and correct it can serve as a guide to doing similar things in the future.

Don't count on it. Consider what you did here:

> Maybe I've came at this from the wrong direction but both @NuclearRedeye and I had looked for docs / existing examples for something similar and this as the best we could come up with.

What was the first place you looked? And does anything exist there now that this is in? How about the second place you looked? I'm not being mean, just pointing out that crossing your fingers and hoping that other people or our future selves won't have this problem is naive. We *all* do this *all* the time. We need to add DODs to our task like "update docs" or "write a guide" or "improve the UI", etc. It's extra work to do and once our problem is solved there is no impetus or momentum or enthusiasm to do it, so it doesn't get done.

PR #761, opened 2021-01-29
py3 docker, installs wheel before installing requirements.

lsh-0:
greenlet dependency is failing to find gcc and build. It should be installed via wheel.


lsh-0:
that isn't the problem, the problem is that pip is ignoring the `greenlet==VERY PRECISE VERSION` and is installing the latest bleeding edge version. This is `gevent` being a bastard again, fml.

PR #763, opened 2021-02-02
accepted-submission-cleaning

lsh-0:
* accepted-submission-cleaning, adds s3 and sns definitions
* elife-bot, adds elife-bot subscription to s3 bucket events for "elife-accepted-submission-cleaning"

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1923,3 +1923,16 @@ bioprotocol:
     vagrant:
         ports:
             1280: 80
+
+accepted-submission-cleaning:
+    description: see https://github.com/elifesciences/bus
+    default-branch: null
+    domain: null
+    intdomain: null
+    aws:
+        ec2: false
+        s3:
+            elife-accepted-submission-cleaning:
+                deletion-policy: retain
+        sns:
+            - elife-accepted-submission-cleaning-s3-events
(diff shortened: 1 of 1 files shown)


gnott:
This looks pretty good to me with my limited knowledge of builder, we have one S3 bucket, and each `elife-bot` environment will have its own SQS queue, reading from the same SNS topic. The ones in prod are most important, but it will also mean continuous running `--continuum-test` environment will also have a duplicate of all the same S3 notifications, yes? It's not a requirement, I'm more concerned whether it would cause any adverse results.

lsh-0:
> will also have a duplicate of all the same S3 notifications, yes?

correct. I can switch the subscription for elife-bot to prod only if you'd like?

Perhaps when the submission-cleaning project is more fleshed out we can set up some end2end tests and add subscriptions to other environments.

gnott:
It's hard to know at this stage what is best right now, in future it may need an adjustment for other environments when we're futher along. I may be able to disable outputs on non-prod environments if there's no output bucket specified in the `elife-bot` settings, and that might be a way to avoid too much junk from accumulating on environments we are not really using for ad-hoc or end2end testing yet.

lsh-0:
I've moved the sqs definition to prod only for now. I think we're on the same page but if this isn't what you wanted let me know.

gnott:
Yes, thanks Luke, I'm in agreement if we are precise and incremental in how the configuration is set we will be is a good place.

lsh-0:
updating `elife-bot--prod` now with it's new SQS subscription ...

PR #804, opened 2021-06-23
chore: add new SQS queue for incoming ERA articles

NuclearRedeye:


lsh-0 (approved):
without any context this seems fine. All environments will see the new SQS so you'll have to do something like:

./bldr update_infrastructure:elife-bot--ci
./bldr update_infrastructure:elife-bot--continuumtest
./bldr update_infrastructure:elife-bot--end2end
./bldr update_infrastructure:elife-bot--prod

lsh-0:
I'm happy to do the updates Fri if you're still squeamish after the recent journal event

NuclearRedeye:
> I'm happy to do the updates Fri if you're still squeamish after the recent journal event

I'll give it a try myself later today but this time via using the builder vm you pointed me at. If I encounter any issues then we can discuss tomorrow.

lsh-0:
elife-bot doesn't use any Terraform configuration so it should work without the VM

I use the VM mostly because the version of Terraform is pinned and the plugin versions don't make it spaz out. I have some PHP build dependencies setup in there as well by the looks of it. Definitely get it setup and give it a go though, I think I've been the only one using it so far. Let me know if anything is missing.

Formula here: https://github.com/elifesciences/builder-formula

PR #839, opened 2021-10-28
fix: default root hdd size should be 40GB not 20GB for eks worker nodes

NuclearRedeye:
Progresses elifesciences/data-hub-issues#276

This PR is to increase the default size of the root disk/partition from 20GB to 40GB for the kubernetes workers/nodes to address the issues with importing large datasets in DataHub failing


NuclearRedeye:
@lsh-0 Thanks, you're approach is much more sensible hence I'll close this off and get yours merged.

PR #842, opened 2021-11-02
elife-bot, adds new bucket with a queue to put s3 events on to.

lsh-0:
* adds new bucket "elife-bot-accepted-submission-cleaning-output" with an env prefix
* adds new sqs for s3 notifications to go into

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -838,9 +838,6 @@ elife-bot:
             "{instance}-elife-bot-decision-letter-input":
             "{instance}-elife-bot-decision-letter-output":
             "{instance}-elife-bot-accepted-submission-cleaning-output":
-                deletion-policy: delete
-                sqs-notifications:
-                    "{instance}-elife-bot-accepted-submission-cleaning-output-s3-events": {}
         sqs:
             # gnott@2021-08: poa-incoming-queue was configured to avoid the rigidly timed PoA packaging process.
             # now the S3 notifications go into the queue, and elife-bot reads the messages at four specific times daily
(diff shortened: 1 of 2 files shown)


projects/elife.yaml
```diff
@@ -960,6 +961,8 @@ elife-bot:
                 "{instance}-elife-bot-decision-letter-input":
                     sqs-notifications:
                         incoming-queue: {}
+                "{instance}-elife-bot-accepted-submission-cleaning-output":
```
gnott, line 964: I may not understand the sections of the yaml file, and I wonder if this is a duplicate of the one created in line 840 above? Or maybe this is how you are specifying the different `deletion-policy`? More of a question I guess whether I'm reading it correctly.

gnott (approved):
Naming looks good. Having a bucket for each environment is probably smart, I don't know how the non-`prod` ones may be used, but they could be useful in some testing on `continuumtest` of the ingest workflow, or if there are `end2end` tests added later on. I don't anticipate any testing of the SQS queue for whether messages appear there, since they're consumed outside of this infrastructure.

projects/elife.yaml
lsh-0: There is the 'default' project configuration and there are 'alternate' (aws-alt) configurations that can selectively override the defaults. 

in this case I've created an s3 bucket for all instances of elife-bot called `{instance}-elife-bot-accepted-submission-cleaning-output` with `{instance}` replaced with the instance ID we give to the Cloudformation stack when we create one, so `prod--elife-bot-accepted-submission-cleaning-output` for example.

the default `deletion-policy` for all buckets is `delete` (we don't need to repeat `deletion-policy: delete` over and over but we have). In the `prod` instance of elife-bot we want to keep the data in any buckets typically. Perhaps not in this case?

lsh-0: > Or maybe this is how you are specifying the different deletion-policy?

tl;dr: yup

lsh-0:
>  I don't anticipate any testing of the SQS queue for whether messages appear ther

good point, I'll shift the queue stuff into `prod` then.

PR #760, opened 2021-01-28
Replace ruby brew install with bash and remove brew  cask tap

Cooryd:
https://raw.githubusercontent.com/Homebrew/install/master/install has been deprecated so this replaces with the suggested bash install of `brew`. `brew cask` has now been replaced with `brew install --cask` and should no longer need to be tapped with `brew tap caskroom/cask`. These changes got me to a `  ◕ ‿‿ ◕   all done` output but I would like to see it work on one other Catalina/ Big Sur machine before merging.

PR #856, opened 2021-12-01
fix: brew cask is now included in homebrew

NuclearRedeye:
Removed the OSX specific `brew cask` check, as this is now included in homebrew by default.


lsh-0:
there is a similar PR here https://github.com/elifesciences/builder/pull/760

lsh-0 (approved):

NuclearRedeye:
> there is a similar PR here #760

Yeah, I recall putting in some work on that but it still had issues depending on which version of OSX you were using. It worked for me, but not for others. There is still some useful work in that though so I'll try and spend some time to clean it up a little and see if we can get it merged.

PR #865, opened 2022-01-10
Create a container for builder that works on OSX

NuclearRedeye:


docker-compose.yaml
```diff
@@ -0,0 +1,19 @@
...
+    image: builder:${TAG-local}
+    container_name: ubuntu-builder
+    stdin_open: true
+    tty: true 
+    environment: 
+      - TZ='Europe/London'
+    volumes:
+      - '/etc/timezone:/etc/timezone:ro'
+      - '/etc/localtime:/etc/localtime:ro'
+      - ${CUSTOM_SSH_KEY-~/.ssh/id_rsa}:/root/.ssh/id_rsa:ro
```
NuclearRedeye, line 16: Note: I wanted to follow up on these mounts. If the files don't exist on the host then docker will create them as directories which causes a lot of bother and isn't the desired behaviour. I need to research if there is a better way to manage this or get the process to fail if these files (or their overrides) don't exist on the host.

lsh-0:
took a quick look, I know it's just your personal working copy but to get it merged in I would recommend:

* renaming `Dockerfile` to `Dockerfile.osx` or similar.
* ditching the docker-compose file in favour of a bash script. It's a long invocation but you're only dealing with a single container.
  - (I get it's convenient but tying docker-compose to builder would be yet-another-technology on top of a largish stack already)

The rest seems ok. Some typos but no deal breakers.

I guess I'm looking to carve out a well defined path in `builder` for mac users to get up and running quickly without too many special instructions/caveats/exceptions. `update.sh` is used for installing builder and it calls out to `prerequisites.py` that does the checks for OSX. We should take a look at that as well, but not here. Docker support should be available but if we don't persevere with a native installation for mac users then wrinkles won't get ironed out and builder will be less robust.

I really do need to revisit *all* of the documentation we have written for builder. It's just too much and too old.

lsh-0:
oh - and I don't say this enough to everybody that takes the time to open a PR or a bug or raise an issue - but I really do appreciate your effort. A ticket by itself just languishs until a person has time whereas a broken PR begs to be fixed.

docker-compose.yaml
NuclearRedeye: With the new script, this is actually handled so happy with that.

NuclearRedeye:
@scottaubrey I updated the PR/branch to address the feedback from @lsh-0, not sure if you want to add your changes for supporting ARM architectures if they're ready? If not, don't worry. We can get this merged and iterate on it some more later down the line.

NuclearRedeye:
CI failures are unrelated to my changes, looks like a dependency needs a newer version of Python 3.

scottaubrey:
> @scottaubrey I updated the PR/branch to address the feedback from @lsh-0, not sure if you want to add your changes for supporting ARM architectures if they're ready? If not, don't worry. We can get this merged and iterate on it some more later down the line.

@NuclearRedeye I've pushed the bare minimum to get it working on ARM mac - which is to say forcing it to build and use an intel image due to the intel-only dependencies. Further ARM-native work would require python 3 and newer libraries, newer terraform providers, and probably a whole host more that I didn't come across for updating the eks clusters. 

I'd also still like to introduce support for the SSH-Agent on the host (rather than mounting/copying in the SSH key), but the changes run deeper into builder, some unrelated to this container image specifically.

lsh-0:
I can recreate the Python3 error that's appearing - it's brand new.

So Python3.6 exited it's LTS period before Ubuntu 18.04 exited *it's* LTS period, so Ubuntu will keep Python3.6 around and patched but the Python org have essentially washed their hands of it.

This state of affairs in programming languages really gets my goat.

Until we migrate from 18.04 LTS to 22.04 LTS, we must stick with Python3.6.

lsh-0:
> I'd also still like to introduce support for the SSH-Agent on the host (rather than mounting/copying in the SSH key)

Sure, just create a ticket for that.

Copying the ssh key into Vagrant machines was what we (I) ended up doing because ssh-agent+vagrant+os was so unreliable back when builder was still under heavy development. There are probably some mentions of support, it may even work under some conditions, but I wouldn't count on it. I know @nlisgo has had ssh-agent pain the past - he's on a mac as well I think.

NuclearRedeye:
Just got this rebased, are you happy for this to be merged @lsh-0?

lsh-0:
I'll take another look.

container.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
lsh-0, line 1: The name "container.sh" could be improved, especially since this looks to be osx-specific. Maybe `mac-bldr.sh` ?

docs/osx.md
```diff
@@ -0,0 +1,27 @@
+# Running on OSX
+
+Due to the underlying dependencies, it can be a bit hit and miss using `builder` on OSX. This can be worked around by running `builder` inside a container with the correct dependencies installed.
+
+```
+./container.sh
```
lsh-0, line 6: rename

PR #874, opened 2022-02-10
update cluster size for flux-prod

scottaubrey:


lsh-0:
Please don't merge broken builds.

NuclearRedeye:
@lsh-0 Apologies, this was me. I wouldn't typically do this especially in a repository where I'm not the maintainer but my reasoning in this instance was that we needed to change some values to progress some other work and I was certain these changes weren't the cause of the failure. 

On reflection, merging these _wasn't_ critical to progressing as we could have proceeded with local changes but I was very keen to make sure that state of the master branch in `builder` represented the current system state.

lsh-0:
> I was very keen to make sure that state of the master branch in builder represented the current system state.

I can appreciate that.

It's a design issue with builder - the project configuration is tied to builder's code. The current work with modelling the users/groups/permissions is also configuration and in this instance it *can't* live alongside builder because it's private, so I'm thinking about the best way to accommodate this...

I have this project https://github.com/elifesciences/builder-projectlib (that isn't complete) but it's intent is to be a separate library for describing, validating and testing project configuration. It could be integrated with builder, or it could be used by a private repo that would house elife configuration. 

Config changes would happen there and in this instance your changes would then be uncoupled from builder and just go through simpler validation checks.

`builder-private` would be a natural home, but it has a lot of credentials in there already and I think it just needs to be 'not-public' rather than 'super-secure'.

Food for thought.

PR #887, opened 2022-02-17
Update shell container to python3

scottaubrey:
Updated the shell container used to create a linux environment to run builder on OS X to use Python 3.

Also updated the `update.sh` script to use the `python3` binary name, which looking at other examples in repo, is the expected binary name too.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -2,7 +2,7 @@
 
 set -e # everything must pass
 
-python .prerequisites.py "$@"
+python3 .prerequisites.py "$@"
 
 # remove any old compiled python files
 find src/ -name '*.pyc' -delete
--- a/Dockerfile.osx
+++ b/Dockerfile.osx
@@ -1,7 +1,7 @@
 from ubuntu:20.04
 
 WORKDIR /builder
-RUN apt update && DEBIAN_FRONTEND=noninteractive apt install -y tzdata build-essential software-properties-common git python virtualenv vagrant awscli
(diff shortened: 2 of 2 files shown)


scottaubrey:
FYI @NuclearRedeye

NuclearRedeye:
Tested locally and works fine for me. @lsh-0 should take a look to the. `update.sh` script though just to comment on its compatibility before we get this merged.

lsh-0:
> @lsh-0 should take a look to the. update.sh script though

Yup, I don't want to be a bottleneck but please cc me in on all changes to builder. I don't necessarily need to review trivial changes, it just helps me keep my mental model current.

update.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e # everything must pass
 
-python .prerequisites.py "$@"
+python3 .prerequisites.py "$@"
```
lsh-0, line 5: Nice catch, I overlooked this in the recent py2-removal. 

On Ubuntu at least `python` will default to `python2`. I wouldn't be surprised if osx did the same. I run Arch and `python` points at the most recent version of python installed:

```
$ python --version
Python 3.10.1
```

lsh-0 (approved):
@scottaubrey , I'm happy to merge any changes to builder that are waiting on my approval and don't need any further feedback.

You can ask me to merge it or I can do it automatically.

I won't merge draft PRs but happy to review them.

PR #891, opened 2022-02-23
project, moves 'type' and 'ports' under 'ec2'

lsh-0:
This change has been a long time in coming. It goes back to the original project definitions where there were only ec2 instances and no other aws resources, so the ec2 'type' and 'ports' were at the top level.

Eventually more configuration came along and it got moved under the 'ec2' heading and other resources became supported and they got their own headings but 'type' and 'ports' were never moved.

This:

* updates all of the project configuration, including test configuration, shifting 'ports' and 'type' under 'ec2'
* issues a warning if 'ports' or 'type' are detected under 'aws'.
  - has to be a warning because it's possible old context data has been merged over the top.
* removes the 'additional-projects' test fixture with a single project in the 'eu' region.
  - project was moved into the main `dummy-project.yaml` file and the references fixed up.
* updated test fixtures


lsh-0:
fyi @NuclearRedeye

PR #899, opened 2022-02-25
Add autoscaler policy to eks clusters

scottaubrey:
This (optional for now) IAM policy allows an autoscaling component of kubernetes to alter the AutoScalingGroup from within the cluster, based on requested cpu/mem for kubernetes resources.

This will eventually find it's way on to the production cluster, but for now is just enabled on `flux-test` cluster to see if it works, and observe it's behaviour.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -201,6 +201,7 @@ defaults:
                 # Route53 hosted zone to create subdomains on
                 #domain-filter: elifesciences.org
             efs: false
+            autoscaler-policy: false
         cloudfront:
             # cloudfront defaults only used if a 'cloudfront' section present in project
             subdomains: []
@@ -2417,4 +2418,3 @@ firewall:
                             - UNIXShellCommandsVariables_QUERYARGUMENTS
                             - UNIXShellCommandsVariables_BODY
                         excluded: []
-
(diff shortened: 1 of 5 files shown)


projects/elife.yaml
```diff
@@ -2227,6 +2227,7 @@ kubernetes-aws:
                 # but not the ExternalDNS chart release
                 external-dns:
                     domain-filter: "elifesciences.org"
+                autoscaler-policy: true
```
lsh-0, line 2230: When introducing new keys ensure they are mentioned in the `defaults` section with a sensible default. Add a comment if it's purpose isn't obvious.

lsh-0: new value should probably be mirrored in the `./src/tests/fixtures/projects/dummy-project.yaml` file as well

scottaubrey: Thanks, added. Also added a new test fixture project and a test for the output while in there.

scottaubrey:
@lsh-0 are you OK for this PR to merge now with the extra defaults and tests?

lsh-0:
yeah, man - you addressed my feedback and didn't introduce any significant new changes, no need to block this waiting on me.

I mentioned it before, but if you ever want me to merge something after review for the sake of expediency just say so. I'll otherwise leave it to you to merge.

PR #905, opened 2022-03-04
fix(eks): allow describe actions to be performed on all resources, so ASG auto-discovery can happen

scottaubrey:
I created a more restrictive a policy than the getting started suggested (as I didn't want the test cluster updating the prod ASGs) However it doesn't allow the autoscaler to list ASGs to discover (as it only has access to one).

This creates a new policy allowing the worker node permission to query all ASGs, but still keeps it only altering the ASG it belongs to.

Helpful documentation is [here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#full-cluster-autoscaler-features-policy-recommended).

### DIFF
--- a/src/buildercore/terraform.py
+++ b/src/buildercore/terraform.py
@@ -835,6 +835,14 @@ def _render_eks_workers_role(context, template):
                             "autoscaling:DescribeAutoScalingInstances",
                             "autoscaling:DescribeLaunchConfigurations",
                             "autoscaling:DescribeTags",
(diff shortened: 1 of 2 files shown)


NuclearRedeye (approved):

scottaubrey:
Thanks for the test changes @lsh-0 , thanks for the second pair of eyes @NuclearRedeye

PR #910, opened 2022-03-14
scripts, quieter commands, verbose command line arguments

lsh-0:


lsh-0:
just a bit sick and tired of wading through progress bar output

PR #917, opened 2022-03-16
Increase ec2 polling interval

lsh-0:


lsh-0:
fyi @NuclearRedeye 

Should help address throttling issues in end2end testing. I see builds fail often when I'm racing around kicking jobs in Jenkins or opening many prs at once.

NuclearRedeye:
Good stuff. I've had it a few times my self with Jenkins where it times-out just starting a stack and waiting for it to reach a 'ready state'.

lsh-0:
That sounds like a different problem. The job will typically fail with a 'Request Failed: traffic throttled' type message.

I have noticed the exact problem you speak of though: https://alfred.elifesciences.org/job/process/job/process-stop-stacks-long-term/

Not sure if it's an RDS problem or a builder problem or a Jenkins problem. It's fairly rare right now but we'll keep our eye on it.

PR #933, opened 2022-04-06
trop.py, rds, adds support for specifying an snapshot

lsh-0:
Adds support for 'rds.snapshot-id' to a project's rds definition to create an instance from a snapshot.

- [x] snapshot dbname override is ~necessary~ available when specifying a snapshot. 
- [x] review


lsh-0:
I've used this and while it appeared to work, it only created the schema, not the data, so I'm a bit confused as to why. I've opened a ticket with support.

src/buildercore/cfngen.py
```diff
@@ -385,18 +386,25 @@ def build_context_rds(pdata, context, existing_context):
...
-    # TODO: shift the below under a 'rds' key
+    context['rds'] = pdata['aws']['rds']
+    context['rds'].update({
+        'deletion-policy': deletion_policy,
+    })
+
+    auto_rds_dbname = slugify(stackname, separator="") # lax--prod => laxprod
+    existing_rds_dbname = existing_context.get('rds_dbname')
+    override = lookup(pdata, 'aws.rds.db-name', None)
+    rds_dbname = override or existing_rds_dbname or auto_rds_dbname
```
lsh-0, line 397: I think `existing` should come first ...

src/buildercore/trop.py
```diff
@@ -542,6 +542,21 @@ def render_rds(context, template):
         data['StorageEncrypted'] = True
         data['KmsKeyId'] = lu('rds.encryption') if isinstance(lu('rds.encryption'), str) else ''
 
+    # use existing snapshot to create instance:
+    # - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-dbsnapshotidentifier
+    if lu('rds.snapshot-id', None):
+        data['DBSnapshotIdentifier'] = lu('rds.snapshot-id')
+        delete_these = [
+            # in use
+            "DBName", "MasterUsername",
```
lsh-0, line 551: add note that even though the dbname is preserved in the buildvars it won't be available in the template (or even necessarily consistent with what is in the snapshot).

src/buildercore/cfngen.py
lsh-0: previously we've been relying on the consistent behaviour of the slugify function to always slugify the given value the exact same way. Nothing wrong with that, but if we have access to the previously slugified and in-use dbname, we should prefer that.

lsh-0:
> I've used this and while it appeared to work, it only created the schema, not the data, so I'm a bit confused as to why. I've opened a ticket with support. Until I know why it failed to port the data I won't merge this.

it created the database, schema and data fine, however it created with the snapshot's database name, which was `laxprod`, as well as the app creating it's own database with the expected database name and schema but obviously no data.

Solution is to ensure the buildvars have the right database name in them so the app doesn't get confused.

projects/elife.yaml
```diff
@@ -85,13 +85,21 @@ defaults:
...
+            # if snapshot-id is used, then the name of the database in the snapshot should be used.
+            db-name: null
             # Standard for backward compatibility
             # https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html
             storage-type: gp2 # standard|gp2|io1
             backup-retention: 28 # days
             # if rds.params are specified, a custom db parameter group is created
             params: []
             encryption: false
+            # create a new database from the snapshot of another one.
```
lsh-0, line 99: "a new database instance"

projects/elife.yaml
```diff
@@ -85,13 +85,21 @@ defaults:
...
+            # updates to the stack do *not* roll the database back to the snapshot, but the snapshot-id *must* be present,
+            # otherwise a new database is created.
```
lsh-0, line 101: a new database instance

src/buildercore/cfngen.py
lsh-0: test

lsh-0: > I think existing should come first ...

no, `override` is explicit, `existing` is implicit and `auto` is even more implicit.

projects/elife.yaml
```diff
@@ -85,13 +85,21 @@ defaults:
             allow-major-version-upgrade: False # do not alter unless you're doing a major version upgrade
             type: db.t2.small
             storage: 5 # GB
+            # name of the database to be created, *not* the RDS Instance ID
```
lsh-0, line 88: fullstop

src/buildercore/cfngen.py
lsh-0: move this whole dbname wrangling into `trop.py`

PR #934, opened 2022-04-08
projects, firewall, adds more bots to blacklist.

lsh-0:
* projects, firewall, adds more bots to blacklist.
* removes rule for 'distributed slow bow' - waf rule wasn't ever effective.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -2248,7 +2248,8 @@ firewall:
                 custom-rules:
                     - elife-Whitelist.json
                     - elife-Blacklist.json
-                    - elife-DistributedSlowBot.json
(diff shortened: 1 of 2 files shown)


lsh-0:
fyi @gnott @NuclearRedeye @scottaubrey 

Just a heads up, I've added more bots to the WAF blacklist. 

This only affects traffic hitting the `journal--prod` load balancer. 

The blacklist at the CDN (Fastly) fronting the `journal--prod` load balancer is almost empty at the moment (two random bots blocked).

If for whatever reason you've built an app using Python's urllib or aiohttp libraries and you *haven't* set a user-agent, you're going to get a 400 or 403 response. We're getting a deluge of these at the moment.

PR #939, opened 2022-04-20
buildvars, fixes to 'force' task

lsh-0:
* buildvars, 'force' can now do nested updates of buildvars.
* 'value' parameter to 'force' now attempts to coerce to none/bool/integer values first.

- [x] review

### DIFF
--- a/src/buildercore/utils.py
+++ b/src/buildercore/utils.py
@@ -192,15 +192,23 @@ def updatein(data, path, newval, create=False):
     for example: `updatein({"foo": {"bar": "boo"}}, "foo.bar", "baz")` => `{"foo": {"bar": "baz"}}`
     use `create=True` to create the path if it doesn't already exist.
     for example: `updatein({}, "foo.bar", "baz")` => `{"foo": {"bar": "baz"}}`"""
-    path_bits = path.split('.')
-    bit, rest = path_bits[0], path_bits[1:]
+
+    bit, rest = path, None
+    if isinstance(path, str):
+        path_bits = path.split('.')
+        bit, rest = path_bits[0], path_bits[1:]
+
     if not rest:
         # we've come to the end of the path
         data[bit] = newval
-        return newval
+        return data
+
     if create and bit not in data:
         data[bit] = {}
-    return updatein(data[bit], ".".join(rest), newval, create)
+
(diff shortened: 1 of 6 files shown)


src/buildercore/utils.py
```diff
@@ -188,8 +188,10 @@ def file_doesnt_exist():
     call_while(file_doesnt_exist, interval=2, update_msg="waiting for /tmp/foo to be created", done_msg="/tmp/foo found")
 
 def updatein(data, path, newval, create=False):
```
lsh-0, line 190: tests?

src/buildvars.py
```diff
@@ -99,20 +99,22 @@ def _fix_single_ec2_node(stackname):
 
     stack_all_ec2_nodes(stackname, (_fix_single_ec2_node, {'stackname': stackname}), username=BOOTSTRAP_USER)
 
-# TODO: deletion candidate. can only ever do a shallow update
 @requires_aws_stack
 def force(stackname, field, value):
-    "replace a specific key with a new value in the buildvars for all ec2 instances in stack"
+    """replace a specific key with a new value in the buildvars for all ec2 instances in stack.
```
lsh-0, line 104: "replaces value at 'field' with given `value` for all buildvars on all ec2 instances on given `stackname`."

src/buildvars.py
```diff
@@ -99,20 +99,22 @@ def _fix_single_ec2_node(stackname):
...
+    can only be used to *replace* an *existing* key, not create new ones.
+    `field` can be a dotted path for targeting values inside nested maps.
+    For example: force(lax, elb.stickiness, true)`
```
lsh-0, line 107: missing tick

src/utils.py
```diff
@@ -145,3 +145,20 @@ def find_region(stackname=None):
     except core.MultipleRegionsError as e:
         print("many possible regions found!")
         return _pick('region', e.regions())
+
+def coerce_string_value(value_str):
+    "attempts to coerce given `value_str` to a None, then a boolean, then an integer, ultimately returning the value as-is"
```
lsh-0, line 150: > ultimately returning the value as-is

returning the given value if coercion not possible

lsh-0: test

PR #954, opened 2022-05-23
Return default None when rds.num-replacements isn't set on stack state context

scottaubrey:
Related to  https://github.com/elifesciences/issues/issues/7430

I'm guessing for preexisting state stored in S3 it doesn't contain this data, therefore defaulting to None (and thus no suffix) is the right behaviour.


nlisgo (approved):

lsh-0:
> I'm guessing for preexisting state stored in S3 it doesn't contain this data, therefore defaulting to None (and thus no suffix) is the right behaviour.

Exactly. I'll merge this, kick the broken tests and come up with a unit test.

PR #975, opened 2022-07-08
Add autoscaler policy for flux-prod k8s cluster

scottaubrey:


lsh-0:
The instance Jenkins runs on has been upgraded and I've noticed a few changes for `builder` fail, presumably because of language changes in Python3.8. I'm looking into these now but it might fail this build.

scottaubrey:
> The instance Jenkins runs on has been upgraded and I've noticed a few changes for `builder` fail, presumably because of language changes in Python3.8. I'm looking into these now but it might fail this build.

No worries. I was trying to do this change this morning as the rest of the team are enjoying fridays off through summer, but it's not urgent and can be done next Friday. Thanks for the heads up @lsh-0

lsh-0:
I'll give it a kick when the bug is fixed

lsh-0:
You're good to go.

PR #989, opened 2022-07-29
Register EKS OIDC as an identity provider in IAM 

scottaubrey:
# Description

Currently, we set permissions for cluster resources in IAM based on the role given to the worker nodes. This works fine, but in effect gives all services running on the clusters the same level of access, even if only 1 service needs it.

Current IAM policy examples that technically all services have access to:
- alter desired-capacity on the ASG (used by `autoscaler`  when it decides to scale up or down)
- create and alter records in Route53 (used by `external-dns` to register new Ingress or services)

For my 10% time, I investigated and wrote this change to add the cluster as an identity provider to IAM. It should then allow us to define IAM policies by cluster service account, rather than by the nodes. If this works, I can then alter the existing policy attachments to be only to the correct service account, reducing the security exposure.

It also will help me feel much more comfortable about installing cluster services that interact with AWS services, as we are not in effect allowing all services the same privileges, and can keep the policies very fine-grained.

# Things of note:

- This requires a newer version of aws provider, which I've tested works on a state refresh of the flux-test cluster, but I'm not sure if it's used anywhere else. It being a minor version should be backwards compatible, and still supported on TF v0.11.
- I've added a new provider, required to get the fingerprint of the OIDC cert. Handily, the earliest version is also the last supported on TF v0.11, not requiring a TF upgrade.
- This is optional and defaulting to false for now, to allow me to iterate on it on just the test cluster to start with. I intend to remove the option and make it the default to register a cluster with IAM when I've figured out the other bits to the equation.

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -221,6 +221,7 @@ defaults:
                 #domain-filter: elifesciences.org
             efs: false
             autoscaler-policy: false
+            iam-oidc-provider: false
         cloudfront:
             # cloudfront defaults only used if a 'cloudfront' section present in project
             subdomains: []
--- a/src/buildercore/terraform.py
+++ b/src/buildercore/terraform.py
@@ -1001,7 +1001,7 @@ def _render_eks_iam_access(context, template):
 
     template.populate_resource('aws_iam_openid_connect_provider', 'default', block={
         'client_id_list': ['sts.amazonaws.com'],
-        'thumbprint_list': ['$data.tls_certificate.cert.certificates[0].sha1_fingerprint'],
+        'thumbprint_list': ['${data.tls_certificate.oidc_cert.certificates.0.sha1_fingerprint}'],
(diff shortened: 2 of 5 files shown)


lsh-0 (approved):
> This requires a newer version of aws provider, which I've tested works on a state refresh of the flux-test cluster, but I'm not sure if it's used anywhere else. It being a minor version should be backwards compatible, and still supported on TF v0.11.

From a skim of the logic only EKS is an AWS resource here. I haven't touched this code much/at all, but if I were to, I'd figure out some way of tying resources to providers to address [this comment](https://github.com/elifesciences/builder/blob/1f31e65c71ff64f4955ad6534803c40058d68509/src/buildercore/terraform.py#L1234-L1236) and make the dependencies clear.

For now if it only affects EKS, and if you're updating EKS and reporting all is fine (and are in a position to fix things if they're not), then PR seems fine.

PR #993, opened 2022-08-09
cloudformation stack template is fetched if not exists

lsh-0:
- [x] cloudformation, if stack not found on disk, stack is downloaded, written to disk and returned.
- [x] review

### DIFF
--- a/src/buildercore/cloudformation.py
+++ b/src/buildercore/cloudformation.py
@@ -22,6 +22,26 @@ def render_template(context):
     ensure('aws' in context, msg, ValueError)
     return trop.render(context)
 
+def download_template(stackname):
+    "downloads and returns the cloudformation stack for the given `stackname` from AWS."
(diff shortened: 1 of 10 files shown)


src/buildercore/cloudformation.py
```diff
@@ -22,6 +22,27 @@ def render_template(context):
...
+    conn = core.boto_conn(stackname, 'cloudformation', client=True)
+    data = conn.get_template(StackName=stackname)['TemplateBody']
+    return json.dumps(data)
+
+def write_template(stackname, contents):
+    "writes a json version of the python cloudformation template to the stacks directory"
+    output_fname = core.stack_path(stackname)
+    with open(output_fname, 'w') as fp:
+        fp.write(contents)
+    return output_fname
```
lsh-0, line 36: add `info` log message when we write files to disk

PR #995, opened 2022-08-17
cfn.ssh and cfn.owner_ssh now allow node information from stackname.

lsh-0:
so this is now possible: `./bldr ssh:lax--prod--2` as well as `./bldr ssh:lax--prod,2`

you'll get a warning if you try something like this: `./bldr ssh:lax--prod--2,1` and the explicit node parameter (1) will be preferred over the implicit node parameter (2) from the stackname.

- [ ] does this interfere with bringing up a stopped stack? (yes)
- [ ] does this still prompt when ssh'ing into a multi-node cluster? (yes)
- [ ] ...
- [ ] review

### DIFF
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -499,15 +499,27 @@ def mk_stackname(project_name, instance_id):
     return "%s--%s" % (project_name, instance_id)
 
 def parse_stackname(stackname, all_bits=False, idx=False):
-    "returns a pair of (project, instance-id) by default, optionally returns the cluster (node) id if all_bits=True"
-    if not stackname or not isstr(stackname):
-        raise ValueError("stackname must look like <pname>--<instance-id>[--<cluster-id>], got: %r" % stackname)
-    # https://docs.python.org/2/library/stdtypes.html#str.split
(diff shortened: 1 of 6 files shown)


src/buildercore/core.py
```diff
@@ -499,15 +499,27 @@ def mk_stackname(project_name, instance_id):
...
 
 def parse_stackname(stackname, all_bits=False, idx=False):
-    "returns a pair of (project, instance-id) by default, optionally returns the cluster (node) id if all_bits=True"
-    if not stackname or not isstr(stackname):
-        raise ValueError("stackname must look like <pname>--<instance-id>[--<cluster-id>], got: %r" % stackname)
-    # https://docs.python.org/2/library/stdtypes.html#str.split
-    bits = stackname.split('--', -1 if all_bits else 1)
+    """returns a pair of (project, instance-id) by default.
+    returns a triple of (project, instance-id, node) if `all_bits` is `True`.
+    returns a map with keys `project_name`, `instance_id` and `cluster_id` (node) if `idx` is `True`."""
```
lsh-0, line 504: if `idx` and `all_bits` is `True`

PR #1002, opened 2022-09-02
Replace some `python` binary calls with `python3`

scottaubrey:
Some local fixes I'm carrying, thought I should get them merged.

### DIFF
--- a/.project.py
+++ b/.project.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # renders a project's final configuration as YAML
 # if run without any arguments, returns a list of known projects
 # requires the builder venv
@@ -32,7 +32,7 @@ config.PROJECTS_FILES += ['src/tests/fixtures/projects/dummy-project.yaml',
 # specific project, specific task
 if args.pname:
     pname = args.pname[0] # multiple projects in future?
-    
+
     if args.task == 'project-data':
         output = project.project_data(pname)
 
--- a/bldr
+++ b/bldr
@@ -19,4 +19,4 @@ if [ -f .use-fabric.flag ]; then
     export BLDR_BACKEND=fabric
 fi
 
-python src/taskrunner.py "$*"
+python3 src/taskrunner.py "$*"


lsh-0:
cheers, @scottaubrey , this looks good.

On my particular flavour of linux `python` points to the latest, but I know on Ubuntu and such it probably assumes Py2

lsh-0 (approved):

scottaubrey:
Screenshot 2022-09-02 at 7 58 00 am

Wow! pre-cog levels of reflexes there @lsh-0  😆

lsh-0:
within a virtualenv (`venv`) `python` will refer to `venv/bin/python`, which is hopefully `python3.8`:

```
$ ls -l venv/bin/ | grep python
lrwxrwxrwx 1 luke luke     9 Aug 31 12:00 python -> python3.8
lrwxrwxrwx 1 luke luke     9 Aug 31 12:00 python3 -> python3.8
lrwxrwxrwx 1 luke luke    18 Aug 31 12:00 python3.8 -> /usr/bin/python3.8
```

scottaubrey:
> within a virtualenv (`venv`) `python` will refer to `venv/bin/python`, which is hopefully `python3.8`:
> 
> ```
> $ ls -l venv/bin/ | grep python
> lrwxrwxrwx 1 luke luke     9 Aug 31 12:00 python -> python3.8
> lrwxrwxrwx 1 luke luke     9 Aug 31 12:00 python3 -> python3.8
> lrwxrwxrwx 1 luke luke    18 Aug 31 12:00 python3.8 -> /usr/bin/python3.8
> ```

Ah, you are right - the symlinks in my `./venv` is setup the same (though different upstream binary path). 

Looking into it a bit more, I've been assuming the venv got activated when running update.sh, which I run every new terminal. Thinking about it now, it will only source the venv in the scope of that script, not my shell. 🤦‍♂️  Thanks for the pointer, really helpful.

lsh-0:
Yeah, `python` or `python3` after the virtualenv has been sourced should be fine, with `python3` probably safer than `python` as a default.

PR #1004, opened 2022-09-09
IRSA permission templates for existing resources

scottaubrey:
This continues the 10% time project to tighten security of resources running on the clusters.

This builds on the work to register a clusters OIDC provider with IAM to allow us to grant permissions to cluster by serviceaccount (roughly a single service) rather than cluster-wide to all services.

This PR copies the current cluster-wide policies to a structure that allows us to define which namespace and serviceaccount should be granted permissions. Again, this will be done on the test cluster first to iron out any issues with the policies, and make sure the controllers running on the cluster can still access the AWS resources they need, before making changes to the prod cluster, or removing the current polcies.


src/buildercore/terraform.py
```diff
@@ -1006,6 +1060,59 @@ def _render_eks_iam_access(context, template):
         'url': '${aws_eks_cluster.main.identity.0.oidc.0.issuer}'
     })
 
+    if 'iam-roles' in context['eks']:
+        for rolename, role_definition in context['eks']['iam-roles'].items():
+            if not role_definition['policy-template']:
```
lsh-0, line 1065: This is a check against a false-y value in `role_definition['policy-template']` but not against `policy-template` being absent in `role_definition`. There are not defaults for this value, either in the `defaults` section of the project config or in [`build_context_eks`](https://github.com/elifesciences/builder/blob/b4215b747c09f326dac7895d2dd759e7dd8350ef/src/buildercore/cfngen.py#L608-L612).

If you *were* to provide defaults, either of the places above would be correct homes, but putting defaults in `terraform.py` (or `trop.py`) should be avoided. By the time it gets to rendering the cloudformation/terraform templates, the context should be whole.

lsh-0:
```python
>>> role_definition = {}
>>> role_definition['policy-template']
  File "<stdin>", line 1, in <module>
KeyError: 'policy-template'

lsh-0: I'm not saying you need to provide defaults, just pointing out it may not be doing what you think it's doing.

lsh-0 (approved):

src/buildercore/terraform.py
scottaubrey: You're right, I wanted to ensure the structure has the required keys, but there is no default for them (hence raise exception).  

I've altered the tests now to test for the keys not existing, rather than the falsey value of a key that might not be set.

PR #1015, opened 2022-10-11
Support for bus-reviewed-preprints—{instance}

nlisgo:
@lsh-0 we need to add support for reviewed-preprints in the search application. While we will not be emit sns notifications upon the creation or modification of reviewed preprints we use bus-sdk-php within the search code.

We must add support for the reviewed-preprint type in the bus. This will allow us to use the bus-sdk-php to process reviewed-preprint in the sqs queue. The queue items will be being created manually for now. 

This is related to: https://github.com/elifesciences/bus-sdk-php/pull/34

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1317,6 +1317,7 @@ bus:
             - bus-metrics--{instance}
             - bus-profiles--{instance}
             - bus-digests--{instance}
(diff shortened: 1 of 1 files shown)


lsh-0:
This is fine. I'll merge it now and create the new sns topic.

PR #1036, opened 2022-11-23
raise IOError when user input required but not possible

lsh-0:
* utils, 'get_input' now raises an IOError when stdin is not available.
* reports, a region is now prompted for when multiple available.
* utils, 'print' statement prompts changed to 'errcho'.


lsh-0:
most of this has been addressed in this PR:
https://github.com/elifesciences/builder/pull/1065

the changes to `find_region` might still be handy

PR #1058, opened 2023-01-31
Increase disk space allocated to elife-libraries--spectrum to 20gb

nlisgo:
We tried to deploy a change to elife-spectrum recently and it ran out of disk space. We cleared out the `/ext/tmp-spectrum` folder which freed up 2GB but it filled up again in the next run.

It's evident that it needs more disk space as 5GB roughly is taken up with the cached-repositories as the elife-articles-xml is ever increasing.

This is preventing necessary changes to elife-spectrum being deployed and it is causing end2end tests to fail.

Related to: https://github.com/elifesciences/elife-spectrum/pull/330

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1216,7 +1216,7 @@ elife-libraries:
             ec2:
                 type: t3.medium # 2 core, 8 GB of memory
             ext:
-                size: 10 # GB
+                size: 20 # GB
         powerful3:
             description: speeding up bot-lax-adaptor tests even more
             ec2:
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -1216,7 +1216,7 @@ elife-libraries:
             ec2:
                 type: t3.medium # 2 core, 8 GB of memory
             ext:
-                size: 20 # GB
+                size: 50 # GB
         powerful3:
             description: speeding up bot-lax-adaptor tests even more
             ec2:


nlisgo:
@lsh-0 if you are happy with this can you merge it in and see it through please. Afterwards we can hit the following pipeline: https://alfred.elifesciences.org/job/test-elife-spectrum/

Then: https://alfred.elifesciences.org/job/test-journal/

projects/elife.yaml
```diff
@@ -1216,7 +1216,7 @@ elife-libraries:
             ec2:
                 type: t3.medium # 2 core, 8 GB of memory
             ext:
-                size: 10 # GB
```
giorgiosironi, line 1219: Would go even larger, 50 or 100 GB as it's a single node being started on demand, and disk space is cheap with respect to CPU time.

giorgiosironi: https://github.com/elifesciences/builder-base-formula/blob/a695d9970284f10cd45142fa976811596329f53f/elife/external-volume.sls#L45 promises to take care of the resizing of the filesystem?

nlisgo: I’ve increased to 50. I will leave any other adjustment to @lsh-0

lsh-0: > https://github.com/elifesciences/builder-base-formula/blob/a695d9970284f10cd45142fa976811596329f53f/elife/external-volume.sls#L45 promises to take care of the resizing of the filesystem?

volume has to be resized while unattached unfortunately. `resizefs` should handle the rest when it comes back up.

lsh-0: > volume has to be resized while unattached unfortunately.

Actually I think this only applies to certain operations on the *root* volume. I had to go through a song and dance recently when switching instances from t2 to t3.

lsh-0:
all done

PR #1065, opened 2023-02-15
deploy, a safer switch_revision_update_instance

lsh-0:
- [x] review

### DIFF
--- a/src/buildercore/cloudformation.py
+++ b/src/buildercore/cloudformation.py
@@ -120,7 +120,6 @@ def stack_creation(stackname, on_start=_noop, on_error=_noop):
         raise
 
 
-# todo: rename. nothing is being bootstrapped here.
 @core.requires_stack_file
 def bootstrap(stackname, context):
     "called by `bootstrap.create_stack` to generate a cloudformation template."
@@ -226,6 +225,11 @@ def template_outputs_map(stackname):
 def template_using_elb_v1(stackname):
     "returns `True` if the stack template file is using an ELB v1 (vs an ALB v2)."
     return trop.ELB_TITLE in template_outputs_map(stackname)
+
+@core.requires_stack_file
+def template_using_elb_v2(stackname):
+    "returns `True` if the stack template file is using an ELB v2 (an ALB)."
+    return trop.ALB_TITLE in template_outputs_map(stackname)
 
 def read_output(stackname, key):
     """finds a literal `Output` from a cloudformation template matching given `key`.
--- a/src/buildercore/cloudformation.py
+++ b/src/buildercore/cloudformation.py
@@ -223,12 +223,12 @@ def template_outputs_map(stackname):
 
 @core.requires_stack_file
 def template_using_elb_v1(stackname):
-    "returns `True` if the stack template file is using an ELB v1 (vs an ALB v2)."
+    "returns `True` if template for `stackname` is using an ELBv1 (vs an ELBv2/ALB)."
     return trop.ELB_TITLE in template_outputs_map(stackname)
 
 @core.requires_stack_file
 def template_using_elb_v2(stackname):
-    "returns `True` if the stack template file is using an ELB v2 (an ALB)."
+    "returns `True` if template for `stackname` is using an ELBv2 (an ALB)."
     return trop.ALB_TITLE in template_outputs_map(stackname)
 
 def read_output(stackname, key):
--- a/src/buildercore/core.py
+++ b/src/buildercore/core.py
@@ -331,10 +331,8 @@ def _ec2_connection_params(stackname, username, **kwargs):
 
 @contextmanager
 def stack_conn(stackname, username=config.DEPLOY_USER, node=None, **kwargs):
-    if 'user' in kwargs:
-        LOG.warning("found key 'user' in given kwargs - did you mean 'username' ??")
-
-    data = stack_data(stackname, ensure_single_instance=False)
+    ensure('user' not in kwargs, "found key 'user' in given kwargs - did you mean 'username' ??")
+    data = ec2_data(stackname, state='running')
     ensure(len(data) == 1 or node, "stack is clustered with %s nodes and no specific node provided" % len(data))
     node and ensure(utils.isint(node) and int(node) > 0, "given node must be an integer and greater than zero")
     didx = int(node) - 1 if node else 0 # decrement to a zero-based value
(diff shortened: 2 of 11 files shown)


src/buildercore/core.py
```diff
@@ -331,10 +331,8 @@ def _ec2_connection_params(stackname, username, **kwargs):
 
 @contextmanager
 def stack_conn(stackname, username=config.DEPLOY_USER, node=None, **kwargs):
-    if 'user' in kwargs:
-        LOG.warning("found key 'user' in given kwargs - did you mean 'username' ??")
-
-    data = stack_data(stackname, ensure_single_instance=False)
+    ensure('user' not in kwargs, "found key 'user' in given kwargs - did you mean 'username' ??")
+    data = ec2_data(stackname, state='running')
```
lsh-0, line 335: looks like this defers to `find_ec2_instances` whose `state` parameter is 'running' by default, so no need to redeclare it here

src/buildvars.py
```diff
@@ -47,30 +47,12 @@ def _update_remote_bvars(stackname, buildvars):
...
-        _update_remote_bvars(stackname, new_data)
-
-    stack_all_ec2_nodes(stackname, _switch_revision_single_ec2_node, username=BOOTSTRAP_USER, concurrency=concurrency)
+# ---
 
 @format_output('python')
 @requires_aws_stack
 def read(stackname):
-    "returns the unencoded build variables found on given instance"
+    "returns the unencoded build variables found on given `stackname`"
```
lsh-0, line 55: "... found on given instance ec2 nodes."

src/utils.py
```diff
@@ -93,6 +91,11 @@ def _pick(name, pick_list, default_file=None, helpfn=None, message='please pick:
         return choice
 
 def uin(param, default=0xDEADBEEF):
```
lsh-0, line 93: trace what else is using this, no surprises please.

src/utils.py
```diff
@@ -49,9 +49,7 @@ def errcho(x):
     return x
 
 def get_input(message):
```
lsh-0, line 51: trace what else is using this, no surprises please.

src/buildvars.py
```diff
@@ -100,6 +86,26 @@ def _fix_single_ec2_node(stackname):
 
     stack_all_ec2_nodes(stackname, (_fix_single_ec2_node, {'stackname': stackname}), username=BOOTSTRAP_USER)
 
+@requires_aws_stack
+def switch_revision(stackname, revision=None, concurrency=None):
+    revision = revision or utils.uin('revision', None)
+    ensure(revision, "a revision is required", utils.TaskExit)
+
+    fix(stackname)
```
lsh-0, line 94: explain

src/buildercore/cloudformation.py
```diff
@@ -227,6 +226,11 @@ def template_using_elb_v1(stackname):
     "returns `True` if the stack template file is using an ELB v1 (vs an ALB v2)."
     return trop.ELB_TITLE in template_outputs_map(stackname)
 
+@core.requires_stack_file
+def template_using_elb_v2(stackname):
+    "returns `True` if the stack template file is using an ELB v2 (an ALB)."
```
lsh-0, line 231: "if template for `stackname` is using an ELB v2 ..."

src/deploy.py
```diff
@@ -81,5 +74,18 @@ def load_balancer_register_all(stackname):
...
         load_balancer_register_all__v2(stackname)
+    else:
+        LOG.warning("no load balancer found: %s", stackname)
+
+@requires_aws_stack
+@requires_aws_stack_template
+def switch_revision_update_instance(stackname, revision=None, concurrency='serial'):
+    "changes the revision of the stack's project and then calls highstate."
+    buildvars.switch_revision(stackname, revision)
+    # if lb present, ensure all nodes are registered,
```
lsh-0, line 87: full stop

src/tests/test_utils.py
```diff
@@ -45,18 +45,30 @@ def test_pick(self, get_input):
...
-        value = utils.uin('project', default='lax')
-        self.assertEqual('lax', value)
+    def test_uin__interactive_with_default(self, get_input):
+        expected = 'lax'
+        actual = utils.uin('project', default='lax')
+        self.assertEqual(expected, actual)
 
     def test_mkdirp_is_idempotent_on_existing_directories(self):
-        utils.mkdirp(".")
+        utils.mkdirp(".") # todo: wtf is this?
```
lsh-0, line 71: todo

src/buildercore/core.py
lsh-0: ... actually, `ec2_data` is `None` by default which is passed to `find_ec2_instances`. 'running' will need to be preserved

src/utils.py
lsh-0: surprisingly little actually. `buildvars.switch_revision`, `cfn.check_user_input` and `project.new`.

`cfn.check_user_input` is used to fetch an `instance_id` if one isn't provided when generating a new stack and probably shouldn't be buried so deep. It has a default of a datestamp.

src/utils.py
lsh-0: a few things were relying on receiving an empty string for 'no input', which is still true, but I'm not 100% sure what would happen when stdin is not available? Indefinite pause? I've switched those to `utils.uin` with an explicit default parameter of `None` so it's no so ambiguous.

src/buildvars.py
```diff
@@ -100,6 +86,31 @@ def _fix_single_ec2_node(stackname):
 
     stack_all_ec2_nodes(stackname, (_fix_single_ec2_node, {'stackname': stackname}), username=BOOTSTRAP_USER)
 
+@requires_aws_stack
+def switch_revision(stackname, revision=None, concurrency=None):
```
lsh-0, line 90: is anything calling this without a revision??

src/master.py
```diff
@@ -222,7 +222,7 @@ def sortbyenv(n):
                         continue
                     LOG.info("*" * 80)
                     LOG.info("updating: %s" % stackname)
-                    prompt and utils.get_input('continue? ctrl-c to quit')
+                    prompt and utils.get_input('continue?')
```
lsh-0, line 225: utils.confirm

src/buildvars.py
lsh-0: can't see anything

PR #1076, opened 2023-03-22
Eks allow controlplane and workers different subnets

scottaubrey:
relates to https://github.com/elifesciences/issues/issues/8188

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -210,6 +210,8 @@ defaults:
             # two subnets in two different availability zones, both used in master nodes and in worker nodes
             subnet-id: subnet-0c60bc3cb24a2816b
             redundant-subnet-id: subnet-009a04cbefc8e3661
+            worker-subnet-id: subnet-0c60bc3cb24a2816b
+            worker-redundant-subnet-id: subnet-009a04cbefc8e3661
             worker:
                 type: t2.small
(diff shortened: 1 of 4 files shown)


projects/elife.yaml
```diff
@@ -210,6 +210,8 @@ defaults:
             # two subnets in two different availability zones, both used in master nodes and in worker nodes
             subnet-id: subnet-0c60bc3cb24a2816b
             redundant-subnet-id: subnet-009a04cbefc8e3661
+            worker-subnet-id: subnet-0c60bc3cb24a2816b
+            worker-redundant-subnet-id: subnet-009a04cbefc8e3661
```
lsh-0, line 214: so I guess the idea is to introduce the concept of a second pair of subnets for separate purposes that, by default, are the same as the subnet and redundant-subnet, then update a cluster instance once the new subnets are in place?

lsh-0 (approved):
this seems ok. I would add a short comment above `worker-subnet-id` in `projects/elife.yaml` about the purpose of the subnets and when they should be different from `subnet` and `redundant-subnet`. Maybe even link to the github issue covering this problem.

A problem I often have when going back through project/infrastructure config is having to guess at the circumstances around _why_ a decision was made, especially if the commit history is unclear.

PR #1077, opened 2023-03-24
bootstrap.sh, updates Github's pubkeys.

lsh-0:
after current one in use was revoked:
https://github.blog/2023-03-23-we-updated-our-rsa-ssh-host-key/


lsh-0:
It's failing to run the tests on the container nodes, which I can't regenerate until the builder changes are in ... chicken meet egg. Merging this.

PR #1102, opened 2023-04-11
terraform.py, drops 'source' from required providers if not present

lsh-0:
@scottaubrey , the change I suggested on the phone. I haven't tested it but think it should work:

Screenshot at 2023-04-11 15-09-18

https://developer.hashicorp.com/terraform/language/v1.1.x/providers/requirements


scottaubrey:
Thanks Luke, I didn't get around to looking at this today, but should have a less meeting heavy day tomorrow.

lsh-0:
This is in the `master` branch now.

PR #1119, opened 2023-05-10
EKS: Add builder configuration to ask terraform not to update `desired-capacity` when it doesn't match config.

scottaubrey:
Useful for when the k8s autoscaler is updating this dynamically, but still allow provisioning

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -243,6 +243,8 @@ defaults:
                 max-size: 3
                 # (currently fixed) number of EC2 nodes
                 desired-capacity: 1
+                # Whether to ignore drift in desired-capacity (due to f.e. the autoscaler taking control)
+                ignore-desired-capacity-drift: false
                 #root:
                 #   size: 20 # GiB, default
             efs: false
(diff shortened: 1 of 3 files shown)


lsh-0:
is this because there are uncontrolled changes to `desired-capacity` happening ?

scottaubrey:
> is this because there are uncontrolled changes to `desired-capacity` happening ?

Changes being made not by builder, yes, but that are controlled by the autoscaler automation. It calculates when there are too few or too many node running based on current workload, and adjusts `desired capacity`, as long as it's within the min and max values also defined here. The whole point of those 3 values is for some automation to read them and some form of metric of load, then adjust `desired-capacity` to meet the metric.

If I remember correctly, flux prod was currently 15 (so 1 less than the last time I corrected builders config) and flux test has 3 (so 1 more than config).

We still need to define `desired-capacity` in builder to create a new cluster ASG, and if we were not running any automation we would still desire to set it in builder, hence the optionality of this new behaviour.

lsh-0:
Gotcha.

PR #1120, opened 2023-05-12
salt, v3006+ introduces some changes to installation.

lsh-0:
it means we can drop certain flags from the salt-boostrap invocation. we can also skip the installation of specific python dependencies that 'fix' bugs.

- [x] review


scripts/clean-stack-for-ami.sh
```diff
@@ -1,14 +1,14 @@
 #!/bin/bash
-# runs as ROOT before AMI creation
-# creation
+# runs before AMI creation creation
```
lsh-0, line 2: typo

PR #1121, opened 2023-05-12
create and define the simple EKS addons in builder config, and test

scottaubrey:


scottaubrey:
Without any changes:
```
------------------------------------------------------------------------

No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, no
actions need to be performed.
INFO - cfn - Create: {'Outputs': {}, 'Parameters': {}, 'Resources': {}}
INFO - cfn - Update: {'Outputs': {}, 'Resources': {}}
INFO - cfn - Delete: {'Outputs': {}, 'Parameters': {}, 'Resources': {}}
INFO - cfn - Terraform delta: 
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
Terraform will perform the following actions:
Plan: 0 to add, 0 to change, 0 to destroy.
```

with this patch to `kubernetes-aws--flux-test` definition:
```
index 5b823c2c..fe12709f 100644
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -2083,6 +2083,11 @@ kubernetes-aws:
                     max-size: 6
                     desired-capacity: 2
                     ignore-desired-capacity-drift: true
+                addons:
+                    kube-proxy:
+                        version: latest
+                    coredns:
+                        version: latest
                 iam-oidc-provider: true
                 iam-roles:
                     kubernetes-autoscaler:
```

We get:
```
------------------------------------------------------------------------

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # aws_eks_addon.eks_addon_coredns will be created
  + resource "aws_eks_addon" "eks_addon_coredns" {
      + addon_name           = "coredns"
      + addon_version        = "v1.9.3-eksbuild.3"
      + arn                  = (known after apply)
      + cluster_name         = "kubernetes-aws--flux-test"
      + configuration_values = (known after apply)
      + created_at           = (known after apply)
      + id                   = (known after apply)
      + modified_at          = (known after apply)
      + resolve_conflicts    = "OVERWRITE"
      + tags                 = {
          + "Cluster"     = "kubernetes-aws--flux-test"
          + "Environment" = "flux-test"
          + "Name"        = "kubernetes-aws--flux-test"
          + "Project"     = "kubernetes-aws"
        }
      + tags_all             = {
          + "Cluster"     = "kubernetes-aws--flux-test"
          + "Environment" = "flux-test"
          + "Name"        = "kubernetes-aws--flux-test"
          + "Project"     = "kubernetes-aws"
        }
    }

  # aws_eks_addon.eks_addon_kube_proxy will be created
  + resource "aws_eks_addon" "eks_addon_kube_proxy" {
      + addon_name           = "kube-proxy"
      + addon_version        = "v1.25.9-eksbuild.1"
      + arn                  = (known after apply)
      + cluster_name         = "kubernetes-aws--flux-test"
      + configuration_values = (known after apply)
      + created_at           = (known after apply)
      + id                   = (known after apply)
      + modified_at          = (known after apply)
      + resolve_conflicts    = "OVERWRITE"
      + tags                 = {
          + "Cluster"     = "kubernetes-aws--flux-test"
          + "Environment" = "flux-test"
          + "Name"        = "kubernetes-aws--flux-test"
          + "Project"     = "kubernetes-aws"
        }
      + tags_all             = {
          + "Cluster"     = "kubernetes-aws--flux-test"
          + "Environment" = "flux-test"
          + "Name"        = "kubernetes-aws--flux-test"
          + "Project"     = "kubernetes-aws"
        }
    }

Plan: 2 to add, 0 to change, 0 to destroy.

------------------------------------------------------------------------
```

Which I think is what we want 👍

scottaubrey:
@lsh-0 As this is replacing existing cluster addons and an existing semi-manual management framework using flux, there's a bit of choreography to turn of management of existing objects so that flux and AWS don't get into a fight over who is managing the resources, and I'm keen not to do it last thing Friday. 

I will likely do all that first thing Monday morning, if you're OK reviewing the general approach I've taken here before I merge and enable? 

There will be two more addons to do, but they also require additional IAM permissions setting up, so I will tackle those next week after (hopefully successfully) enabling these easier ones. I will just build on this generalised approach here with a cross check for IRSA permissions being enabled, then create the necessary IAM resources. I won't bother with asking for review for those if you're happy with that approach?

Then then fun happens - an upgrade to the clusters with hopefully no intervention on my part for these addons.

src/buildercore/terraform.py
```diff
@@ -1229,6 +1229,46 @@ def _render_eks_master_security_group(context, template):
...
+        resource_block['service_account_role_arn'] = service_account_role_arn
+
+    template.populate_resource(
+        'aws_eks_addon',
+        'eks_addon_%s' % name,
+        block=resource_block,
+    )
+
+def _render_eks_addons(context, template):
+    if (lookup(context, 'eks.addons.kube-proxy')):
```
lsh-0, line 1265: this is python, we don't need to wrap expressions in braces unnecessarily.

src/buildercore/terraform.py
```diff
@@ -1229,6 +1229,46 @@ def _render_eks_master_security_group(context, template):
...
+        addon_version = lookup(context, 'eks.addons.kube-proxy.version', 'latest')
```
lsh-0, line 1266: `terraform.py` should rely on the context building that happens in `cfngen.py` to supply these defaults. If the values it needs are not present, it should fail rather than try to fallback to other values.

This is the *intent* anyway, a separation of concerns, but it's not always followed and it certainly wasn't followed in the beginning when the logic for `builder` was still condensing.

src/buildercore/terraform.py
```diff
@@ -1229,6 +1229,46 @@ def _render_eks_master_security_group(context, template):
...
+        _render_eks_addon(context, template, 'kube_proxy', 'kube-proxy', addon_version)
+    if (lookup(context, 'eks.addons.coredns')):
```
lsh-0, line 1268: This is beginning to look repetitive, and you're hardcoding specific addons. Perhaps:

```python
for addon_name, addon_context in context["eks"]["addons"].items():
    _render_eks_addon(context, template, addon_name, addon_context)
```

And change the signature of `_render_eks_addon` to accept the 'addon context'.

The function `_render_eks_addon` is also introducing values that should be handled during context building:
* `configuration_values=False`
* `resolve_conflicts='OVERWRITE'`
* `service_account_role_arn=False`

lsh-0 (approved):
looks fine, no show stoppers to your cluster update, but I'll tweak things in a separate PR

src/buildercore/cfngen.py
```diff
@@ -702,8 +702,8 @@ def build_context_eks(pdata, context):
     addons = {}
     for label, data in pdata['aws']['eks'].get('addons', {}).items():
         addons[label] = {
-            'name': data.get('name', label), # "kube_proxy" or "kube-proxy"
-            'label': label, # "kube-proxy"
+            'name': data.get('name', label), # name of the addon returned from DescribeAddonVersions API request, e.g. kube-proxy
+            'label': data.get('label', label), # local label for terraform resource e.g. "kube_proxy"
```
lsh-0, line 706: yes, this a good place to document these values. :+1:

src/buildercore/terraform.py
lsh-0: Nothing to apologise for, it's valid syntax just not idiomatic. It even looks a little lisp-y which through me:

```clojure
(if (lookup (context "eks.addons.kube-proxy")) ; ...
```

src/buildercore/terraform.py
lsh-0: You can't fix (or review or learn from) non-existent code. Ploughing into it is the best way sometimes, don't worry about it.

> I think the interaction between config and the generated output in terraform will be a little more complicated

It's not complicated *right now* so we can save that engineering effort for another PR.

> I'll maybe fire a PR your way once I've thought about it a bit more today.

More than happy to review changes to `builder`, it helps keep my mental model up to date.

src/buildercore/terraform.py
scottaubrey: Yup I did hardcode them intentionally - these are simple addons, but the next ones require coordination between IAM, cluster and addon resources. My intention was to build that knowledge into code, not config - so when `addon.vpc-cni` is configured, then verify the IAM OIDC is setup, add additional IAM resources needed and set `service_account_role_arn`. So the next if blocks would have had multiple `_render*` function calls. Otherwise, I'd have to assume the required parts are configured and leave it as an exercise to the user to figure out which config is needed to work together, potentially leaving a cluster in a broken state if they don't. 

Assuming you agree fewer foot-guns are better - I think then I will probably need to move some of the current IAM logic up the stack from `terraform.py` into `cfngen.py` to to pull that off, but it sounds like that logic should always have been there. 

While I'm in this head space - some of these configurations have been there to aid migration (i.e. do it on test first, iterate until it works, then roll out to prod), but ideally we would always configure them now the migration is done to remove the number of moving parts. An example is this one - if I can assume our configuration of EKS always enables the IAM-OIDC, then I don't need to verify that is is enabled (so at least 1 less condition and 1 less verification, possibly simpler tests, etc). However, it is still valid to have a cluster without this IAM integration, I just don't see us/me ever using it. Is that your preference from a builder perspective - fewer, but more integrated configurations? or would it be better to keep configuration around?

lsh-0: When dealing with cloudformation/terraform I suggest taking the code from one stable, tested, coherent working version to another. If this code works for your current usecase, merge it in [you just did ;)]. 

If you *know* configuration will be short-lived and won't be reused or extended in the future, feel free to hardcode it in, just remember to also remove it when the transition is done.

PR #1135, opened 2023-05-24
Add support for EKS addons with IAM roles

scottaubrey:
This enables EKS addons to be optionally connected with IAM permissions, either via an existing policy template, or via pre-created managed policies.

This covers the last of the use cases needed to enable all currently manually managed addons and migrate to managed addons in EKS clusters.

Task under https://github.com/elifesciences/issues/issues/7228

### DIFF
--- a/src/buildercore/cfngen.py
+++ b/src/buildercore/cfngen.py
@@ -709,16 +709,27 @@ def build_context_eks(pdata, context):
 
     context['eks'] = pdata['aws']['eks']
 
+    def _build_addon_policy(label, data):
+        return {
+            'service-account': data.get('service-account'),
+            'namespace': data.get('namespace'),
+            'managed-policy': data.get('managed-policy', False),
+            'policy-template': data.get('policy-template', False),
+        }
+
     addons = {}
     for label, data in pdata['aws']['eks'].get('addons', {}).items():
         addons[label] = {
             'name': data.get('name', label), # name of the addon returned from DescribeAddonVersions API request, e.g. kube-proxy
             'label': data.get('label', label), # local label for terraform resource e.g. "kube_proxy"
             'version': data.get('version', 'latest'),
-            'configuration_values': None,
-            'resolve_conflicts': 'OVERWRITE',
-            'service_account_role_arn': None,
+            'configuration-values': data.get('configuration-values', None),
+            'resolve-conflicts': 'OVERWRITE',
         }
+
+        # Check if this addons needs additional permissions granting to a kubernetes service account via IRSA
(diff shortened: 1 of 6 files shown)


src/buildercore/cfngen.py
```diff
@@ -709,16 +709,27 @@ def build_context_eks(pdata, context):
...
             'version': data.get('version', 'latest'),
-            'configuration_values': None,
-            'resolve_conflicts': 'OVERWRITE',
-            'service_account_role_arn': None,
+            'configuration-values': data.get('configuration-values', None),
+            'resolve-conflicts': 'OVERWRITE',
         }
+
+        # Check if this addons needs additional permissions granting to a kubernetes service account via IRSA
+        addons[label]['irsa-role'] = _build_addon_policy(label, data.get('irsa-role')) if data.get('irsa-role', False) != False else False
```
lsh-0, line 731: This is a middling, piffling, hair splitting point I'm about to make with no objective right or wrong answer.

---

Two schools of thought on empty data: it should be excluded or explicitly set to a none-type value, of which Python has a literal `None`.

Excluding it completely means you won't have to deal with semantic differences around `None`, `False`, an empty map, etc, nor with transitioning legacy code when you change what `None`, `False`, empty map, etc mean in this context. It's simple presence or absence has no alternative meaning (but is not always possible given circumstances).

Setting it to `None` is possibly friendlier for humans and readability but redundant when excluding it entirely is possible.

You've set it to `False` here when it could be `None` or excluded entirely so we have `foo['bar'] = False` that is less readable than `foo['bar'] = None`. That's purely subjective, data set to `False` may be overwhelmingly more readable for you.

I prefer excluding the data entirely when possible as setting it to `False` or `None` or an empty map in the past has often come back to bite me.

---

This

    addons[label]['irsa-role'] = _build_addon_policy(label, data.get('irsa-role')) if data.get('irsa-role', False) != False else False

could be:

    irsa_role = data.get('irsa-role')
    if irsa_role:
        addons[label]['irsa-role'] = _build_addon_policy(label, irsa_role)

and then later this:

    if addon['irsa-role']:

would have to be switched to:

    if 'irsa-role' in addon:

and nothing else need be touched.

lsh-0: PR: https://github.com/elifesciences/builder/pull/1137

lsh-0 (approved):

src/buildercore/cfngen.py
scottaubrey: Thanks! I'm happy with the change and explanation - it makes sense 👍

lsh-0: No worries

PR #1167, opened 2023-07-18
update max node for prod k8s cluster

scottaubrey:


scottaubrey:
> that's a lotta nodes. I feel less bad reserving extra capacity for a t3.large this year. Does k8s kind of generally idle at the `desired-capcity` level (16)?

Generally it has been, yes. It sometimes adds some nodes so that deployments can rollout (i.e. we have two of some services while one deploys and the other is replaced) so max is a bit larger than "usual". 

This increase is because we're gearing up to deploying new services for the import process in EPP so there will be a new baseline and new max nodes. I'm guessing at the amount of new resource right now, but moving that max up so we have headroom for a few more nodes, without jeopardising the ability to rollout other service upgrades.

PR #1187, opened 2023-09-20
increase max number of nodes in flux-prod

scottaubrey:


scottaubrey:
This is the maximum nodes to scale to - most of the time we're happy with 24-26. However, when we roll out a new instance of a service we need headroom in the cluster to roll it out to alongside the old, wait for it to be healthy, then it terminates the old one. 

For smaller services (like the epp webapp), there's usually enough headroom around on the nodes running that this is smooth, but some of the bigger services (like the data hub worker, or the EPP import worker, or peerscout), they fill a node on their own. In other words, we need a spare node just to roll out a new version. The autoscaler handles this for us, but needs the ASG to have a max setting higher than our run rate to add nodes to. Once the new version is rolled out and the old ones terminated and the autoscaler downscales.

This is what happened here - we were  running at 26 nodes, but then couldn't roll out a new version of EPP import without more nodes.

As for what is running for the node run rate (obviously a snapshot in time):

```
> kubectl get pods -A | grep Running | wc -l
     245

> > kubectl get deployments.apps,statefulsets.apps -A | grep '/' | wc -l
      95
```

The numbers are roughly:
- 36 data-hub services (API, web, database, worker, queue) x 3 envs
- 22 EPP (API, webapp, database, worker, queue) x 2 envs
- 33 cluster services (logging, monitoring, database operators, autoscaler)
- 4 misc - peerscout, basex-validator, podinfo (test service).

95 services, with 245 pods on 26 nodes sounds pretty efficient to me. 

I'll admit, I rarely do much in optimisation of the services requests until there's a problem, but even with that limited optimisation, we get pretty good utilisation of those nodes:
Screenshot 2023-09-22 at 8 29 21 am

Some of the newer services though have guessed requests, and overtime I'll likely be able to right size them better. For now though, I think it's alright.

PR #1298, opened 2024-11-07
Remove dependance on a context for observer--end2end existing in s3 elifesciences/issues#9044

scottaubrey:
Relates to: elifesciences/issues#9044

### DIFF
--- a/src/tests/test_buildercore_bootstrap.py
+++ b/src/tests/test_buildercore_bootstrap.py
@@ -36,27 +36,29 @@ pillar_roots:
     def test_unsub_sqs(self):
         stackname = 'observer--end2end'
         with open(join(self.fixtures_dir, 'sns_subscriptions.json')) as fh:
-            fixture = json.load(fh)
-        with mock.patch('buildercore.core._all_sns_subscriptions', return_value=fixture):
-            # observer no longer wants to subscribe to metrics
-            context = {stackname: ['bus-articles--end2end', 'bus-metrics--end2end']}
-            del context[stackname][1]
+            subs_fixture = json.load(fh)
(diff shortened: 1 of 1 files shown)


scottaubrey:
> I think these tests are more trouble than they are worth:
> 
> * every time we update a stack involving an SQS queue, we unsubscribe from all topics and then subscribe again (because it's too difficult to compute a diff)
> * we only test unsubscription of queues to topics, and not (re-)subscription
> * we only test in `dry_run` mode, so the first half of the unsubscription code
> 
> We have a better strategic question here: how can an application team create and maintain an SQS resource for a project in a certain environment (e.g. `search--prod`), and tune what it subscribes to, without using `builder`? It is not a straightforward migration path to Kubernetes, unless [ACK](https://aws-controllers-k8s.github.io/community/docs/community/overview/) or equivalents have started to see some adoption.

Agreed, I did consider just removing them to unblock PRs before finally cracking the mock expectation. I'll bear in mind for next time if it creates more problems, but the focus is more on migration than maintain.

As to the question you pose, we have seen some adoption using ACK for the search project with creating a [queue](https://github.com/elifesciences/journal-team-deployment/blob/main/kustomizations/search/queue/search-queue.yaml) and  [subscriptions](https://github.com/elifesciences/journal-team-deployment/blob/main/kustomizations/search/queue-subscriptions/search-queue-subscriptions.yaml). It worked well enough, but the dependency between the resources was a little tricky to express well. Probably k8s would handle it in an eventually consistent way, but in that example, we created a kustomization dependency between them, which worked out well. 

Having said that, it's not *that* easy to wrap your head around, and a nicer abstraction would be really handy, if we can come up with one.

Thanks @giorgiosironi

PR #1305, opened 2024-11-15
switch box to a multi-arch, multi-provider base box

scottaubrey:
This allows for virtualbox and vmware as providers.

I've tested this on vmware/mac/arm64 and virtualbox/linux/amd64

### DIFF
--- a/projects/elife.yaml
+++ b/projects/elife.yaml
@@ -422,7 +422,7 @@ defaults:
     gcp:
         bigquery: false
     vagrant:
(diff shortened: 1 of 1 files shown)


ali-tm-amin (approved):

scottaubrey:
Virtualbox on Mac with a newer version of VB started installing VboxAdditions inside the VM, then seemed to hang after installation complete. 

Unless this is reliable, we shouldn't change image to support VMWare for now.

